PR #1, opened 2025-03-25
Embed simplified version of pg_replicate

nbiscaro:
Replaces the external `pg_replicate` dependency with a simplified version integrated directly into the codebase.

- Removed dependency on `pg_replicate` from `Cargo.toml`
- Added new direct dependencies 
- Created a new replication module 
- Added Postgres replication client (`clients/postgres.rs`)
- Moved postgres source and pipeline components into the replication module
- Moved our custom `sink.rs` into `pipeline/sinks`. 
- Updated import paths in main.rs to reference the new module structure

Theres still some dead code given our implementation is fairly simple for now, but a big improvement on using `pg_replicate` directly.

### DIFF
--- a/rust/moonlink/src/main.rs
+++ b/rust/moonlink/src/main.rs
@@ -2,8 +2,8 @@ mod replication;
 
 use replication::pipeline::{
     batching::{data_pipeline::BatchDataPipeline, BatchConfig},
-    sources::postgres::{PostgresSource, TableNamesFrom},
     sinks::sink::Sink,
+    sources::postgres::{PostgresSource, TableNamesFrom},
     PipelineAction,
 };
 use std::{error::Error, time::Duration};
--- a/rust/moonlink/src/replication/conversions/cdc_event.rs
+++ b/rust/moonlink/src/replication/conversions/cdc_event.rs
@@ -93,7 +93,8 @@ impl CdcEventConverter {
             .old_tuple()
             .map(|tuple| Self::try_from_tuple_data_slice(column_schemas, tuple.tuple_data()))
             .transpose()?;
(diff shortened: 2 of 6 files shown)


rust/moonlink/src/replication/mod.rs
```diff
@@ -0,0 +1,7 @@
+//TODO: Remove this once we determine our actual usage of the replication module
+#![allow(dead_code)]
```
nbiscaro, line 2: Silencing dead code warnings for now until we have a better idea of exactly what we want from `pg_replicate`. Most of the warnings are from very basic elements that we will likely end up using as our implementation grows.

rust/moonlink/src/main.rs
```diff
@@ -1,11 +1,11 @@
-mod sink;
+mod replication;
 
-use pg_replicate::pipeline::{
+use replication::pipeline::{
     batching::{data_pipeline::BatchDataPipeline, BatchConfig},
+    sinks::sink::Sink,
```
nbiscaro, line 5: We should probably rename our `sink.rs` to something else.

rust/moonlink/src/replication/conversions/cdc_event.rs
```diff
@@ -0,0 +1,200 @@
...
+        }
+    }
+}
+
+#[derive(Debug)]
+pub enum CdcEvent {
+    Begin(BeginBody),
+    Commit(CommitBody),
+    Insert((TableId, TableRow)),
+    Update((TableId, Option<TableRow>, TableRow)),
```
nbiscaro, line 186: Noting that this is part of our custom implementation. Everything else is taken as is from `pg_replicate`.

nbiscaro:
We will make the changes directly to the pg_replicate fork and continue to use this as a dependency from here rather than directly embedding.

PR #4, opened 2025-04-11
Stream uncommitted transactions

nbiscaro:
Adds support for streaming concurrent uncommitted transactions. 

This leverages the previous update to logical replication v2 in [`rust-postgres`](https://github.com/Mooncake-Labs/rust-postgres/commit/e329510b1b6104a07fa7523bdd6deee27707773c) and [`pg_replicate`](https://github.com/Mooncake-Labs/pg_replicate/commit/6185151c9746cece94fa2bb3792bbd536ac35b93).

We do this by allocating a `mem_slice` and pending deletions per transaction. We apply all events to this `mem_slice` during the transactions lifetime. When a `StreamCommit` is received, rather than trying to merge and remap the transaction specific deletion locations with the main table `mem_slice`, we simply `flush_transaction_stream` to disk and remap the deletion positions to their location on disk. Note that in the streaming case we do not know the final LSN at delete time. At commit time we iterate the deletion record, add this LSN, and add them to the  `snapshot_task`. 

 In the case of `StreamAbort` we simply drop the `mem_slice`. 

Having an isolated `mem_slice` per transaction allows us to easily rollback the uncommitted changes and also lets us stream transactions concurrently. (Note: xids are never interleaved within a stream segment but may be interleaved between segments.)

We also add `xact_id` and streaming message event functions to `table_handler.rs` to be called from the sink. 

TODO: We should add a proper test setup and tear down. Currently lots of repeated code and will quickly grow even for basic unit testing. Can do so in a follow-up PR.


rust/moonlink/src/storage/mooncake_table.rs
```diff
@@ -254,49 +269,123 @@ impl MooncakeTable {
...
+        Ok(())
     }
 
-    pub fn _commit_in_stream_batch(&mut self, _xact_id: u64) -> Result<()> {
-        todo!("Implement commit in stream batch");
+    pub fn delete_in_stream_batch(&mut self, row: MoonlinkRow, xact_id: u32) {
+        let lookup_key = (self.metadata.get_lookup_key)(&row);
+        let mut record = RawDeletionRecord {
+            lookup_key,
+            lsn: 0, // Updated at commit time
```
nbiscaro: When streaming, we don't know the final LSN of the transaction until commit time.

rust/moonlink/src/storage/mooncake_table.rs
```diff
@@ -412,8 +501,6 @@ impl MooncakeTable {
         if next_snapshot_task.new_record_batches.len() > 0 {
             let new_batches = take(&mut next_snapshot_task.new_record_batches);
             // previous unfinished batch is finished
-            assert!(snapshot.batches.values().last().unwrap().data.is_none());
-            assert!(snapshot.batches.keys().last().unwrap() == &new_batches.first().unwrap().0);
```
nbiscaro: When handling separate `mem_slices` we are using non globally unique `batch_ids`. Depending on how they are merged we can't assume they will be monotonically increasing. 

I imagine the fact that they are now non-unique will cause issues in `batch_id_to_idx`. The simple fix is to make batch_ids globally unique but I'm not sure if this is correct.

rust/moonlink/src/storage/mooncake_table.rs
```diff
@@ -254,49 +269,123 @@ impl MooncakeTable {
...
+            pos: None,
+            _row_identity: None,
+        };
+
+        let stream_state = self.get_or_create_stream_state(xact_id);
+        let pos = stream_state.mem_slice.delete(&record);
+
+        // This is fine since we will flush and remap to disk position on commit
+        record.pos = pos;
+        stream_state.new_deletions.push(record);
```
nbiscaro: This may seem a little odd at first, since `pos` is relative to the transaction's `MemSlice`.  
However, this is safe and intentional, because we always remap it later at snapshot time.  
The row we're deleting can be in one of the following places:

### 1. The row is in this transaction’s `MemSlice`:
- `stream_state.mem_slice.delete(...)` searches the local index, finds the record, and returns its `pos`.  
- We store that `pos` here.  
- At commit/snapshot time, we flush the `MemSlice` and remap `pos` to the corresponding position in the resulting `DiskSlice`, using `batch_id_to_idx`.

### 2. The row is already flushed to disk *or* lives in the main table’s global `MemSlice`:
- In these cases, `stream_state.mem_slice.delete(...)` finds nothing and returns a null/invalid `pos` (e.g., `None`).  
- We still track the deletion, but at snapshot time, we resolve the record's position by searching:
  - the global `MemIndex` (for in-memory rows), and/or  
  - the file index (for disk-flushed rows).  
- We then overwrite the placeholder `pos` with the resolved global or on-disk position.

rust/moonlink/src/storage/mooncake_table.rs
```diff
@@ -352,17 +354,27 @@ impl MooncakeTable {
...
+                    deletion.lsn = lsn;
+                }
+            }
+            for deletion in snapshot.lock().unwrap().uncommitted_deletion_log.iter_mut() {
+                if let Some(deletion) = deletion {
+                    if deletion.xact_id == Some(xact_id) {
+                        deletion.lsn = lsn;
+                    }
+                }
+            }
```
nbiscaro: As part of flush (aka commit) for the stream case we update the lsns for all deletion records for the current `xact_id`. This means iterating over `new_deletions` and `uncommitted_deletions`.

Is it fine that we acquire a lock here to iterate over `uncommited_deletions`

rust/moonlink/src/storage/mem_slice.rs
```diff
@@ -75,6 +75,10 @@ impl MemSlice {
     pub fn get_commit_check_point(&self) -> RecordLocation {
         self.column_store.get_commit_check_point()
     }
+
+    pub fn find_position(&self, lookup_key: i64) -> Option<RecordLocation> {
+        return self.mem_index.get(&lookup_key).cloned();
```
zhousun: you need to do something similar to delete(&mut self, record: &RawDeletionRecord.
(mem_index is a multimap, and we are not deleting from it)
so instead of just call mem_index.get (which returns first element in bucket), we need to get a list of matches from index/ check if there is one not deleted.

or probably just add option param to delete() and avoid duplicate logic?

zhousun: also please add a devNote, there will be a very weird edge case when we support table with no pk (we use full row as identifier).

if there are two identical rows A and B.
A is in main memslice, and B is on disk.

If we delete row in streaming xact and we found row A.
And then nonstreaming xact may try to deletes row A again.

The correct final state is both row A and row B being deleted.

I don't have a full design for how that works yet.

rust/moonlink/src/storage/mooncake_table.rs
```diff
@@ -254,49 +270,140 @@ impl MooncakeTable {
...
+            self.next_snapshot_task.new_deletions.push(record);
+        }
     }
 
-    pub fn _abort_in_stream_batch(&mut self, _xact_id: u64) -> Result<()> {
-        todo!("Implement abort in stream batch");
+    pub fn abort_in_stream_batch(&mut self, xact_id: u32) {
+        // Simply remove the transaction stream state
+        self.transaction_stream_states.remove(&xact_id);
```
zhousun: what happens to the uncommited deleted?

rust/moonlink/src/storage/mooncake_table.rs
```diff
@@ -254,49 +270,140 @@ impl MooncakeTable {
...
+            for deletion in snapshot_task.new_deletions.iter_mut() {
+                if deletion.xact_id == Some(xact_id) {
+                    deletion.lsn = lsn;
+                }
+            }
+            let mut guard = snapshot.write().unwrap();
+            for deletion in guard.uncommitted_deletion_log.iter_mut() {
+                if let Some(deletion) = deletion {
+                    if deletion.xact_id == Some(xact_id) {
+                        deletion.lsn = lsn;
```
zhousun: avoid taking lock in the sync function, instead add them to snapshot task.

rust/moonlink/src/storage/mem_slice.rs
nbiscaro: resolved in this commit: https://github.com/Mooncake-Labs/moonlink/pull/4/commits/7f18fee64c2f650fc2ecd103176d1606fd19b479

rust/moonlink/src/storage/mooncake_table.rs
nbiscaro: good catch, now we also need to delete from `uncommitted_deletion_log` when dropping the transaction.

rust/moonlink/src/storage/mooncake_table.rs
nbiscaro: Updated:

We track `xact_id, flush_lsn` in flushed_xacts and update deletes that match the `xact_id` during snapshot
We track `xact_id` in `aborted_xacts`  and drop any deletions that match the `xact_id`

PR #5, opened 2025-04-15
Cover flush before snap edgecase

nbiscaro:
Guards against two somewhat related edge cases.

1) The first case where we delete an in memory row, then flush, and then apply the snapshot.

The deletion record is not remapped during flush, since it is not processed at this point.

Consequently, when the snapshot is taken, the deletion record still points to the in memory row. When we attempt to commit this deletion it fails since the batch has since been flushed.

In this case the deletion is already applied so we can simply add to `commited_deletion_log`. 

2) The second case is where we attempt to remap a deletion record that points to a row that never made it to disk. 

This occurs when we have uncommitted deletes recorded in the snapshot, then flush a batch that has these deletions, then in the snapshot note that the batch was flushed and attempt to remap, but fail since these rows never made it to disk.


rust/moonlink/src/storage/disk_slice.rs
```diff
@@ -160,13 +160,17 @@ impl DiskSliceWriter {
...
-                    new_location.1,
-                );
+                // Guard the case where the record was deleted before making it to disk
+                // if let Some(new_location) = self.row_offset_mapping.get(&old_location) {
+                //     deletion.pos = RecordLocation::DiskFile(
+                //         FileId(Arc::new(self.files[new_location.0].0.clone())),
+                //         new_location.1,
+                //     );
+                // }
             }
```
zhousun: remove?

nbiscaro: oops yes from a temp commit. Removed.

PR #14, opened 2025-04-22
[storage] [1/N] Store data files for iceberg snapshot

dentiny:
This PR is the first part for iceberg snapshot, which dumps the data file into hard-coded in-memory catalog.

Dev plan and followup PR (in order):
- Implement deletion file related load and store operations.
- Implement store and load for manifest files, manifest lists, and catalog metadata, which is not supported in iceberg-rust.
  + pyiceberg already implemented everything: https://gist.github.com/dentiny/1c3d09fb391db18b9d52183947fddd8a, a committed transaction automatically creates manifest files, manifest list, etc
  + but it's not supported in rust sdk, https://github.com/apache/iceberg-rust/discussions/1232, so we need to implement our own
- Current implementation only works for in-memory catalog (intentionally for the ease of dev), we need to support remote access.
  + Before development, setup local minio for integration testing purpose.
- (largely unrelated) Support all data types, other than major primitive types.


Cargo.toml
```diff
@@ -14,14 +14,19 @@ edition = "2021"
 async-trait = "0.1"
 
 tokio-postgres = { git = "ssh://git@github.com/Mooncake-labs/rust-postgres.git"}
-arrow = "54.2.1"
+arrow = "53.4.1"
```
dentiny, line 17: Latest iceberg-rust relies on a few old versions of arrow and parquet related packages, so downgrade here to make sure only one version exists in the dependency tree.

dentiny:
There're two followup for this PR:
- we need to use latest development version for iceberg-rust, a few necessary features (like deletion file) is only available there;
- should use rest catalog for dev and testing, all other catalogs doesn't have transaction support (i.e. `update_table`).

PR #15, opened 2025-04-22
[storage] [2/N] Use rest catalog for production and testing

dentiny:
This PR does a few things:
- Use rest catalog to replace in-memory catalog, because as of now, it's the only catalog type which supports transaction
  + Function support: https://github.com/apache/iceberg-rust/blob/53ceb2b5426f56efd1152a9daaa2fca25847b6ff/crates/catalog/rest/src/catalog.rs#L727
  + There's officially maintained docker image for testing: https://hub.docker.com/r/tabulario/iceberg-rest
- Use docker compose to checkin the above test container into moonlink devcontainer
  + There're three ways to do that: (1) compile from source code (java) and start it ourselves; (2) leverage docker-in-docker; (3) docker compose to start container at devcontainer initialization
  + In this PR I decided to go (3) because it's the easiest to achieve; the drawback of which is our tests become stateful, so I write util functions to cleanup before tests;
  + dind is generally discouraged: https://jpetazzo.github.io/2015/09/03/do-not-use-docker-in-docker-for-ci/, and require more setup like privilidged mode and bind to host machine
  + I usually prefer compiling from source code (if it's in python/golang/C++), but java seems more intimidating considering the dependency we need to install
- I also upgrade the iceberg-rust dependency to latest main branch commit
  + iceberg latest release lacks a few key features we need, for example, deletion file write
  + risingwave did the same thing: https://github.com/risingwavelabs/risingwave/blob/a5353e4b3c3ec083588dbc313a9c6276b6e69274/Cargo.toml#L167-L174

### DIFF
--- a/.devcontainer/docker-compose.yml
+++ b/.devcontainer/docker-compose.yml
@@ -3,7 +3,7 @@ services:
     image: mcr.microsoft.com/devcontainers/rust:latest
     volumes:
       - ..:/workspaces/pg_moonlink:cached
-    network_mode: host
+    network_mode: service:postgres
     command: sleep infinity
 
   postgres:
@@ -11,10 +11,7 @@ services:
     environment:
       POSTGRES_PASSWORD: postgres
     command: ["-c", "wal_level=logical"]
-    network_mode: host
 
   iceberg:
     image: tabulario/iceberg-rest:latest
-    ports:
-      - "8181:8181"
-    network_mode: host
+    network_mode: service:postgres


Cargo.toml
```diff
@@ -14,9 +14,9 @@ edition = "2021"
 async-trait = "0.1"
 
 tokio-postgres = { git = "ssh://git@github.com/Mooncake-labs/rust-postgres.git"}
-arrow = "53.4.1"
+arrow = "54.2.1"
```
dentiny, line 17: After using the latest version of iceberg-rust, I bump up the arrow/parquet related version to what we were: https://github.com/Mooncake-Labs/moonlink/pull/14/files

src/moonlink/src/storage/iceberg/iceberg_snapshot.rs
```diff
@@ -102,7 +102,7 @@ async fn get_or_create_iceberg_table(
             let tbl_creation = TableCreation::builder()
                 .name(table_name.to_string())
                 .location(format!(
-                    "memory://{}/{}",
+                    "file:///tmp/iceberg-test/{}/{}",
```
dentiny, line 105: One good thing is we could check local filesystem for debugging.

PR #36, opened 2025-04-25
[storage] [4/N] Implement filesystem catalog

dentiny:
In this PR, I implemented the basic functionality for filesystem catalog, which serves two purposes:
- As unit test utils, which is better than official maintained rest docker image due to its hermeticity and zero setup pre-requisite;
- It could be used for pg_mooncake for easier user adoption.

Some thing I think low-priority and leave TODO for:
- Property related operations, it's not required for file append and deletion, and it's not fully implemented in java SDK also: https://github.com/apache/iceberg/blob/01fe380d455949abb49ebfecd9509afce8764fae/core/src/main/java/org/apache/iceberg/hadoop/HadoopCatalog.java#L353-L363
- Allow concurrent access; it's not easy to achieve via filesystem (file lock is not straightforward to use)
- Add cache for metadata, should be considered together with concurrent access

Next PR will be integrate filesystem catalog with existing snapshot operations.

The most important operation for catalog is `update_table`, here's my interpretation for iceberg transaction commit path:
1. create a transaction: `txn = Transaction::new(catalog)`
2. create an operation: `action = txn.fast_append(...)`
Take append action for example, there's a snapshot inside, and used at commit
3. make changes to data files: `action.add_data_files(...)`
4. apply the operation: `action.apply(...)`
(1) delegate apply to the internal snapshot action: https://github.com/apache/iceberg-rust/blob/841ef0a3dc87a3ecffaeb1aa62ca9ca4ea4c1712/crates/iceberg/src/transaction/append.rs#L163-L165
(2) create and write manifest files: https://github.com/apache/iceberg-rust/blob/841ef0a3dc87a3ecffaeb1aa62ca9ca4ea4c1712/crates/iceberg/src/transaction/snapshot.rs#L299-L301
(3) create and write manifest list: https://github.com/apache/iceberg-rust/blob/841ef0a3dc87a3ecffaeb1aa62ca9ca4ea4c1712/crates/iceberg/src/transaction/snapshot.rs#L312-L334
(4) apply changes to in-memory transaction states and return: https://github.com/apache/iceberg-rust/blob/841ef0a3dc87a3ecffaeb1aa62ca9ca4ea4c1712/crates/iceberg/src/transaction/snapshot.rs#L347-L372
5. commit the transaction: the completion of a transaction is marked as catalog pointer change, so here catalog takes over the left work
(1) take snapshot changed, reflect at metadata files and persist
(2) update version hint file

### DIFF
--- a/src/moonlink/src/storage/iceberg/file_catalog.rs
+++ b/src/moonlink/src/storage/iceberg/file_catalog.rs
@@ -3,7 +3,9 @@ use std::path::PathBuf;
 
 use async_trait::async_trait;
 use iceberg::io::FileIO;
-use iceberg::spec::{TableMetadata, TableMetadataBuilder};
+use iceberg::spec::{
+    SnapshotReference, SnapshotRetention, TableMetadata, TableMetadataBuilder, MAIN_BRANCH,
+};
 use iceberg::table::Table;
 use iceberg::Error as IcebergError;
 use iceberg::Result as IcebergResult;
@@ -19,6 +21,7 @@ use iceberg::{
 /// 1. Implement property related functionalities.
 /// 2. Implement concurrent accesses features, currently we assume it's only single thread access.
 /// 3. The initial version access everything via filesystem, for performance consideration we should cache metadata in memory.
+/// (not related to functionality) 4. Set snapshot retention policy at metadata.
 
 /// Iceberg table format from filesystem's perspective:
 /// - data
@@ -69,15 +72,17 @@ impl FileSystemCatalog {
             table_ident.namespace.to_url_string(),
             table_ident.name()
         );
+
+        // Get metadata file path via version hint file.
         let version_hint_path = format!("{}/version-hint.text", metadata_directory);
         let input_file: iceberg::io::InputFile = self.file_io.new_input(&version_hint_path)?;
         let version = String::from_utf8(input_file.read().await?.to_vec()).expect("");
+
+        // Get table metadata.
         let metadata_file_path = format!("{}/v{}.metadata.json", metadata_directory, version);
         let input_file: iceberg::io::InputFile = self.file_io.new_input(&metadata_file_path)?;
         let metadata_content = input_file.read().await?.to_vec();
         let metadata = serde_json::from_slice::<TableMetadata>(&metadata_content)?;
-
-        // TODO(hjiang): Need to make sure snapshot information (i.e. current snapshot) is correctly populated.
         Ok((metadata_file_path, metadata))
     }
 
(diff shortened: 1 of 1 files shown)


src/moonlink/src/storage/iceberg/file_catalog.rs
```diff
@@ -0,0 +1,777 @@
...
+/// - data
+///   - parquet files
+/// - metdata
+///   - version hint file
+///     + version-hint.text, contains the latest version number for metadata
+///   - metadata file
+///     + v0.metadata.json ... vn.metadata.json
+///     + records table schema
+///   - snapshot file / manifest list
+///     + snap-<snapshot-id>-<attempt-id>-<commit-uuid>.avro
```
dentiny, line 37: Reference: https://github.com/apache/iceberg-rust/blob/841ef0a3dc87a3ecffaeb1aa62ca9ca4ea4c1712/crates/iceberg/src/transaction/snapshot.rs#L281-L291

src/moonlink/src/storage/iceberg/file_catalog.rs
```diff
@@ -0,0 +1,780 @@
...
+///     + version-hint.text
+///     + contains the latest version number for metadata
+///   - metadata file
+///     + v0.metadata.json ... vn.metadata.json
+///     + records table schema
+///   - snapshot file / manifest list
+///     + snap-<snapshot-id>-<attempt-id>-<commit-uuid>.avro
+///     + points to manifest files and record actions
+///   - manifest files
+///     + <commit-uuid>-m<manifest-counter>.avro
```
dentiny, line 40: Reference: https://github.com/apache/iceberg-rust/blob/841ef0a3dc87a3ecffaeb1aa62ca9ca4ea4c1712/crates/iceberg/src/transaction/snapshot.rs#L158-L165

src/moonlink/src/storage/iceberg/file_catalog.rs
```diff
@@ -0,0 +1,780 @@
...
+///     + which points to data files and stats
+
+#[derive(Debug)]
+pub struct FileSystemCatalog {
+    file_io: FileIO,
```
dentiny, line 47: In theory, `file_io` is an abstraction for storage backends, so all IO operation should go through file io instance, but currently it only supports file-level operation but not directory level operations. So in this PR, I mix `std::fs` and `file_io`.

dentiny:
CC @dpxcc / @zhousun welcome to post-merge comments. :)

src/moonlink/src/storage/iceberg/file_catalog.rs
```diff
@@ -0,0 +1,869 @@
...
+        assert_eq!(
+            table.identifier(),
+            &table_ident,
+            "Updated table identifier should match"
+        );
+        assert_eq!(
+            table_metadata.current_snapshot_id(),
+            Some(1),
+            "Current snapshot ID should be 1"
+        );
```
dentiny, line 865: I think I should check all the metadata file existence, will followup in the next PR.

PR #57, opened 2025-04-28
Implement index

zhousun:
## Summary
Persisted index implementation

// Hash index
// that maps a u64 to [seg_idx, row_idx]
//
// Structure:
// Buckets:
// [entry_offset],[entry_offset]...[entry_offset]
//
// Values
// [lower_bit_hash, seg_idx, row_idx]

write to disk and mmap to search

also added basic stress/ perf test.

## Related Issues

https://github.com/Mooncake-Labs/moonlink/issues/21

### DIFF
--- a/src/moonlink/src/storage/index/persisted_bucket_hash_map.rs
+++ b/src/moonlink/src/storage/index/persisted_bucket_hash_map.rs
@@ -1,6 +1,7 @@
 use crate::storage::storage_utils::{FileId, RecordLocation};
 use bitstream_io::{BigEndian, BitRead, BitReader, BitWrite, BitWriter};
 use memmap2::Mmap;
+use std::collections::BinaryHeap;
 use std::fmt;
 use std::fmt::Debug;
 use std::fs::File;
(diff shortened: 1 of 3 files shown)


src/moonlink/src/storage/index/persisted_bucket_hash_map.rs
```diff
@@ -0,0 +1,685 @@
...
+    file_name: String,
+    directory: PathBuf,
+    entry_writer: BitWriter<BufWriter<File>, BigEndian>,
+    current_bucket: u32,
+    current_entry: u32,
+}
+
+impl IndexBlockBuilder {
+    pub fn new(bucket_start_idx: u32, bucket_end_idx: u32, directory: PathBuf) -> Self {
+        let file_name = format!("index_block_{}.bin", uuid::Uuid::new_v4());
```
dentiny, line 193: nit: I usually place uuid as prefix to avoid hotspoting, if range query is not a requirement

src/moonlink/src/storage/index/persisted_bucket_hash_map.rs
```diff
@@ -0,0 +1,685 @@
...
+        }
+        let bucket_start_offset = (self.current_entry as u64)
+            * (metadata.hash_lower_bits + metadata.seg_id_bits + metadata.row_id_bits) as u64;
+        for bucket in self.buckets {
+            self.entry_writer
+                .write_unsigned_var(metadata.bucket_bits, bucket)
+                .unwrap();
+        }
+        self.entry_writer.byte_align().unwrap();
+        drop(self.entry_writer);
```
dentiny, line 247: Curious should we use `sync_all` instead of `drop`?
According to the [doc](https://doc.rust-lang.org/std/fs/struct.File.html#method.sync_all),
- Failure happen during `drop` are silently ignored;
- `sync_` family ensure persistence to storage if return ok.

zhousun: ++

PR #71, opened 2025-05-01
[storage] [8/N] Deletion vector store and load

dentiny:
## Summary

This PR implements store and load for deletion vector (which is introduced in iceberg v3 spec).
In the initial version, I made a few simplifications:
- deletion vector only supports u32, instead of u64
- I only implement deletion vector store and load for filesystem catalog
  + Because iceberg-rust sdk has no support for deletion vector, I write it myself via append blob into manifest files and rewrite

Next step:
- Support u64 as row index for deletion vector
- object storage catalog is also needed when we release pg_mooncake v0.2, so need to write the corresponding logic in `update_table` as well, and integrate into snapshot iceberg part
- Need to double check and potentially fix sequential number (and other key fields) related logic
  + only deletion vector related ones are supported
- Confirm with query side about the integration interface
  + As of now, data files, deletion vector and all metadata files are stored together, likely we need to split them into different stages
- (minor) When we access an iceberg table, we require namespace / table information, now it's hard-coded to be 

TODO items (not related to functionality):
- A lot of store and load IO operations could be parallelized
- Add metrics for all IO operations
- Add validation to manifest / metadata precondition
- Implement property related logic; there're a few must-have fields in spec

## Checklist

- [x] Code builds correctly

### DIFF
--- a/src/moonlink/src/storage/iceberg/deletion_vector.rs
+++ b/src/moonlink/src/storage/iceberg/deletion_vector.rs
@@ -91,6 +91,7 @@ impl DeletionVector {
     ///
     /// Serialization storage format:
     /// | len for magic and vector | magic | vector | crc32c |
+    /// crc32c field is checksum of the magic bytes and serialized vector as 4 bytes in big-endian.
     pub fn serialize(&self, properties: HashMap<String, String>) -> Blob {
         DeletionVector::check_properties(&properties);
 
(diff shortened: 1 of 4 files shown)


src/moonlink/src/storage/iceberg/validation.rs
```diff
@@ -0,0 +1,68 @@
+use iceberg::spec::{DataContentType, DataFileFormat, ManifestEntry};
+/// This module validates iceberg specification.
+use iceberg::Result as IcebergResult;
+
+/// Validate invariants for the given puffin manifest, if precondition detected broken return error.
+pub(crate) fn validate_puffin_manifest_entry(entry: &ManifestEntry) -> IcebergResult<()> {
```
dentiny, line 6: It's hard to add unit test because of the struct visibility :(

PR #78, opened 2025-05-02
[storage] [9/N] Add deletion vector store for object storage catalog

dentiny:
## Summary

This PR does a few things:
- Add deletion vector store and load for object storage catalog, and add unit test
- Remove rest catalog from out snapshot-iceberg code, because it doesn't support deletion vector related features
  + In other words, we only support filesystem catalog and object storage catalog
  + rest catalog is still useful (i.e, talk to AWS iceberg service), but that requires ourselves to do either wrapping or source code modification
- Fix `FileIO` creation, which requires us to configure secret id, secret key, bucket again; just as what we do for opendal
- Fix flaky test, which is caused by the fact that rust runs unit tests in parallel, so we need to randomize minio bucket name and warehouse uri

Something iceberg side requires from client-side:
- warehouse uri: where to store tables
- namespace and table name
- secret id and key required for warehouse access

## Checklist

- [x] Code builds correctly

### DIFF
--- a/src/moonlink/src/storage/iceberg/test_utils.rs
+++ b/src/moonlink/src/storage/iceberg/test_utils.rs
@@ -107,6 +107,7 @@ async fn create_test_s3_bucket_impl(bucket: Arc<String>) -> IcebergResult<()> {
     Ok(())
 }
 
+/// Creates the provided bucket with exponential backoff retry; this function assumes the bucket doesn't exist, otherwise it will return error.
 #[allow(dead_code)]
 pub(crate) async fn create_test_s3_bucket(bucket: String) -> IcebergResult<()> {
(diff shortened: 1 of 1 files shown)


src/moonlink/src/storage/iceberg/iceberg_snapshot.rs
```diff
@@ -70,7 +72,8 @@ async fn get_or_create_iceberg_table<C: Catalog + ?Sized>(
             let tbl_creation = TableCreation::builder()
                 .name(table_name.to_string())
                 .location(format!(
-                    "file:///tmp/iceberg-test/{}/{}",
+                    "{}/{}/{}",
```
dentiny, line 75: This is a bug, metadata location is used to decide generated file location.
It wasn't discovered because rest catalog acts as a proxy, and allows warehouse uri to be local filesystem.

Issue #80, opened 2025-05-02
iceberg: revisit iceberg related dependency


dentiny:
## Summary

As of now (May.2, 2025) moonlink takes ~10 seconds to link;
we should be able to prune iceberg related dependencies.

PR #82, opened 2025-05-02
[storage] Dependency cleanup for iceberg related packages

dentiny:
## Summary

Dependency cleanup:
- Remove currently unused rest catalog, we could add it back when needed
  + Also possible to make a catalog registry
- Move test related packages into `[dev-dependencies]`, which is only used in testing

## Related Issues

Closes https://github.com/Mooncake-Labs/moonlink/issues/80

## Checklist

- [x] Code builds correctly

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,9 +1,5 @@
 [workspace]
-members = [
-    "src/moonlink",
-    "src/moonlink_connectors",
-    "src/moonlink_backend",
-]
+members = ["src/moonlink", "src/moonlink_connectors", "src/moonlink_backend"]
 resolver = "2"
 
 [workspace.package]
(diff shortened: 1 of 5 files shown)


Cargo.toml
```diff
@@ -1,9 +1,5 @@
 [workspace]
-members = [
-    "src/moonlink",
-    "src/moonlink_connectors",
-    "src/moonlink_backend",
-]
+members = ["src/moonlink", "src/moonlink_connectors", "src/moonlink_backend"]
```
dentiny, line 2: changes made by `taplo-cli`.

src/moonlink/src/storage/iceberg/test_utils.rs
```diff
@@ -1,16 +1,7 @@
 /// This module provides a few test util functions.
 use crate::storage::iceberg::object_storage_catalog::{S3Catalog, S3CatalogConfig};
 
-use std::sync::Arc;
-
-use aws_sdk_s3::config::{Credentials, Region};
```
dentiny, line 6: Main change here: move aws-s3 and tokio-retry related packages into test guard, and only used in unit tests.

PR #83, opened 2025-05-03
Optimize moonlink dependency

dentiny:
## Summary

This PR did a few things to optimize dependency:
- Use `default-features` to disable default features explicitly, otherwise they will be included
  + Reference: https://doc.rust-lang.org/cargo/reference/features.html
- Use minor version instead of patch version as dependency

## Checklist

- [x] Code builds correctly


dentiny:
I don't see dependency number before and after the change. @dpxcc 
```sh
[/workspaces/pg_moonlink] (hjiang/prune-dependency) 
vscode@b09be3cd0f70$ cargo tree 2>&1 | wc -l
1234
```

PR #85, opened 2025-05-03
[storage] Fix synchronous filesystem operations

dentiny:
## Summary

This PR fixes a severe performance-critical bug:
- I'm mixing synchronous filesystem access within async tokio-based runtime
- Update with opendal-native async functions

## Checklist

- [x] Code builds correctly

### DIFF
--- a/src/moonlink/src/storage/iceberg/file_catalog.rs
+++ b/src/moonlink/src/storage/iceberg/file_catalog.rs
@@ -916,7 +916,15 @@ mod tests {
             Path::new(&version_hint_filepath).exists(),
             "{version_hint_filepath} doesn't exist"
         ); // created at table creation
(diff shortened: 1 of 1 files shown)


src/moonlink/src/storage/iceberg/file_catalog.rs
```diff
@@ -57,13 +85,20 @@ pub struct FileSystemCatalog {
 impl FileSystemCatalog {
     /// Creates a rest catalog from config, if the given warehouse location doesn't exist, it will be created.
     pub fn new(warehouse_location: String) -> Self {
-        std::fs::create_dir_all(warehouse_location.clone()).unwrap();
+        // opendal prepends root directory to all paths for all operations, here we do path manipulation ourselves and set root directory as filesystem root.
+        let builder = Fs::default().root("/");
```
dentiny, line 89: I think it's slightly better to let opendal to handle warehouse location, but keep the old logic to reduce code change.

PR #87, opened 2025-05-04
[storage] Fix deletion vector sequence number consistency

dentiny:
## Summary

Under an iceberg snapshot, all files (i.e. data files, deletion vectors) should have the same sequence number.
This PR also updates tokio retry package, which is better maintained and supports more feature;
one of them is necessary for iceberg spec: control the overall timeout for retry.

## Checklist

- [x] Code builds correctly

### DIFF
--- a/src/moonlink/src/storage/iceberg/iceberg_snapshot.rs
+++ b/src/moonlink/src/storage/iceberg/iceberg_snapshot.rs
@@ -560,9 +560,6 @@ mod tests {
     async fn test_store_and_load_snapshot_with_filesystem_catalog() -> IcebergResult<()> {
         let tmp_dir = tempdir()?;
         let warehouse_path = tmp_dir.path().to_str().unwrap();
-
-        // let warehouse_path = "/tmp/iceberg--2h2-3-sjsssj";
-
         let catalog = create_catalog(warehouse_path)?;
         test_store_and_load_snapshot_impl(catalog, warehouse_path).await?;
         Ok(())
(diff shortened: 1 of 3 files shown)


src/moonlink/src/storage/iceberg/iceberg_snapshot.rs
```diff
@@ -134,6 +133,19 @@ async fn write_record_batch_to_iceberg(
     Ok(data_files[0].clone())
 }
 
+// TODO(hjiang): A few data file properties need to respect and consider.
+// Reference:
+// - https://iceberg.apache.org/docs/latest/configuration/#table-properties
+// - https://iceberg.apache.org/docs/latest/configuration/#table-behavior-properties
+//
+// - write.parquet.row-group-size-bytes
```
dentiny, line 141: I checked iceberg spec these days, two types of properties we might be interested in:
- parquet data file related, listed here
- transaction commit retry behavior

src/moonlink/src/storage/iceberg/tokio_retry_utils.rs
```diff
@@ -0,0 +1,55 @@
+/// This module contains tokio retry related util functions.
+use iceberg::Error as IcebergError;
+use tokio_retry2::RetryError as TokioRetryError;
```
dentiny, line 3: iceberg has specification on retry behavior: https://iceberg.apache.org/docs/latest/configuration/#table-behavior-properties
I'm thinking to implement with the tokio-retry library.

PR #115, opened 2025-05-08
[storage] Initial integration with iceberg and mooncake table

dentiny:
## Summary

This PR does a few things:
- (major) Integrate iceberg snapshot commit with mooncake table
- (major) Fix deletion vector write bug, see inline comments for details
  + Deletion vector manifest is stored separately from data file manifest
  + So data file manifest files are not touched, deletion vectors are merged and rewritten
- (minor) Attach metadata to arrow schema, otherwise [conversion](https://github.com/apache/iceberg-rust/blob/513376cb04b899edf457918c3852a786a292071c/crates/iceberg/src/arrow/schema.rs#L215-L223) to iceberg schema fail

## Iceberg snapshot persistence decisions made

**Goals**
- We don’t do IO operations too frequently
  + Only flush when threshold exceeds
- Commit iceberg table transaction in a consistent state

**Plan**
Write data file and deletion vector to iceberg table after snapshot update, and only when we have new local data files created
- Two things here:
  + As of now, we still write to local filesystem first, then upload to remote storage; later on, we could write to both backends concurrently
  + Iceberg transaction tracks delta (i.e. new data files, new puffin files, etc); we store old table content (maps from <data file, deletion vector> to compare with current snapshot), and only append delta to iceberg transaction

Verbose documentation: https://docs.google.com/document/d/1kpZblgMdVfZw9M9Eygtx9rTacDFQBhPMvMBg1LboYkk/edit?usp=sharing

## Deletion vector write proposal

https://docs.google.com/document/d/1fIvrRfEHWBephsX0Br2G-Ils_30JIkmGkcdbFbovQjI/edit?usp=sharing

## Related Issues

## Checklist

- [ ] Code builds correctly

### DIFF
--- a/src/moonlink/src/storage/iceberg/iceberg_table_manager.rs
+++ b/src/moonlink/src/storage/iceberg/iceberg_table_manager.rs
@@ -600,7 +600,7 @@ mod tests {
         // Perform a few table write operations.
         //
         // Operation series 1: append three rows, delete one of them, flush, commit and create snapshot.
-        // Expects to see one data file with no deletion vector, because mooncake table handle deletion inline before persistence.
+        // Expects to see one data file with no deletion vector, because mooncake table handle deletion inline before persistence, and all record batches are dumped into one single data file.
         // The three rows are deleted in three operations series respectively.
         let row1 = MoonlinkRow::new(vec![
             RowValue::Int32(1),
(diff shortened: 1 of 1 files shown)


src/moonlink/src/storage/iceberg/iceberg_snapshot.rs
```diff
@@ -250,11 +249,6 @@ impl IcebergSnapshot for Snapshot {
         let mut disk_file_to_deletion_vector: HashMap<PathBuf, BatchDeletionVector> =
             HashMap::new();
         for manifest_file in manifest_list.entries().iter() {
-            // We're only interested in DATA files when recover snapshot status.
```
dentiny, line 253: This is a bug, deletion vector manifest files should have type `Deletes`
https://github.com/apache/iceberg-rust/blob/513376cb04b899edf457918c3852a786a292071c/crates/iceberg/src/spec/manifest_list.rs#L604-L611

Issue #142, opened 2025-05-09
Index: implement async BitWriter


dentiny:
## Summary

`tokio` doesn't support async `BitWriter`, I raise a github issue here: https://github.com/tokio-rs/tokio/issues/7325

I don't think it's too hard to implement, basically some endian-related logic above internal writer should be enough; I will try to write our own version and upstream.

dentiny:
Requirement for `IndexBlock`:
- Write and creation should be async, so it doesn't block table handler's write path;
- Read could be sync (via mmap), which lives in search path.

Issue #141, opened 2025-05-09
Index: use async IO for index files creation, write, flush and close


dentiny:
## Summary

Index write and read all happens inside of tokio thread, which requires to be async otherwise it blocks executor.

PR #143, opened 2025-05-10
[index] Adapt index block IO to async

dentiny:
## Summary

Index block creation and write happens in table handler tokio eventloop, which shouldn't have sync IOs.

## Related Issues

Closes https://github.com/Mooncake-Labs/moonlink/issues/142
Closes https://github.com/Mooncake-Labs/moonlink/issues/141

## Checklist

- [x] Code builds correctly

### DIFF
--- a/src/moonlink/src/storage/index/persisted_bucket_hash_map.rs
+++ b/src/moonlink/src/storage/index/persisted_bucket_hash_map.rs
@@ -260,10 +260,11 @@ impl IndexBlockBuilder {
         }
         let bucket_start_offset = (self.current_entry as u64)
             * (metadata.hash_lower_bits + metadata.seg_id_bits + metadata.row_id_bits) as u64;
-        for bucket in self.buckets.clone() {
+        let buckets = std::mem::take(&mut self.buckets);
+        for cur_bucket in buckets {
             self.get_entry_writer()
                 .await
-                .write(metadata.bucket_bits, bucket)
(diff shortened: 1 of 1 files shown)


src/moonlink/src/storage/index/persisted_bucket_hash_map.rs
```diff
@@ -217,40 +233,48 @@ impl IndexBlockBuilder {
...
+        for bucket in self.buckets.clone() {
+            self.get_entry_writer()
+                .await
+                .write(metadata.bucket_bits, bucket)
+                .await
                 .unwrap();
         }
-        self.entry_writer.byte_align().unwrap();
+        self.get_entry_writer().await.byte_align().await.unwrap();
+        self.get_entry_writer().await.flush().await.unwrap();
```
dentiny, line 272: If an error happens at flush or close, `drop` silently proceed
Reference: https://doc.rust-lang.org/std/fs/struct.File.html#method.sync_all

src/moonlink/Cargo.toml
```diff
@@ -24,6 +24,7 @@ roaring = { workspace = true }
 serde_json = { workspace = true }
 thiserror = { workspace = true }
 tokio = { workspace = true }
+tokio-bitstream-io = { workspace = true }
```
dentiny, line 27: Pretty lightweight dependency; implementation-wise it's basically the same as bitstream-io apart from using async writer internally.

PR #155, opened 2025-05-13
Upgrade iceberg rust

dentiny:
## Summary

Sync upstream mainly for two changes:
- Declare `FileRead` trait to be `Sync`-safe, so we won't have blocking IO operation at eventloop: https://github.com/apache/iceberg-rust/pull/1319
- Dependency upgrade: https://github.com/apache/iceberg-rust/pull/1294

## Checklist

- [x] Code builds correctly


src/moonlink/src/storage/mooncake_table/snapshot.rs
```diff
@@ -63,7 +62,7 @@ impl SnapshotTableState {
         let mut snapshot = Snapshot::new(metadata);
         if iceberg_table_config.is_some() {
             let mut table_manager = IcebergTableManager::new(iceberg_table_config.unwrap());
-            snapshot = block_on(table_manager.load_snapshot_from_table()).unwrap();
+            snapshot = table_manager.load_snapshot_from_table().await.unwrap();
```
dentiny, line 65: This is the most important change, which loads snapshot from iceberg in an async style without blocking eventloop.

PR #157, opened 2025-05-13
[backend] Moonlink back take a directory as base path

dentiny:
## Summary

[PR](https://github.com/Mooncake-Labs/moonlink/pull/156) breaks the interface with pg_mooncake;
this PR tries to save it by taking a passed directory from pg_moocake, and used it to hold all tables data.

## Checklist

- [x] Code builds correctly

### DIFF
--- a/src/moonlink_backend/src/lib.rs
+++ b/src/moonlink_backend/src/lib.rs
@@ -25,9 +25,13 @@ impl<T: Eq + Hash> Default for MoonlinkBackend<T> {
 }
 
 impl<T: Eq + Hash> MoonlinkBackend<T> {
-    pub fn new(moonlink_table_base_path: String) -> Self {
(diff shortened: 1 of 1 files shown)


src/moonlink_backend/src/lib.rs
```diff
@@ -9,46 +9,46 @@ use std::sync::Arc;
 use std::{collections::HashMap, hash::Hash};
 use tokio::sync::RwLock;
 
+// Default local filesystem directory where all tables data will be stored under.
+const DEFAULT_MOONLINK_TABLE_BATH_PATH: &str = "./mooncake_test/";
```
dentiny, line 13: Default value comes from https://github.com/Mooncake-Labs/moonlink/blob/d766842f36a12094bbb7c718b13ff0246b3c56dd/src/moonlink_connectors/src/postgres.rs#L73

PR #167, opened 2025-05-15
[storage] Fix snapshot dump

dentiny:
## Summary

This PR is mostly bug fixing PR:
- We shouldn't create empty hash index puffin files;
- We should use committed deletion log for persistence, instead of disk-files ones.

An error cases for second fix:
- In txn-1, data file 1 and its deletion vector gets persisted into disk;
- In txn-2, we add new rows, and delete rows in data file 1, then commit
  + At mooncake table snapshot creation, new arrow batches still sits in memory, because we only flush to disk when overall size reaches threshold;
  + but deleted rows have been reflected to data file 1's deletion vector (in `disk_files` map)
  + There's inconsistency (from LSN's perspective) between data file and deletion vector

The solution here (fix in this PR), is to 
- generate iceberg snapshot based on persisted data files
- use flush LSN and committed deletion log as source of truth
- index file is not a problem, because it goes with data files (flush at the same time)

Some other changes:
- Rename deletion vector and hash index manifest file to better understand

## TODO items

- Reduce iceberg snapshot frequency, currently it's created every mooncake table snapshot creation
- Add state-based unit tests: https://docs.google.com/document/d/1hZ0H66_eefjFezFQf1Pdr-A63eJ94OH9TsInNS-6xao/edit?usp=sharing

## Checklist

- [x] Code builds correctly

### DIFF
--- a/src/moonlink/src/storage/iceberg/iceberg_table_manager.rs
+++ b/src/moonlink/src/storage/iceberg/iceberg_table_manager.rs
@@ -11,8 +11,6 @@ use crate::storage::index::{FileIndex as MooncakeFileIndex, MooncakeIndex};
 use crate::storage::mooncake_table::delete_vector::BatchDeletionVector;
 use crate::storage::mooncake_table::Snapshot as MooncakeSnapshot;
 use crate::storage::mooncake_table::TableMetadata as MooncakeTableMetadata;
-use crate::storage::storage_utils;
-use crate::storage::storage_utils::ProcessedDeletionRecord;
 
 use std::collections::{HashMap, HashSet};
 use std::path::PathBuf;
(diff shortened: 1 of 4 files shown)


src/moonlink/src/storage/iceberg/iceberg_table_manager.rs
```diff
@@ -373,11 +384,11 @@ impl IcebergTableManager {
     /// TODO(hjiang): Need to configure (1) the number of blobs in a puffin file; and (2) the number of file index in a puffin blob.
     /// For implementation simpicity, put everything in a single file and a single blob.
     async fn sync_file_indices(&mut self, file_indices: &[MooncakeFileIndex]) -> IcebergResult<()> {
-        if file_indices.is_empty() {
+        if file_indices.len() == self.persisted_file_index_ids.len() {
```
dentiny, line 387: If hash index files don't change between two iceberg snapshots, we shouldn't dump index files.

PR #175, opened 2025-05-16
[storage] Fetch mooncake table metadata to pg_mooncake

dentiny:
## Summary

This PR gets `TableMetadata`, which represents data files and their deletion vectors for queries, to pg_mooncake.
A few included items:
- Prune committed deletion logs based on flush LSN, so we don't suffer unconstrained memory consumption;
- Mooncake table snapshot include both persisted deletion vector blob and batch deletion vector;
- Updated encoding and decoding for table metadata, which is passed to and used at pg_mooncake.

## Checklist

- [x] Code builds correctly

### DIFF
--- a/src/moonlink/src/storage/iceberg/tests.rs
+++ b/src/moonlink/src/storage/iceberg/tests.rs
@@ -814,11 +814,17 @@ async fn check_prev_data_files(
     assert_eq!(loaded_arrow_batch, expected_arrow_batch);
 
     // In the test suite, we only delete the second prepared row.
-    assert!(!deletion_vector.batch_deletion_vector.is_deleted(/*row_idx=*/ 0));
+    assert!(!deletion_vector
+        .batch_deletion_vector
+        .is_deleted(/*row_idx=*/ 0));
     if deleted {
-        assert!(deletion_vector.batch_deletion_vector.is_deleted(/*row_idx=*/ 1));
+        assert!(deletion_vector
+            .batch_deletion_vector
+            .is_deleted(/*row_idx=*/ 1));
     } else {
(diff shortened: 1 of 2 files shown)


src/moonlink/src/storage/iceberg/puffin_utils.rs
```diff
@@ -4,9 +4,19 @@ use std::collections::HashMap;
...
 use iceberg::puffin::{Blob, PuffinReader};
-use iceberg::spec::DataFile;
 use iceberg::{Error as IcebergError, Result as IcebergResult};
 
+/// Reference to puffin blob.
+#[derive(Clone, Debug, Eq, PartialEq)]
+pub(crate) struct PuffinBlobRef {
+    /// Path for the puffin file.
+    pub(crate) puffin_filepath: String,
+    /// Start offset for the blob.
```
dentiny, line 14: Currently we only store one blob in the puffin file, so start offset is always 4; but keep start offset for possible future change.

src/moonlink/src/union_read/table_metadata.rs
```diff
@@ -41,6 +73,8 @@ fn write_usize<W: Writer>(writer: &mut W, value: usize) -> Result<(), EncodeErro
 #[cfg(test)]
 impl TableMetadata {
     pub fn decode(data: &[u8]) -> Self {
+        use crate::storage::mooncake_table::PuffinDeletionBlobAtRead;
```
dentiny, line 76: I will convert the function into `Decode` trait implementation in the followup PR, it's not a blocking issue.

dentiny:
Notice, this is a breaking change, which changes the interface between moonlink and pg_mooncake.
@dpxcc will followup with the pg_mooncake side change, merge the PR for now.

PR #184, opened 2025-05-18
Improve test speed by waiting for data changes

nbiscaro:
## Summary
- speed up `test_moonlink_service` by polling the table until the data updates or a timeout occurs

## Testing
- `cargo test -p moonlink_backend --offline` *(fails: failed to get `iceberg` as a dependency)*

### DIFF
--- a/src/moonlink_backend/src/lib.rs
+++ b/src/moonlink_backend/src/lib.rs
@@ -109,9 +109,18 @@ mod tests {
             .await
             .unwrap();
         let old = service.scan_table(&"test", None).await.unwrap();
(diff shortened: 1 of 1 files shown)


src/moonlink_backend/src/lib.rs
```diff
@@ -109,9 +109,18 @@ mod tests {
             .await
             .unwrap();
         let old = service.scan_table(&"test", None).await.unwrap();
```
dpxcc, line 111: A better fix is to get the current LSN and pass it to service instead of None

Issue #173, opened 2025-05-15
iceberg: parquet parameter re-check before release


dentiny:
## Summary

iceberg spec has clear requirement on parquet file parameters, for example, row group size, codec;
these parameters have their default value, or should be persisted somewhere and accessible when visiting iceberg table,
otherwise query engines don't know how to parse data.

I left TODOs for some key metadata, will revisit before release:
https://github.com/Mooncake-Labs/moonlink/blob/b9fbfc07d64cbc8e6b6649b0d94f6233bcac99e2/src/moonlink/src/storage/iceberg/utils.rs#L161-L173

PR #186, opened 2025-05-18
[storage] Fix iceberg table compression property

dentiny:
## Summary

Iceberg table spec clarifies, by default data files are compression with zstd, which we don't.
This PR sets proper table property so external iceberg table reader could parse it out.
Checked a few other properties listed in the spec (https://iceberg.apache.org/docs/latest/configuration/#table-properties), don't think they affect functionality.

We currently have two parquet data files involved: one for write through cache, another for iceberg data file upload;
If we plan to compress, the ideal situation would be:
- Create write through cache with iceberg data file writer directly and get the `DataFile` struct;
- Copy the file to iceberg table, instead of reading it out then write again.

## Related Issues

Closes https://github.com/Mooncake-Labs/moonlink/issues/173

## Checklist

- [x] Code builds correctly

### DIFF
--- a/src/moonlink/src/storage/iceberg.rs
+++ b/src/moonlink/src/storage/iceberg.rs
@@ -5,6 +5,7 @@ pub(crate) mod index;
 pub(crate) mod moonlink_catalog;
 pub(crate) mod puffin_utils;
 pub(crate) mod puffin_writer_proxy;
+mod table_property;
 pub(crate) mod test_utils;
 pub(crate) mod tokio_retry_utils;
 pub(crate) mod utils;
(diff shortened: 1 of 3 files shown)


zhang-wenchao:
Sorry for commenting here. I have a question. Does this mean that Parquet files are now not compressed? Will this significantly increase storage pressure and slow down queries?

dentiny:
> Sorry for commenting here. I have a question. Does this mean that Parquet files are now not compressed? Will this significantly increase storage pressure and slow down queries?

Hi @zhang-wenchao thanks for the interest and ask.
Yes we don't compress parquet for now, both on write through cache and iceberg table storage, but we plan to enable before v0.2 official release.

dentiny:
> Sorry for commenting here. I have a question. Does this mean that Parquet files are now not compressed? Will this significantly increase storage pressure and slow down queries?

Hi @zhang-wenchao , I just enable compression for data file with ZSTD level 5.
PR for reference: https://github.com/Mooncake-Labs/moonlink/pull/464
Let me know if you have any further questions / feature requests!

Issue #178, opened 2025-05-16
iceberg: provide snapshot creation interface in moonlink backend


dentiny:
## Summary

Provide an interface in moonlink backend to create an iceberg snapshot for easier testing.

PR #193, opened 2025-05-20
[backend] Provide an interface to create adhoc iceberg snapshot

dentiny:
## Summary

As titled, this PR provides an additional interface at moonlink backend to create an iceberg snapshot (if possible).
The core idea and principle to is leverage table handler's eventloop and channel as the threading model, so we don't need to deal with lock issues ourselves.

## Related Issues

Closes https://github.com/Mooncake-Labs/moonlink/issues/178

## Checklist

- [x] Code builds correctly

### DIFF
--- a/src/moonlink/src/table_handler.rs
+++ b/src/moonlink/src/table_handler.rs
@@ -153,12 +153,14 @@ impl TableHandler {
                         }
                     }
                 }
-                // wait for iceberg snapshot requests.
+                // wait for snapshot requests.
                 Some(()) = iceberg_snapshot_initiation_receiver.recv() => {
                     assert!(!has_outstanding_iceberg_snapshot_request, "There should be at most one outstanding iceberg snapshot request for one table!");
-                    has_outstanding_iceberg_snapshot_request = true;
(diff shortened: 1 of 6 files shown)


src/moonlink/src/storage/iceberg/iceberg_snapshot_state_manager.rs
```diff
@@ -0,0 +1,53 @@
+/// This module interacts with iceberg snapshot status.
+use tokio::sync::mpsc;
+
+/// At most one outstanding snapshot request is allowed.
+pub struct IcebergSnapshotStateManager {
```
dentiny, line 5: At the moment only initiate snapshot and sync is supported, in the future could use as a sink for async snapshot event.

src/moonlink/src/table_handler.rs
```diff
@@ -131,12 +153,25 @@ impl TableHandler {
                         }
                     }
                 }
+                // wait for iceberg snapshot requests.
```
dentiny: The key part for this PR, is to leverage existing TableHandler's eventloop to interact with snapshot, otherwise we need to resolve threading issue.

PR #198, opened 2025-05-20
[index] Make all index lookup function async

dentiny:
## Summary

Use async for all index-related IOs (mainly `AsyncBitReader`).

## Checklist

- [x] Code builds correctly

### DIFF
--- a/src/moonlink/benches/microbench_index_stress.rs
+++ b/src/moonlink/benches/microbench_index_stress.rs
@@ -21,7 +21,8 @@ fn bench_index_stress(c: &mut Criterion) {
     group.bench_function("search_10m_entries", |b| {
         b.iter(|| {
             let mut rng = rand::rng();
-            let result = black_box(index.search(&(rng.random_range(0..10000000) as u64)));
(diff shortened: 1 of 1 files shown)


src/moonlink/src/storage/index/persisted_bucket_hash_map.rs
```diff
@@ -397,20 +390,15 @@ impl<'a> IndexBlockIterator<'a> {
         metadata: &'a GlobalIndex,
         file_id_remap: &'a Vec<u32>,
     ) -> Self {
-        let mut bucket_reader = BitReader::endian(
+        let mut bucket_reader = AsyncBitReader::endian(
             Cursor::new(collection.data.as_ref().unwrap().as_ref()),
-            BigEndian,
+            AsyncBigEndian,
```
dentiny, line 395: iterator and debug / fmt related function requires to be sync, or we change to use [`AsyncIterator`](https://doc.rust-lang.org/std/async_iter/trait.AsyncIterator.html)

Issue #207, opened 2025-05-21
[index] Avoid sync IO for index iterators


dentiny:
## Summary

Code pointer:
https://github.com/Mooncake-Labs/moonlink/blob/3afbf9a24580c4038713aaec13437581c920bc3d/src/moonlink/src/storage/index/persisted_bucket_hash_map.rs#L372-L545

PR #208, opened 2025-05-21
[index] Replace index block iterator with async next

dentiny:
## Summary

This PR updates index iterator's blocked IO to async style.
I don't think I'm doing idiomatic rust async iterator, which is supposed to be `Stream`, but found it doesn't support internal async calls directly.
Eg: https://www.qovery.com/blog/a-guided-tour-of-streams-in-rust/ 

## Related Issues

Closes https://github.com/Mooncake-Labs/moonlink/issues/207

## Checklist

- [x] Code builds correctly

### DIFF
--- a/src/moonlink/src/storage/mooncake_table/test_utils.rs
+++ b/src/moonlink/src/storage/mooncake_table/test_utils.rs
@@ -6,7 +6,6 @@ use crate::storage::iceberg::puffin_utils;
 use crate::storage::mooncake_table::snapshot::PuffinDeletionBlobAtRead;
 use arrow::array::Int32Array;
 use arrow::datatypes::{DataType, Field};
-use futures::executor::block_on;
 use futures::future::join_all;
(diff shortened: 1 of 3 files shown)


src/moonlink/src/storage/index/persisted_bucket_hash_map.rs
```diff
@@ -311,6 +317,41 @@ impl GlobalIndexBuilder {
...
+            hash_lower_bits: lower_bits,
+            seg_id_bits,
+            row_id_bits: 32,
+            bucket_bits,
+            index_blocks: vec![],
+        };
+        let mut index_blocks = Vec::new();
+        let mut index_block_builder =
+            IndexBlockBuilder::new(0, num_buckets + 1, self.directory.clone());
+        for entry in iter {
```
dentiny, line 348: Iteration through iteration is destined to be different for container (sync) and merging iterator (async)

Issue #209, opened 2025-05-21
iceberg: provide option to drop snapshot on table creation


dentiny:
## Summary

```sql
DROP EXTENSION pg_mooncake CASCADE;
DROP TABLE r; 
CREATE EXTENSION pg_mooncake; 
CREATE TABLE r (a int primary key, b text); 
CALL create_mooncake_table('c', 'r'); 
INSERT INTO r SELECT gs AS a, 'val_' || gs AS b FROM generate_series(1, 100) AS gs; 
SELECT count(*) AS total_rows FROM c;
```
On columnstore table creation, we shouldn't recovery previous state from iceberg;
instead, everything should be dropped.

My idea is to provide an option, to specify whether we want to recover or drop;
based on the table manipulation function called, whether it's `create_table` or `get_table`.

PR #217, opened 2025-05-21
[storage] Provide an option to drop iceberg table at creation

dentiny:
## Summary

Current behavior for pg_mooncake: 
```sql
pg_mooncake (pid: 7060) =# DROP EXTENSION pg_mooncake CASCADE;
DROP TABLE r;                                                                                                                                       
CREATE EXTENSION pg_mooncake; 
INSERT INTO r (a, b) 
SELECT gs AS a, 'val_' || gs AS b FROM generate_series( (SELECT COALESCE(MAX(a), 0) + 1 FROM r), (SELECT COALESCE(MAX(a), 0) + 100 FROM r) ) AS gs; 
SELECT count(*) AS total_rows FROM c;
DROP EXTENSION
DROP TABLE
CREATE EXTENSION
CREATE TABLE
CALL
INSERT 0 100
 total_rows 
------------
        300
(1 row)
```
which clearly display when we create a new columnstore table, iceberg loads and recovers previous snapshot state.
At the initial release we don't support recovery, so provide an option to drop existing table and by default drop for sink.

How I tested:
- Unit test to make sure we're able to drop table at iceberg table side
- Close psql and re-connect to pg server, confirm reconnect won't load previous states
```sql
pg_mooncake (pid: 31031) =# DROP EXTENSION pg_mooncake CASCADE;
DROP TABLE r;                                                                                                                                       
CREATE EXTENSION pg_mooncake; 
INSERT INTO r (a, b) 
SELECT gs AS a, 'val_' || gs AS b FROM generate_series( (SELECT COALESCE(MAX(a), 0) + 1 FROM r), (SELECT COALESCE(MAX(a), 0) + 100 FROM r) ) AS gs; 
SELECT count(*) AS total_rows FROM c;
DROP EXTENSION
DROP TABLE
CREATE EXTENSION
CREATE TABLE
CALL
INSERT 0 100
 total_rows 
------------
        100
(1 row)
``` 

## Related Issues

Closes https://github.com/Mooncake-Labs/moonlink/issues/209

## Checklist

- [x] Code builds correctly


src/moonlink/src/storage/iceberg/iceberg_table_manager.rs
```diff
@@ -48,6 +48,9 @@ pub struct IcebergTableConfig {
     /// Iceberg table name.
     #[builder(default = "table".to_string())]
     pub table_name: String,
+    /// Whether to drop the old table at creation.
```
dentiny, line 51: It's a workaround before we implement `drop_table`, which drops iceberg table

PR #226, opened 2025-05-22
avoid using pathbuf to identify record

zhousun:
## Summary

## Related Issues

## Changes

- 
- 
- 

## Checklist

- [ ] Code builds correctly


src/moonlink/src/storage/iceberg/iceberg_table_manager.rs
```diff
@@ -280,18 +280,23 @@ impl IcebergTableManager {
             } else {
                 0
             };
-
+        let mut file_name_to_id = HashMap::new();
```
dentiny, line 283: I don't think it's used anywhere?

zhousun: good catch. yea it's part of the undone Update file id in persisted_file_indices

Issue #213, opened 2025-05-21
iceberg: force to create snapshot when explicitly requested


dentiny:
## Summary

When receive request to snapshot, force to flush all in-memory states and into mooncake table snapshot, and create an iceberg snapshot.
https://github.com/Mooncake-Labs/moonlink/blob/cb66822a42abeb80a1bb7c89040a9561f5d80428/src/moonlink/src/table_handler.rs#L156-L166

dentiny:
The difficulty lies in:
- When requested to create a snapshot, it's possible to have periodic snapshot event ongoing
- Periodic snapshot doesn't guarantee new iceberg snapshot will be generated

PR #227, opened 2025-05-22
[storage] Add iceberg snapshot logic into table handler eventloop

dentiny:
## Summary

A few things in this PR:
- Put force snapshot request as a table event, instead of a separate future
  + tokio select statement evaluates all futures concurrently, but we need to make sure snapshot happens after commit
  + A typical use case is, insert a bunch of rows and request to create a new snapshot
- Fix eventloop usage, which places all side-affects into statements rather than futures
  + Futures will be evaluated for multiple times, which shouldn't have any side effect
- Bug fix for re-creating snapshot task: https://github.com/Mooncake-Labs/moonlink/pull/227/files#r2103557118

How I tested:
No new test case added, manually run previously flaky tests for 50 times and all pass without issue

## Checklist

- [x] Code builds correctly

### DIFF
--- a/src/moonlink/src/table_handler.rs
+++ b/src/moonlink/src/table_handler.rs
@@ -109,14 +109,10 @@ impl TableHandler {
                         }
                         TableEvent::Commit { lsn } => {
                             table.commit(lsn);
-                            println!("commit at lsn {}", lsn);
                             if table.should_flush() {
-                                println!("flush expected");
                                 if let Err(e) = table.flush(lsn).await {
                                     println!("Flush failed in Commit: {}", e);
                                 }
-                            } else {
-                                println!("no!! flush");
                             }
                         }
                         TableEvent::StreamCommit { lsn, xact_id } => {


src/moonlink/src/storage/mooncake_table.rs
```diff
@@ -600,8 +600,10 @@ impl MooncakeTable {
         }
         self.next_snapshot_task.new_rows = Some(self.mem_slice.get_latest_rows());
         let next_snapshot_task = take(&mut self.next_snapshot_task);
+        self.next_snapshot_task = SnapshotTask::new(self.metadata.config.clone());
```
dentiny, line 603: Bug here: we need to reset snapshot task with table config.

src/moonlink/src/table_handler.rs
```diff
@@ -152,53 +137,49 @@ impl TableHandler {
...
+                                iceberg_snapshot_completion_tx.send(()).await.unwrap();
+                            } else {
+                                has_outstanding_iceberg_snapshot_request = true;
+                            }
+                        }
                     }
                 }
-                // wait for the snapshot to complete
-                Some(()) = async {
-                    if let Some(handle) = snapshot_handle.take() {
```
dentiny, line 174: Bug here: `select` statement future part shouldn't have any side affects.

src/moonlink/src/table_handler.rs
```diff
@@ -31,6 +30,8 @@ pub enum TableEvent {
     StreamFlush { xact_id: u32 },
     /// Shutdown the handler
     _Shutdown,
+    /// Force a mooncake and iceberg snapshot.
```
dentiny, line 33: I will followup https://github.com/Mooncake-Labs/moonlink/issues/213 in the next PR

PR #236, opened 2025-05-23
[storage] Implement force snapshot creation

dentiny:
## Summary

Feature request: allow adhoc iceberg snapshot creation for both testing and user experience purpose.
The difficulty lies in, at the moment, only when commit we do flushing, otherwise we don't get a consistent view of the mooncake table.

I use two ways to solve the problem:
- During commit, if we have outstanding snapshot request, which is no later than the current LSN, trigger a force flush and force snapshot;
- During periodic snapshot, if we have outstanding request and there's no more operation since last commit (table is at a consistent view), we do a force flush and commit.

Alternative considered:
- Read state manager does manual compliment upon mooncake snapshot, and manually delete the temporary files after execution
  + it's bad for our case because we need to persist files into iceberg table, thus hard to track temporary files
- Prune committed records out of in-memory data structures
  + It's design-wise better than the above solution, but hard to implement because you need to do re-mapping for all in-memory data structures

Test with unit test and SQL statement:
```sql
DROP TABLE r;
DROP EXTENSION pg_mooncake CASCADE;
CREATE EXTENSION pg_mooncake;
SELECT * FROM c;

CALL create_snapshot('c');
SELECT * FROM c;
```

## Related issues

Closes https://github.com/Mooncake-Labs/moonlink/issues/213

## Checklist

- [x] Code builds correctly

### DIFF
--- a/src/moonlink/src/table_handler.rs
+++ b/src/moonlink/src/table_handler.rs
@@ -76,21 +76,20 @@ impl TableHandler {
         let mut periodic_snapshot_interval = time::interval(Duration::from_millis(500));
 
         // Requested minimum LSN for a force snapshot request.
-        // If assigned, means there's force snapshot requested.
         let mut force_snapshot_lsn: Option<u64> = None;
 
-        // Record LSN if the last handled table event is commit, so table could be flushed safely.
-        let mut last_event_is_commit_and_commit_lsn: Option<u64> = None;
+        // Record LSN if the last handled table event is committed, which indicates mooncake table stays at a consistent view, so table could be flushed safely.
+        let mut table_consistent_view_lsn: Option<u64> = None;
 
         // Process events until the receiver is closed or a Shutdown event is received
         loop {
             tokio::select! {
                 // Process events from the queue
                 Some(event) = event_receiver.recv() => {
-                    last_event_is_commit_and_commit_lsn = match event {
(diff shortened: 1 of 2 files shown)


src/moonlink/src/storage/mooncake_table.rs
```diff
@@ -556,6 +574,8 @@ impl MooncakeTable {
     }
 
     pub async fn commit_transaction_stream(&mut self, xact_id: u32, lsn: u64) -> Result<()> {
+        self.next_snapshot_task.new_flush_lsn = Some(lsn);
```
dentiny, line 577: This is a bug, streaming write should set flush LSN also; I should've changed it in my previous PRs.

src/moonlink/src/table_handler.rs
```diff
@@ -109,13 +122,24 @@ impl TableHandler {
...
                                 if let Err(e) = table.flush(lsn).await {
                                     println!("Flush failed in Commit: {}", e);
                                 }
                             }
+                            if need_force_snapshot {
+                                snapshot_handle = table.force_create_snapshot();
+                            }
                         }
                         TableEvent::StreamCommit { lsn, xact_id } => {
+                            // TODO(hiang): Support force snapshot creation.
```
dentiny, line 141: I haven't checked streaming commit codepath yet, and no unit tests added.

Issue #43, opened 2025-04-25
[ISSUE] Create multiple table support


nbiscaro:
## Summary

Currently we only support streaming for a single, newly created table. We need to add support for adding another table to the replication stream while its already started. 

## Context

The proposed workflow for this is as follows:

pg_mooncake calls moonlink-backend.create_table will call ingest_source.add_table

ingest_source.add_table is the one starting the cdc

Of course, table creation and cdc start should be decoupled. 

## Tasks

- [ ] 

## Notes

Any extra notes or details that might be useful.

PR #246, opened 2025-05-25
Support creating multiple tables

nbiscaro:
## Summary

Allows the addition of tables to an already running replication connection. 

`event_senders` and `table_schemas` are currently shared state with the sink and stream, respectively. We need to share `event_senders` between the backend <-> sink so that we can insert event handlers for dynamically added tables. We need to share `table_schemas` with the stream so that the stream knows how to parse CDC events for dynamically added tables. 

This is somewhat unfortunate as it means we need to guard these maps with a RW lock. It is especially bad in the `table_schemas` case since we need to guard the map with a synchronous lock so we can pass into the stream. That being said, these maps are read-heavy and write-rare so I don't think these will be any source of performance bottleneck. 

Ideally, we would have no shared state between the backend and the replication connection task. A potential solution for this is to add the ability to send control messages to the stream. In this case, in addition to parsing the normal cdc stream, we can also parse arbitrary messages from the back-end, such as Add Table, Drop Table, and pass along any data we need, such as table schemas. In this case, the replication connection would own all of the state, and the back-end's job would only be to communicate control events.

The other potential solution is to build tables within the stream by parsing the `Relation` message. One problem with relying solely on the replications stream for DDL is that there's no way to drop the table.

For now, we can use shared state and return to this when we need to support more complex DLL such as `DROP TABLE`/`ALTER TABLE`. 

## Related Issues

Closes https://github.com/Mooncake-Labs/moonlink/issues/43

## Changes

- Share `event_senders` and `table_schema`s with sync and stream, respectively. 
- First, build and start the replication connection. Then, add tables to the already running connection. 
- 

## Checklist

- [x] Code builds correctly

### DIFF
--- a/src/moonlink_connectors/src/pg_replicate/clients/postgres.rs
+++ b/src/moonlink_connectors/src/pg_replicate/clients/postgres.rs
@@ -388,7 +388,7 @@ impl ReplicationClient {
         Ok(table_schemas)
     }
 
-    async fn get_table_schema(
+    pub async fn get_table_schema(
         &self,
         table_name: TableName,
         publication: Option<&str>,
--- a/src/moonlink_connectors/src/pg_replicate/table.rs
+++ b/src/moonlink_connectors/src/pg_replicate/table.rs
@@ -10,6 +10,12 @@ pub struct TableName {
 }
 
 impl TableName {
+    pub fn parse_schema_name(table_name: &str) -> (String, String) {
+        let tokens: Vec<&str> = table_name.split('.').collect();
+        let schema = tokens[0].to_string();
+        let name = tokens[1].to_string();
+        (schema, name)
+    }
     pub fn as_quoted_identifier(&self) -> String {
         let quoted_schema = quote_identifier(&self.schema);
         let quoted_name = quote_identifier(&self.name);


src/moonlink_connectors/src/pg_replicate/postgres_source.rs
```diff
@@ -250,10 +274,14 @@ impl Stream for CdcStream {
     fn poll_next(self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Option<Self::Item>> {
         let this = self.project();
         match ready!(this.stream.poll_next(cx)) {
-            Some(Ok(msg)) => match CdcEventConverter::try_from(msg, this.table_schemas) {
-                Ok(row) => Poll::Ready(Some(Ok(row))),
-                Err(e) => Poll::Ready(Some(Err(e.into()))),
-            },
+            Some(Ok(msg)) => {
+                let table_schemas = this.table_schemas.read().unwrap();
+                println!("table_schemas: {:?}", table_schemas);
```
dentiny, line 279: Wondering can we remove the logging? It's way too much and flushing the log file
```sh
[/workspaces/pg_mooncake/moonlink] (main) 
vscode@c29a7942bbad$ grep "table_schemas: " /home/vscode/.pgrx/17.log | wc -l
900119
```

nbiscaro: Yes can remove

Issue #242, opened 2025-05-24
iceberg: move snapshot persistence out of main eventloop


dentiny:
## Summary

Current IO operation blocks main eventloop:
https://github.com/Mooncake-Labs/moonlink/blob/d57a2195212cef5305a4be5087b7eb1453595d61/src/moonlink/src/table_handler.rs#L198-L210

Consideration:
- Whether it's necessary to make mooncake and iceberg snapshot invocation 1-1
- Whether it's safe to generate iceberg payload multiple times from mooncake snapshot

PR #248, opened 2025-05-25
[storage] Move iceberg snapshot out of main eventloop

dentiny:
## Summary

This PR resolves two problems:
- Move iceberg snapshot out of table handler eventloop, which blocks all concurrent events handling and affect system performance;
- Fix a potential issue for batched iceberg snapshot creation (see inline comments for details).

How I tested:
- Unit tests for moonlink
- SQL test from pg_mooncake
```sql
pg_mooncake (pid: 44455) =# DROP TABLE IF EXISTS r;
DROP EXTENSION pg_mooncake CASCADE;
CREATE EXTENSION pg_mooncake;
CALL create_snapshot('c');
SELECT * FROM c;
DROP TABLE
DROP EXTENSION
CREATE EXTENSION
CREATE TABLE
CALL
INSERT 0 3
CALL
 a | b 
---+---
 1 | a
 2 | b
 3 | c
(3 rows)

pg_mooncake (pid: 44455) =# UPDATE r SET b = a + 1 WHERE a > 2;
UPDATE 1
pg_mooncake (pid: 44455) =# CALL create_snapshot('c');
CALL
```

## Related Issues

Closes https://github.com/Mooncake-Labs/moonlink/issues/242

## Checklist

- [x] Code builds correctly

### DIFF
--- a/src/moonlink/src/storage/mooncake_table/snapshot.rs
+++ b/src/moonlink/src/storage/mooncake_table/snapshot.rs
@@ -51,7 +51,7 @@ pub(crate) struct SnapshotTableState {
     /// ---- Items not persisted to iceberg snapshot ----
     ///
     /// Iceberg snapshot is created in an async style, which means it doesn't correspond 1-1 to mooncake snapshot, so we need to ensure idempotency for iceberg snapshot payload.
-    /// The following fields record unpersisted content, which will be place in iceberg payload everytime.
+    /// The following fields record unpersisted content, which will be placed in iceberg payload everytime.
     unpersisted_iceberg_records: UnpersistedIcebergSnapshotRecords,
 }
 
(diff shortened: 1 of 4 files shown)


src/moonlink/src/storage/mooncake_table/snapshot.rs
```diff
@@ -213,24 +267,19 @@ impl SnapshotTableState {
             self.last_commit = cp;
         }
 
+        // Batch new data files, whether we decide to create an iceberg snapshot.
```
dentiny, line 270: It's a potential problem: as long as iceberg snapshot and mooncake snapshot doesn't correspond to each other 1-1, we need to make sure unpersisted changes (from snapshot task) is stored somewhere.

Issue #26, opened 2025-04-24
Type: Support Decimal


zhousun:

zhousun:
Decimal works but scale/ precision was not set correctly.

> pg_mooncake (pid: 42631) =# create table r(a int primary key, b decimal(10,5), c text[]);
CREATE TABLE
pg_mooncake (pid: 42631) =# call create_mooncake_table('c','r');
CALL
pg_mooncake (pid: 42631) =# insert into r values(2,23476, array['a','fafad']);
INSERT 0 1
pg_mooncake (pid: 42631) =# select * from c;
 a |    b    |     c     
---+---------+-----------
 2 | 0.23476 | {a,fafad}
(1 row)

PR #267, opened 2025-05-28
Support precision and scale

nbiscaro:
## Summary

Decode precision and scale from `atttypmod` and use in the arrow schema. Apply the scale when converting to arrow. 

Decoding with reference to
https://github.com/postgres/postgres/blob/4fbb46f61271f4b7f46ecad3de608fc2f4d7d80f/src/backend/utils/adt/numeric.c#L929v

and 

https://github.com/postgres/postgres/blob/4fbb46f61271f4b7f46ecad3de608fc2f4d7d80f/src/backend/utils/adt/numeric.c#L944

for precision and scale, respectively. 

## Related Issues

Closes https://github.com/Mooncake-Labs/moonlink/issues/26

## Changes

- Added `numeric_precision_scale` to decode precision and scale from `atttypmod` and applied it when mapping numeric columns to Arrow schemas
- Updated schema conversion to propagate type modifiers
- Update testing
- Add arrow schema to`Cargo.toml` and update `Cargo.lock`

Also tested with:

```
DROP EXTENSION IF EXISTS pg_mooncake CASCADE;
DROP TABLE IF EXISTS r;
CREATE EXTENSION pg_mooncake;
CREATE TABLE r (a INT PRIMARY KEY, b DECIMAL(10, 5), c TEXT[]);
INSERT INTO r VALUES
(1, 12345, ARRAY['x', 'y']),
(2, 12.34567, ARRAY['a', 'b']),
(3, 0.5, ARRAY['only one']),
(4, 99.999999, ARRAY['round']),
(5, 0.00001, ARRAY['tiny']),
(6, 99999.99999, ARRAY['max']),
(7, 0.00000, ARRAY['zero']),
(8, 1.11111, ARRAY['a', 'b', 'c', 'd', 'e', 'f']);
SELECT * FROM c ORDER BY a;
pg_mooncake (pid: 34657) =# SELECT * FROM c; 
 a |      b      |       c       
---+-------------+---------------
 1 | 12345.00000 | {x,y}
 2 |    12.34567 | {a,b}
 3 |     0.50000 | {"only one"}
 4 |   100.00000 | {round}
 5 |     0.00001 | {tiny}
 6 | 99999.99999 | {max}
 7 |     0.00000 | {zero}
 8 |     1.11111 | {a,b,c,d,e,f}
(8 rows)
```

## Checklist

- [x] Code builds correctly

### DIFF
--- a/src/moonlink_connectors/src/pg_replicate/util.rs
+++ b/src/moonlink_connectors/src/pg_replicate/util.rs
@@ -342,7 +342,18 @@ impl From<PostgresTableRow> for MoonlinkRow {
                 Cell::Numeric(value) => {
                     match value {
                         PgNumeric::Value(bigdecimal) => {
-                            values.push(RowValue::Decimal(bigdecimal.to_i128().unwrap()));
+                            let (int_val, scale) = bigdecimal.as_bigint_and_exponent();
+                            let multiplier = 10_i128.pow(scale as u32);
+                            if let Some(scaled_integer) =
+                                int_val.to_i128().and_then(|v| v.checked_mul(multiplier))
+                            {
+                                values.push(RowValue::Decimal(scaled_integer));
+                            } else {
+                                values.push(RowValue::Null); // handle overflow safely
+                                eprintln!(
+                                    "Decimal value too large to fit in i128 — storing as NULL"
+                                );
+                            }
                         }
                         _ => {
                             // DevNote:


src/moonlink_connectors/src/pg_replicate/util.rs
```diff
@@ -24,7 +40,11 @@ fn postgres_primitive_to_arrow_type(
         Type::INT8 => (DataType::Int64, None),
         Type::FLOAT4 => (DataType::Float32, None),
         Type::FLOAT8 => (DataType::Float64, None),
-        Type::NUMERIC => (DataType::Decimal128(38, 10), None),
+        Type::NUMERIC => {
+            let (precision, scale) = numeric_precision_scale(modifier)
+                .unwrap_or((DECIMAL128_MAX_PRECISION, DECIMAL_DEFAULT_SCALE));
```
nbiscaro, line 45: We might need to lower `DECIMAL_128_MAX_PRECISION` to leq 18 for parquet efficiency reasons.

Parquet encodes DECIMAL values as either:
	•	INT32 (if precision ≤ 9)
	•	INT64 (if precision ≤ 18)
	•	FIXED_LEN_BYTE_ARRAY (if precision > 18)
	

https://github.com/apache/parquet-format/blob/master/LogicalTypes.md#decimal

We effectively double our storage size and lose `INT64/32` efficiency on query.

src/moonlink_connectors/Cargo.toml
```diff
@@ -5,6 +5,7 @@ edition = { workspace = true }
 
 [dependencies]
 arrow = { workspace = true }
+arrow-schema.workspace = true
```
dentiny, line 8: It fails cargo-sort on my end.
```sh
cargo-sort...............................................................Failed
- hook id: cargo-sort
- exit code: 101

Checking pg_moonlink...
Finished: Cargo.toml for "pg_moonlink" has been rewritten
Checking moonlink...
Finished: Cargo.toml for "moonlink" has been rewritten
Checking moonlink_backend...
Finished: Cargo.toml for "moonlink_backend" has been rewritten
Checking moonlink_connectors...

thread 'main' panicked at /usr/local/cargo/registry/src/index.crates.io-1949cf8c6b5b557f/cargo-sort-1.0.9/src/sort.rs:113:46:
called `Result::unwrap()` on an `Err` value: TomlError { message: "TOML parse error at line 8, column 13\n  |\n8 | arrow-schema.workspace = true\n  |             ^\nUnexpected `.`\nExpected `=`\n" }
```

PR #274, opened 2025-05-28
re-arch streaming delete & fix update+snapshot

zhousun:
## Summary
Re-design how deletion works for streaming xact.
1. Delete 'streaming appending row', record directly in XactState
2. Delete flushed table row, record in snapshot.uncommited_deletion
3. Delete memslice table row, record as special 'pending_deletes' in xactState, and process at commit.

Also, fix a race condition if we delete + add new file in same snapshot
When trying to apply delete, skip files that's added after the LSN.

## Checklist

- [x] Code builds correctly

### DIFF
--- a/src/moonlink/src/storage/mooncake_table/transaction_stream.rs
+++ b/src/moonlink/src/storage/mooncake_table/transaction_stream.rs
@@ -4,9 +4,10 @@ use crate::storage::storage_utils::ProcessedDeletionRecord;
 /// Used to track the state of a streamed transaction
 /// Holds appending rows in memslice and files.
 /// Deletes are more complex,
-/// 1. Deletes on appending rows is tracked in `local_deletions`
-/// 2. Deletes on flushed files are directly pushed to snapshot_task.new_deletions
-/// 3. Deletes on "in-memory" rows are tracked in `pending_deletions_in_main_mem_slice`
+/// 1. row belong to stream state memslice, directly delete it.
+/// 2. row belong to stream state flushed file, add to `local_deletions`
+/// 3. row belong to main table's flushed files, directly pushed to snapshot_task.new_deletions and let snapshot handle it.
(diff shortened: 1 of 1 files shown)


src/moonlink/src/storage/mooncake_table/transaction_stream.rs
```diff
@@ -0,0 +1,311 @@
+use super::*;
+use crate::storage::index::Index;
+use crate::storage::storage_utils::ProcessedDeletionRecord;
+/// Used to track the state of a streamed transaction
+/// Holds appending rows in memslice and files.
+/// Deletes are more complex,
+/// 1. Deletes on appending rows is tracked in `local_deletions`
+/// 2. Deletes on flushed files are directly pushed to snapshot_task.new_deletions
+/// 3. Deletes on "in-memory" rows are tracked in `pending_deletions_in_main_mem_slice`
```
nbiscaro: I think there are 4 cases in the streaming delete case: 
1) Delete a row that was appended in the stream mem slice
2) Delete a row that is in the main mem slice
3) Delete a row that is in the stream disk file
4) Delete a row that is in the main disk file

src/moonlink/src/storage/mooncake_table/transaction_stream.rs
```diff
@@ -0,0 +1,296 @@
...
+        }
+    }
+}
+
+// DevNote:
+// This is a trick to track xact of uncommitted deletions
+// we set first 32 bits to 1, so it will be 'uncommitted' as the value is larger than any possible lsn.
+// And we use the last 32 bits to store the xact_id, so we can find deletion for a given xact_id.
+fn get_lsn_for_pending_xact(xact_id: u32) -> u64 {
+    0xFFFF_FFFF_0000_0000 | xact_id as u64
```
nbiscaro, line 51: We should use `u64::Max` https://doc.rust-lang.org/std/primitive.u64.html#associatedconstant.MAX

PR #278, opened 2025-05-29
Improve devcontainer setup

dentiny:
## Summary

A few things:
- Update directory from `pg_moonlink` to repository name to `moonlink` 
- Remove apt cache to reduce devcontainer size
- Move clippy installation to dockerfile 

I checked a few things:
- `cargo build` and `cargo test` works with no problem
- precommit hook works (clippy and cargo sort)

## Checklist

- [x] Code builds correctly

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -2,13 +2,13 @@ FROM mcr.microsoft.com/devcontainers/rust:latest
 
 RUN apt update && \
     apt install -y python3-pip && \
-    pip3 install pre-commit --break-system-packages &&
+    pip3 install pre-commit --break-system-packages && \
     rm -rf /var/lib/apt/lists/*
 
 # Now switch to vscode user for cargo installations, otherwise later cargo commands required root permission.
 USER vscode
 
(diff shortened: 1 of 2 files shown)


.devcontainer/Dockerfile
```diff
@@ -1,15 +1,14 @@
...
+    apt install -y python3-pip && \
+    pip3 install pre-commit --break-system-packages && \
+    rm -rf /var/lib/apt/lists/*
 
 # Now switch to vscode user for cargo installations, otherwise later cargo commands required root permission.
 USER vscode
 
-RUN cargo install cargo-sort
+RUN cargo install cargo-sort && \
+    rustup component add clippy
```
dpxcc, line 12: `clippy` is already part of `mcr.microsoft.com/devcontainers/rust:latest`
rm

dentiny: sure, I do find it in the content list: https://github.com/microsoft/vscode-dev-containers/blob/main/containers/rust/history/dev.md#contents
I put it here since last time I installed it manually, but seems to work now.

Issue #235, opened 2025-05-23
[ISSUE] Implement drop table


nbiscaro:
## Summary

A basic drop table implementation would be useful for testing as well. Don't worry about file clean up for now just drop the state. 

## Context

Why is this issue important? Link any relevant discussions, designs, or documents.

## Tasks

- [ ] 

## Notes

Any extra notes or details that might be useful.

PR #283, opened 2025-05-29
[backend] [2/N] Drop table implementation

dentiny:
## Summary

This PR is the second part for dropping postgres table, which performs the reserve operation as table creation in reserver order.

## Related Issues

Closes https://github.com/Mooncake-Labs/moonlink/issues/235

## Checklist

- [x] Code builds correctly

### DIFF
--- a/src/moonlink_backend/src/lib.rs
+++ b/src/moonlink_backend/src/lib.rs
@@ -87,7 +87,7 @@ impl<T: Eq + Hash + Clone> MoonlinkBackend<T> {
 mod tests {
     use super::*;
     use tempfile::TempDir;
-    use tokio_postgres::{connect, NoTls, Client};
+    use tokio_postgres::{connect, Client, NoTls};
 
     #[test]
     fn test_recreate_directory() {


src/moonlink_connectors/src/replication_connection.rs
```diff
@@ -181,6 +230,14 @@ impl ReplicationConnection {
         Ok(table_schema.table_id)
     }
 
+    /// Remove the given table from connection.
+    pub async fn drop_table(&mut self, table_id: u32) -> Result<()> {
+        let table_name = self.source.remove_table_schema(table_id);
+        self.remove_table_from_publication(&table_name).await?;
+        self.remove_table_from_replication(table_id).await?;
+        Ok(())
```
nbiscaro: I think you should be removing the table from the publication as the first step. We should ensure we are not still receiving CDC events while we are dropping the state and schemas t avoid any inconsistent state. 

For example in the current implementation, we are dropping the schema before removing from the publication. It's possible in this case that we receive a CDC event but have already dropped the schema (which we need to parse the event)

src/moonlink_connectors/src/replication_manager.rs
```diff
@@ -63,6 +63,14 @@ impl<T: Eq + Hash> ReplicationManager<T> {
         Ok(())
     }
 
+    /// Drop table specified by the given table id.
+    pub async fn drop_table(&mut self, external_table_id: T) -> Result<()> {
+        let (uri, table_id) = self.table_info.remove(&external_table_id).unwrap();
+        let repl_conn = self.connections.get_mut(&uri).unwrap();
+        repl_conn.drop_table(table_id).await?;
```
nbiscaro: I think it makes sense to swap the order here so that we are
1) drop table from replication connection
2) drop table info

src/moonlink_connectors/src/replication_connection.rs
```diff
@@ -98,6 +98,27 @@ impl ReplicationConnection {
...
+            .unwrap();
+
+        // Optionally reset REPLICA IDENTITY to default
+        self.postgres_client
+            .simple_query(&format!(
+                "ALTER TABLE {} REPLICA IDENTITY DEFAULT;",
+                table_name
+            ))
+            .await
+            .unwrap();
```
nbiscaro: How do you know that it was originally default? I think we should remove this.

dentiny: done

src/moonlink_connectors/src/replication_connection.rs
dentiny: Updated.

src/moonlink_connectors/src/replication_manager.rs
dentiny: updated

nbiscaro:
I think logically this looks good but needs to be rebased onto https://github.com/Mooncake-Labs/moonlink/pull/287. 

Instead of removing from shared state send a `DropTable` command to the sink and drop it from within.

dentiny:
> I think logically this looks good but needs to be rebased onto #287.
> 
> Instead of removing from shared state send a `DropTable` command to the sink and drop it from within.

Sure, I will merge after conflict resolved. Thanks for the review!

Issue #205, opened 2025-05-21
Cleanup synchronous IO operation inside `Drop`


dentiny:
## Summary

https://github.com/Mooncake-Labs/moonlink/blob/c0049498fb73d68c47fedeae80cda1561d4e12af/src/moonlink/src/union_read/read_state.rs#L19-L26

PR #284, opened 2025-05-29
[storage] Cleanup associated file deletion in async style

dentiny:
## Summary

Our current way to delete associated files (temporary files for union read) are blocking, which blocks tokio runtime.
This PR turns it into fire-and-forget manner, with detached tokio tasks.

## Related Issues

Closes https://github.com/Mooncake-Labs/moonlink/issues/205

## Checklist

- [x] Code builds correctly

### DIFF
--- a/src/moonlink/src/union_read/read_state.rs
+++ b/src/moonlink/src/union_read/read_state.rs
@@ -21,7 +21,7 @@ impl Drop for ReadState {
         let associated_files = std::mem::take(&mut self.associated_files);
         println!("Dropping files: {:?}", associated_files);
         // Perform best-effort deletion by spawning detached task.
-        let _ = tokio::runtime::Handle::current().spawn(async move {
+        tokio::spawn(async move {
             for file in associated_files.into_iter() {
                 if let Err(e) = tokio::fs::remove_file(&file).await {
                     println!("Failed to delete file {} because {:?}", file, e);


src/moonlink/src/union_read/read_state.rs
```diff
@@ -18,10 +18,16 @@ pub struct ReadState {
...
 impl Drop for ReadState {
     fn drop(&mut self) {
-        println!("Dropping files: {:?}", self.associated_files);
-        for file in self.associated_files.iter() {
-            std::fs::remove_file(file).unwrap();
-        }
+        let associated_files = std::mem::take(&mut self.associated_files);
+        println!("Dropping files: {:?}", associated_files);
+        // Perform best-effort deletion by spawning detached task.
+        tokio::spawn(async move {
```
dentiny, line 24: Caveat is we have no guarantee these temporary files will be cleanup.
But TBH it's never guaranteed before this change also, for example, process crashes the temp files will just leave there forever.

Not an issue for unit tests, since they have their own temporarily directories.

dpxcc: Can we place temp files in a separate dir so that we can clean them up upon server restart?

dentiny: Temporary files are now places under mooncake table directory: https://github.com/Mooncake-Labs/moonlink/blob/ec9c73a3a4607c5b9ba896c3abf810110fe7433c/src/moonlink/src/storage/mooncake_table.rs#L165-L173, which makes it hard to track and cleanup.

dentiny: Followup PR: https://github.com/Mooncake-Labs/moonlink/pull/290

PR #297, opened 2025-05-30
[storage] Implement force snapshot for stream write

dentiny:
## Summary

Finish a TODO item: to support force iceberg snapshot in streaming write.
The only added change is to trigger force snapshot at stream commit if necessary.

Also fix a bug on force iceberg snapshot, see https://github.com/Mooncake-Labs/moonlink/pull/297/files#r2114994829 for details.

## Checklist

- [x] Code builds correctly

### DIFF
--- a/src/moonlink/src/storage/mooncake_table/transaction_stream.rs
+++ b/src/moonlink/src/storage/mooncake_table/transaction_stream.rs
@@ -106,7 +106,6 @@ impl MooncakeTable {
             pos: None,
             row_identity: self.metadata.identity.extract_identity_columns(row),
         };
-
         let stream_state = Self::get_or_create_stream_state(
             &mut self.transaction_stream_states,
             &self.metadata,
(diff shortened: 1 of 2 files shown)


src/moonlink/src/storage/mooncake_table/snapshot.rs
```diff
@@ -279,8 +279,12 @@ impl SnapshotTableState {
         //
         // TODO(hjiang): Error handling for snapshot sync-up.
         let mut iceberg_snapshot_payload: Option<IcebergSnapshotPayload> = None;
-        let flush_by_data_files =
-            self.create_iceberg_snapshot_by_data_files(new_data_files.as_slice(), force_create);
+        let flush_by_data_files = self.create_iceberg_snapshot_by_data_files(
```
dentiny, line 282: This is a bug, chances are that:
- A mooncake snapshot is ongoing;
- Force snapshot is requested;
- Now threshold to create iceberg snapshot is still standard / high;
- Next mooncake snapshot we actually get "force snapshot marker", but new data files are empty;
- No iceberg snapshot will be created.

PR #308, opened 2025-05-30
Fix pre commit

nbiscaro:
## Summary

Our pre-commit currently only runs for default features. We can add the `--all-features` flag to ensure we lint all features.
Additionally, we only lint files that were touched (which is good) but this means that any linting errors introduced before the pre-commit hook was added are still around.

## Related Issues

## Changes

- add `--all-features` flag
- fix lints behind `feature = "storage-s3"`
- fix old lints
- indent `yml`

## Checklist

- [ ] Code builds correctly

### DIFF
--- a/src/moonlink/benches/bench_write.rs
+++ b/src/moonlink/benches/bench_write.rs
@@ -141,22 +141,24 @@ fn bench_write(c: &mut Criterion) {
                         );
                     }
                     let handle = table.flush_transaction_stream(1);
-                    let _ = handle.await.unwrap();
+                    handle.await.unwrap();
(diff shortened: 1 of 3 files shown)


.pre-commit-config.yaml
```diff
@@ -2,25 +2,25 @@ repos:
   - repo: https://github.com/doublify/pre-commit-rust
     rev: v1.0
     hooks:
-    - id: fmt
+      - id: fmt
```
dentiny, line 5: Just for my own knowledge, is yml indentation necessary / must / recommended?

PR #316, opened 2025-05-31
Add basic CI run

nbiscaro:
## Summary

Adds a basic CI that runs on PR. This will be useful for maintaining integrity of the codebase and providing peace of mind for us and our open source contributors going forward. It will be good to play with this internally for the next week before open sourcing.

The pipeline is pretty basic, and runs two main jobs:
1) The first job runs our format and linting with `rustfmt` and `clippy`. In order to ensure the formatting and linter expectations are the same between our upstream and local `pre-commit` hooks, we use a `rust-toolchain.toml` to pin the versions. 

2) The second job runs our unit tests and also generates a code coverage report with `llvm-cov`. _We don't run any integration tests_. Currently, `llvm-cov` simply generates an HTML report and stores it in artifacts. We post a comment on the PR with a link to the downloadable artifact (the comment will update in-place on new pushes so no spam). This is purely temporary and we should upload them to something like https://about.codecov.io/codecov-free-trial/ (note they are free for open source!). We can eventually add a parameter to `llvm-cov` to fail the CI if test coverage dips below a certain percentage, but I will keep this off for now. 

To reduce build times during run we use `swatinem/rust-cache@v2` which will cache build artifacts across jobs and runs (does not cache on failure). 

(Heres an example of what the comment looks like with artifact link: https://github.com/nbiscaro/pg_moonlink/pull/2#issuecomment-2923718125)

## Related Issues

## Changes

- 
- 
- 

## Checklist

- [ ] Code builds correctly

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -73,7 +73,7 @@ jobs:
       - name: Upload to Codecov
         uses: codecov/codecov-action@v5
         env:
-          CODECOV_TOKEN: ${{ secrets.CODECOV_ORG_TOKEN }}
+          CODECOV_TOKEN: ${{ secrets.CODECOV_TOKEN }}
         with:
           files: lcov.info
           verbose: true


.github/workflows/ci.yml
```diff
@@ -0,0 +1,80 @@
...
+      - uses: dtolnay/rust-toolchain@stable
+        id: toolchain
+        with:
+          components: rustfmt,clippy
+
+      - uses: swatinem/rust-cache@v2
+        with:
+          shared-key: ${{ steps.toolchain.outputs.cachekey }}
+
+      - name: rustfmt check
```
dentiny, line 42: You could use `pre-commit` tool to run and check all pre-commit hooks.
The benefit is, if we add more precommit hooks in the future, we don't need to update both places.

dentiny:
Hi I didn't see CI page for my PR (https://github.com/Mooncake-Labs/moonlink/pull/323), is that expected? Thanks!

Issue #327, opened 2025-05-31
iceberg: clean up puffin blob write logic


dentiny:
## Summary

Current implementation is at an intermediate state, which rewrite manifest list for multiple times; I left TODO last time but forgot to revisit.
https://github.com/Mooncake-Labs/moonlink/blob/b0b49b711d59024814ae20d8760267d6bade8cda/src/moonlink/src/storage/iceberg/file_catalog.rs#L725-L733

PR #328, opened 2025-05-31
[storage] Cleanup puffin blob write logic

dentiny:
## Summary

Current implementation is bad and in-performant:
- It would rewrite manifest list for multiple times
- It iterates all manifest entries for multiple times

I left a TODO then but forgot to revisit :(

## Related Issues

Closes https://github.com/Mooncake-Labs/moonlink/issues/327

## Checklist

- [x] Code builds correctly

### DIFF
--- a/src/moonlink/src/storage/iceberg/puffin_writer_proxy.rs
+++ b/src/moonlink/src/storage/iceberg/puffin_writer_proxy.rs
@@ -354,7 +354,7 @@ pub(crate) async fn append_puffin_metadata_and_rewrite(
 
     // Delete existing manifest list file and rewrite.
     let mut manifest_list_writer =
-        create_new_manifest_list_writer(table_metadata, &cur_snapshot, file_io).await?;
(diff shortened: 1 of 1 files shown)


src/moonlink/src/storage/iceberg/file_catalog.rs
```diff
@@ -719,18 +719,8 @@ impl Catalog for FileCatalog {
...
-        for (puffin_filepath, puffin_blob_metadata) in self.puffin_blobs.iter() {
-            append_puffin_metadata_and_rewrite(
-                &metadata,
-                &self.file_io,
-                puffin_filepath,
-                puffin_blob_metadata.clone(),
-            )
-            .await?;
-        }
+        // TODO(hjiang): Add unit test for update and check manifest population.
```
dentiny, line 722: Involves types are from iceberg-rust internal and it's not easy to construct ourselves :(

Issue #335, opened 2025-06-02
iceberg: no block wait in critical section


dentiny:
## Summary

https://github.com/Mooncake-Labs/moonlink/blob/ead85eb639bf8d01af9c2e4a2e76e3cf287022ca/src/moonlink_backend/src/lib.rs#L75-L83

Issue #332, opened 2025-06-01
iceberg: support multiple iceberg snapshot creation requests


dentiny:
## Summary

Currently I haven't implemented multiple force snapshot support, simply check failure.
https://github.com/Mooncake-Labs/moonlink/blob/ead85eb639bf8d01af9c2e4a2e76e3cf287022ca/src/moonlink/src/table_handler.rs#L200-L201
Before make it public, at least I should pass error status into channel.

There's no technical blocker for the feature, just not yet implemented.

PR #337, opened 2025-06-02
[storage] Support multiple snapshot creation

dentiny:
## Summary

This PR does two main changes:
- Support multiple iceberg snapshot create requests
- Move iceberg snapshot creation out of critical section

Basically, we use an ordered map to store LSN and its receive channel;
when iceberg snapshot completes, notify all channels if their LSN satisfied.

## Related Issues

Closes https://github.com/Mooncake-Labs/moonlink/issues/335
Closed https://github.com/Mooncake-Labs/moonlink/issues/332

## Checklist

- [x] Code builds correctly

### DIFF
--- a/src/moonlink/src/table_handler/tests.rs
+++ b/src/moonlink/src/table_handler/tests.rs
@@ -845,8 +845,8 @@ async fn test_empty_table_snapshot_creation() {
             .await;
         rx_vec.push(rx);
     }
-    for idx in 0..=9 {
-        rx_vec[idx].recv().await.unwrap().unwrap();
+    for mut cur_rx in rx_vec {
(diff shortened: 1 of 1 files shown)


src/moonlink/src/error.rs
```diff
@@ -32,6 +32,9 @@ pub enum Error {
 
     #[error("Iceberg error: {source}")]
     IcebergError { source: IcebergError },
+
+    #[error("Iceberg error: {0}")]
+    IcebergMessage(String),
```
dentiny, line 37: I need to create an error, either it's constructed from a copiable type (error is not), or constructed from string.

Issue #343, opened 2025-06-03
Bug: batch deletion vector delete row failure


dentiny:
Repro SQL:
```sql
pg_mooncake (pid: 4069) =# DROP EXTENSION IF EXISTS pg_mooncake CASCADE;
DROP TABLE IF EXISTS r;
CREATE EXTENSION pg_mooncake;
CREATE TABLE r (a int primary key, b text);
CALL create_mooncake_table('c', 'r');
INSERT INTO r SELECT a, 'val_' || a FROM generate_series(1, 1000000) AS a;
SELECT count(*) FROM c;
DELETE FROM r;
SELECT count(*) FROM c;
```

Error message:
```sh
thread 'tokio-runtime-worker' panicked at moonlink/src/moonlink/src/storage/mooncake_table/snapshot.rs:138:17:
assertion failed: deletion_vector.delete_row(*row_idx)
stack backtrace:
   0: rust_begin_unwind
             at /rustc/05f9846f893b09a1be1fc8560e33fc3c815cfecb/library/std/src/panicking.rs:695:5
   1: core::panicking::panic_fmt
             at /rustc/05f9846f893b09a1be1fc8560e33fc3c815cfecb/library/core/src/panicking.rs:75:14
   2: core::panicking::panic
             at /rustc/05f9846f893b09a1be1fc8560e33fc3c815cfecb/library/core/src/panicking.rs:145:5
   3: moonlink::storage::mooncake_table::snapshot::SnapshotTableState::aggregate_committed_deletion_logs
             at /workspaces/pg_mooncake/moonlink/src/moonlink/src/storage/mooncake_table/snapshot.rs:138:17
   4: moonlink::storage::mooncake_table::snapshot::SnapshotTableState::update_snapshot::{{closure}}
             at /workspaces/pg_mooncake/moonlink/src/moonlink/src/storage/mooncake_table/snapshot.rs:314:17
   5: moonlink::storage::mooncake_table::MooncakeTable::create_snapshot_async::{{closure}}
             at /workspaces/pg_mooncake/moonlink/src/moonlink/src/storage/mooncake_table.rs:644:14
```

dentiny:
The reason is, when we flush record batches to disk, we cut off by **accumulated size** but not **row count**.

The issue is, currently we initialize batch deletion vector (which pairs with in-memory batch record, or on-disk data file) with batch size.

[...]

PR #345, opened 2025-06-03
[storage] Fix deletion vector max number

dentiny:
## Summary

Batch deletion vector's max row number is not correctly assigned: it should be either unknown, or known parquet row number.
Assertion failure will be triggered when row number in parquet is larger than batch size.
See attached issue for details.

## Related Issues

Closes https://github.com/Mooncake-Labs/moonlink/issues/343

## Checklist

- [x] Code builds correctly

### DIFF
--- a/src/moonlink/src/storage/iceberg/iceberg_table_manager.rs
+++ b/src/moonlink/src/storage/iceberg/iceberg_table_manager.rs
@@ -372,7 +372,7 @@ impl IcebergTableManager {
     async fn sync_data_files(
         &mut self,
         new_data_files: Vec<MooncakeDataFileRef>,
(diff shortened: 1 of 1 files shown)


dentiny:
Checked with the following SQL for 50 times, no issue found:
```sql
pg_mooncake (pid: 25917) =# DROP EXTENSION IF EXISTS pg_mooncake CASCADE;
DROP TABLE IF EXISTS r;
CREATE EXTENSION pg_mooncake;
INSERT INTO r SELECT a, 'val_' || a FROM generate_series(1, 400000) AS a;
SELECT count(*) FROM c;
DELETE FROM r;
SELECT count(*) FROM c;
```

Issue #331, opened 2025-06-01
Refactor join handles in table handler


dentiny:
## Summary

All IO related operations inside of table handlers are initiated and synchronized via join handle.
https://github.com/Mooncake-Labs/moonlink/blob/ead85eb639bf8d01af9c2e4a2e76e3cf287022ca/src/moonlink/src/table_handler.rs#L91-L97

But in the future we will have more events, for example, flush event, merge index, compaction, etc.
Having multiple handles make code messy.

PR #367, opened 2025-06-04
[storage] Refactor table handler to use channel instead of join handle

dentiny:
## Summary

This PR refactors table handlers to use channel for completion notification, instead of join handle.
The main motivation is to 
(1) improve code quality to prepare for future table internal events (i.e. index merge, compaction, etc)
(2) less error prune to use channels
(3) perf improvement due to reduced branch number (less future created at poll).

## Related Issues

Closes https://github.com/Mooncake-Labs/moonlink/issues/331

## Checklist

- [x] Code builds correctly


src/moonlink/src/table_handler.rs
```diff
@@ -81,60 +93,56 @@ impl TableHandler {
...
-        let mut iceberg_snapshot_handle: IcebergSnapshotHandle = None;
-
         // Requested minimum LSN for a force snapshot request.
         let mut force_snapshot_lsns: BTreeMap<u64, Vec<Sender<Result<()>>>> = BTreeMap::new();
 
         // Record LSN if the last handled table event is committed, which indicates mooncake table stays at a consistent view, so table could be flushed safely.
         let mut table_consistent_view_lsn: Option<u64> = None;
 
+        // Whether there's an ongoing mooncake snapshot operation.
+        let mut mooncake_snapshot_ongoing = false;
```
dentiny, line 108: This is the trickiest part of this PR, we used to determine whether ongoing snapshot by checking handle none/some, need to leverage an extra boolean value.

PR #369, opened 2025-06-04
[storage] Fix hash index merge

dentiny:
## Summary

This PR fixes hash index merge, which should consider file id remapping when merge. 

## Checklist

- [x] Code builds correctly

### DIFF
--- a/src/moonlink/src/storage/index/persisted_bucket_hash_map.rs
+++ b/src/moonlink/src/storage/index/persisted_bucket_hash_map.rs
@@ -481,8 +481,13 @@ impl<'a> IndexBlockIterator<'a> {
                 .read_entry(&mut self.entry_reader, self.metadata)
                 .await;
             self.current_entry += 1;
-            if *self.file_id_remap.get(seg_idx).unwrap() != _INVALID_FILE_ID {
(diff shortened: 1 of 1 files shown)


src/moonlink/src/storage/index/persisted_bucket_hash_map.rs
```diff
@@ -481,8 +481,13 @@ impl<'a> IndexBlockIterator<'a> {
                 .read_entry(&mut self.entry_reader, self.metadata)
                 .await;
             self.current_entry += 1;
-            if *self.file_id_remap.get(seg_idx).unwrap() != _INVALID_FILE_ID {
-                return Some((lower_hash + self.current_upper_hash, seg_idx, row_idx));
+            let seg_idx = self.file_id_remap.get(seg_idx).unwrap();
+            if *seg_idx != _INVALID_FILE_ID {
```
dentiny, line 485: hmm looks we should assertion failure, rather than skip the non-existent remapped file id

zhousun: Oh this is intentional. After you implement compaction, files could be deleted.

PR #389, opened 2025-06-05
Add dependabot and its check

dentiny:
## Summary

This PR adds dependabot into our repo and validation into CI, so we don't need to upgrade ourselves.

I have tested on my forked repo and confirmed it works.
image

If it doesn't work as expected I will get it reverted.
It doesn't affect moonlink functionality.

## Checklist

- [ ] Code builds correctly


dentiny:
I experimented in my own repo (after merge the dependabot change) and does receive upgrade PR, I will investigate and fix / revert if it doesn't work.

Issue #381, opened 2025-06-05
Integration data file compaction and iceberg snapshot into mooncake table


dentiny:
## Summary

- Periodically check the necessity of data file compaction
- Trigger async data file compaction in table handler
- After completion, buffer the the change to next snapshot task and later persist to iceberg

One thing worth noticing, data file compaction cannot happen at the same time with file index merge.

dentiny:
Possible states: https://docs.google.com/document/d/1aiQqhl5F8QODJm3HPl47BZX0rfNyUbUPSHGArUcCIw4/edit?usp=sharing
Plan to do TDD

Update: initial compaction only happens on those persisted to iceberg, so no need to deal with complicated remap.

PR #412, opened 2025-06-09
[storage] Integrate data compaction with snapshot and eventloop

dentiny:
## Summary

This PR integrate data compaction into eventloop and mooncake/iceberg snapshot.
To prevent unintended production performance and reliability issues, compaction is **by defaulted turned off** until we figure out integration test with TPC-C.

To reduce code change (since it's already a huge PR), I left a few TODO item for performance issues.
It's also worth noting, current implementation only compacts data files with deletion vectors, but not small files, will followup with other PRs.

## Related Issues

Closes https://github.com/Mooncake-Labs/moonlink/issues/381

## Checklist

- [x] Code builds correctly


src/moonlink/src/storage/mooncake_table/snapshot.rs
```diff
@@ -346,29 +452,184 @@ impl SnapshotTableState {
...
+            // If there's deletion record, try remap to the new compacted data file.
+            let deleted_rows = old_entry.batch_deletion_vector.collect_deleted_rows();
+            for cur_deleted_row in deleted_rows {
+                let old_record_location =
+                    RecordLocation::DiskFile(cur_old_data_file.file_id(), cur_deleted_row as usize);
+                let new_record_location =
+                    remapped_data_files_after_compaction.get(&old_record_location);
+                // Case-1: The old record still exists, need to remap.
+                if let Some(new_record_location) = new_record_location {
+                    // TODO(hjiang): A quick hack, there's only one compacted data file.
```
dentiny, line 580: I will address in the followup PR (considering the PR size).

src/moonlink/src/storage/mooncake_table/snapshot.rs
```diff
@@ -272,17 +294,101 @@ impl SnapshotTableState {
...
+        if tentative_data_files_to_compact.len()
+            < self
+                .mooncake_table_config
+                .data_compaction_config
+                .data_file_to_compact as usize
+        {
+            return None;
+        }
+
+        // TODO(hjiang): Current implementation take the easy path to code, which does a sequential scan to get all file indices.
```
dentiny, line 370: I will followup in another PR, it's guarded by config and won't be triggered.

PR #444, opened 2025-06-12
[storage] Unify local and remote path

dentiny:
## Summary

This PR is a big refactor, to use remote file path (aka, iceberg data file path) in mooncake snapshot, so later when integrated with data file, write through cache could be evicted after successful persistence.

## Checklist

- [x] Code builds correctly


src/moonlink/src/storage/iceberg/tests.rs
```diff
@@ -1390,19 +1392,6 @@ async fn mooncake_table_snapshot_persist_impl(warehouse_uri: String) -> IcebergR
     check_row_index_on_disk(&snapshot, &row4).await;
     assert_eq!(snapshot.data_file_flush_lsn.unwrap(), 500);
 
-    // The old data file and deletion vector is unchanged.
-    let old_data_entry = iceberg_table_manager
```
dentiny, line 1394: We used to use local file as key, so before and after recovery it's the same entry to remove; but hard to do so with file id as key.

Issue #457, opened 2025-06-13
Enable compression for parquet files


dentiny:
## Summary

As titled, no technical blocker.

PR #464, opened 2025-06-15
[storage] ZSTD compress parquet data file

dentiny:
## Summary

This PR enables ZSTD level 5 to compress data files inline.
The consideration for compression level is 
(1) overhead for small files, not a big concern here, since we only have small files for last streaming writes
(2) tradeoff between IO/storage and CPU overhead
I used ZSTD for a few years, and level 5 is the default I used.

I tested by writing a few parquet files, and confirm file size does change.
Also checked with pg_mooncake, and confirm zstd compression is already handled (by duckdb parquet reader).
```sql
pg_mooncake (pid: 5837) =# DROP EXTENSION IF EXISTS pg_mooncake CASCADE;
DROP TABLE IF EXISTS r;
CREATE EXTENSION pg_mooncake;
SELECT count(*) FROM c;
SELECT count(*) FROM c;
DROP EXTENSION
DROP TABLE
CREATE EXTENSION
CREATE TABLE
CALL
INSERT 0 250000
 count  
--------
 250000
(1 row)

DELETE 125000
 count  
--------
 125000
(1 row)
```

## Related Issues

Closes https://github.com/Mooncake-Labs/moonlink/issues/457

## Checklist

- [x] Code builds correctly

### DIFF
--- a/src/moonlink/src/storage/compaction/compactor.rs
+++ b/src/moonlink/src/storage/compaction/compactor.rs
@@ -9,6 +9,8 @@ use futures::TryStreamExt;
 use more_asserts as ma;
 use parquet::arrow::async_reader::ParquetRecordBatchStreamBuilder;
 use parquet::arrow::AsyncArrowWriter;
+use parquet::basic::{Encoding, ZstdLevel};
+use parquet::file::properties::WriterProperties;
 
 use crate::storage::compaction::table_compaction::{
     CompactedDataEntry, DataCompactionPayload, DataCompactionResult, RemappedRecordLocation,
@@ -26,6 +28,9 @@ use crate::storage::storage_utils::{
 use crate::{create_data_file, Result};
 
 type DataFileRemap = HashMap<RecordLocation, RemappedRecordLocation>;
+
+/// Default ZSTD compression level for parquet data.
+const PARQURT_ZSTD_COMPRESSION_LEVEL: i32 = 5;
 
 pub(crate) struct CompactionFileParams {
     /// Local directory to place compacted data files.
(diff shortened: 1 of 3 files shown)


src/moonlink/src/storage/compaction/compactor.rs
```diff
@@ -27,6 +29,9 @@ use crate::{create_data_file, Result};
 
 type DataFileRemap = HashMap<RecordLocation, RemappedRecordLocation>;
 
+/// Default ZSTD compression level for parquet data.
+const PARQURT_ZSTD_COMPRESSION_LEVEL: i32 = 5;
```
dentiny, line 33: No need to take the same compression config for disk slice and compaction, since one happens in background the other in foreground, thus different latency requirement.

dpxcc:
By default, DuckDB uses SNAPPY instead of ZSTD: https://github.com/duckdb/duckdb/blob/main/extension/parquet/parquet_extension.cpp#L208
(Note: their default ZSTD compression level is 3, but that’s not directly relevant here)

pg_mooncake v0.1 also follows the same default: https://github.com/Mooncake-Labs/pg_mooncake/blob/v0.1/src/columnstore/columnstore_table.cpp#L66

Unless we've done thorough performance testing, I’d be hesitant to change these defaults, as it could introduce subtle and hard-to-diagnose performance regressions

zhang-wenchao:
The compression ratio of zstd is very high. I think one of the keys to the success of clickhouse is the super high compression ratio. https://google.github.io/snappy. Snappy is a compression/decompression library. It does not aim for maximum compression.

One of the dissatisfactions we have with duckdb is that the compression ratio is not high.

Is it possible to make it configurable?

dpxcc:
> One of the dissatisfactions we have with duckdb is that the compression ratio is not high.

I believe you are comparing DuckDB and Clickhouse's own formats, but not Parquet
The compression ratio of DuckDB's own format also improved significantly over the year: https://duckdb.org/2022/10/28/lightweight-compression.html

For Parquet, Clickhouse also defaults to LZ4 instead of ZSTD
Even when user manually chooses ZSTD, it defaults to compression level 3
https://github.com/ClickHouse/ClickHouse/blob/master/src/Processors/Formats/Impl/Parquet/Write.h#L27

Sure, we can and should definitely make Parquet options configurable in the future
Both DuckDB and Clickhouse support that
But we shouldn't change defaults without thorough performance testing

dentiny:
I use ZSTD since it's the default option for iceberg: https://iceberg.apache.org/docs/latest/configuration/#write-properties
But sure, I could use duckdb default config to make it compatible with v0.1 until we do perf test.

Issue #459, opened 2025-06-14
cache: implement state-based unit tests


dentiny:
## Summary

Considering the complicated state transfer, leverage TDD.
State transfer see https://docs.google.com/document/d/1KDuOOZDkJjrgFmTqFKgCr9-DVOk9W10ZZd19lijD7g8/edit?usp=sharing

A few test setup:
- A fake / mock cache handle

dentiny:
Least changes involved:
- Record per-file read reference count, record ongoing-read but deleted files
- Import new disk slice into data file cache at mooncake snapshot
- Unreference after iceberg snapshot completion
- Read request returns files involved (i.e. download remote, pin local cache), and unpin / dereference after request completion
- After compaction completes, check read reference count and delete if possible (could be a separate PR)

dentiny:
A second thought, we could simplify state transfer a lot by
- Loose constraint: no need to delete local files right after compaction finishes, we could unpin file in cache and get deleted after all references unpinned
- Treat read request and compaction as the same: use a certain file, and try best to leverage local file cache

Update state machine: https://docs.google.com/document/d/1f2d0E_Zi8FbR4QmW_YEhcZwMpua0_pgkaNdrqM1qh2E/edit?usp=sharing

PR #470, opened 2025-06-16
[storage] [1/N] Implement cache integration with iceberg persistence and snapshot read

dentiny:
## Summary

This PR integrate read-through/write-through cache into mooncake table.
- All mooncake tables share one single data file cache;
- Initially when writing local files, they're used as write-through cache;
- When iceberg snapshot persists, write through cache entries are unpinned and ready for eviction;
- On read path, moonlink tries to pin local cache entries at first; then download the whole file and add into cache;
- When read requests completes, all involved cache entries are unpinned and ready for eviction.

TODO items:
- Object cache should handle oversized objects
- Properly handle evicted deleted files
  + This PR has been super large (>2000 LOC), prefer not to add additional features
  + It's not a regression; data files are never deleted at the moment
- Cache should be integrated with compaction
- Unreference all data files at table drop

Test with pg_mooncake:
```sql
pg_mooncake (pid: 48561) =# DROP EXTENSION IF EXISTS pg_mooncake CASCADE;
DROP TABLE IF EXISTS r;
CREATE EXTENSION pg_mooncake;
SELECT count(*) FROM c;
SELECT count(*) FROM c;
DROP EXTENSION
DROP TABLE
CREATE EXTENSION
CREATE TABLE
CALL
INSERT 0 250000
 count  
--------
 250000
(1 row)

DELETE 125000
 count  
--------
 125000
(1 row)
```

## Related Issues

Related to https://github.com/Mooncake-Labs/moonlink/issues/459

## Checklist

- [x] Code builds correctly

### DIFF
--- a/src/moonlink/src/storage/mooncake_table/snapshot_read_output.rs
+++ b/src/moonlink/src/storage/mooncake_table/snapshot_read_output.rs
@@ -1,0 +1,135 @@
+use crate::storage::cache::object_storage::base_cache::CacheTrait;
+use crate::storage::cache::object_storage::object_storage_cache::ObjectStorageCache;
+use crate::storage::storage_utils::FileId;
+use crate::storage::PuffinDeletionBlobAtRead;
+use crate::table_notify::TableNotify;
+use crate::NonEvictableHandle;
+use crate::ReadState;
+
+use std::sync::Arc;
+
+use tokio::sync::mpsc::Sender;
+
+/// Mooncake snapshot for read.
+
+/// Pass out two types of data files to read.
+#[derive(Clone)]
+pub enum DataFileForRead {
+    /// Temporary data file for in-memory unpersisted data, used for union read.
+    TemporaryDataFile(String),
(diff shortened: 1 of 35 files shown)


src/moonlink/src/storage/mooncake_table/snapshot_read_output.rs
```diff
@@ -0,0 +1,99 @@
...
+    pub async fn take_as_read_state(mut self) -> Arc<ReadState> {
+        // Resolve remote data files.
+        let mut resolved_data_files = Vec::with_capacity(self.data_file_paths.len());
+        let mut cache_handles = vec![];
+        for cur_data_file in self.data_file_paths.into_iter() {
+            match cur_data_file {
+                DataFileForRead::TemporaryDataFile(file) => resolved_data_files.push(file),
+                DataFileForRead::RemoteFilePath((file_id, remote_filepath)) => {
+                    // TODO(hjiang):
+                    // 1. Delete evicted data files.
```
dentiny, line 67: Actual file deletion will be handle properly in the next PR (this one is already over 2000 LOC).

PR #486, opened 2025-06-17
[storage] Use temporary directory for test default cache

dentiny:
## Summary

This PR uses test-specific temporary directory to construct data file cache, instead of hard-coded fixed directory;
considering the fact that the hard-coded dir is re-created every construction, it's possible to suffer data race.

## Checklist

- [ ] Code builds correctly

### DIFF
--- a/src/moonlink/src/storage/cache/object_storage/cache_config.rs
+++ b/src/moonlink/src/storage/cache/object_storage/cache_config.rs
@@ -1,3 +1,6 @@
+#[cfg(test)]
+use tempfile::TempDir;
+
 /// Configuration for object storage cache.
 #[derive(Clone, Debug, PartialEq, Eq)]
 pub struct ObjectStorageCacheConfig {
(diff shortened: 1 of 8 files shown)


src/moonlink/src/storage/cache/object_storage/cache_config.rs
```diff
@@ -16,25 +19,13 @@ impl ObjectStorageCacheConfig {
     }
 
     /// Provide a default option for ease of testing.
+    /// It requires to take a testcase-unique temporary directory.
     #[cfg(test)]
-    pub fn default_for_test() -> Self {
+    pub fn default_for_test(temp_dir: &TempDir) -> Self {
         const DEFAULT_MAX_BYTES_FOR_TEST: u64 = 1 << 30; // 1GiB
-        const DEFAULT_CACHE_DIRECTORY: &str = "/tmp/moonlink_test_cache";
```
dentiny, line 22: `cargo test` executes all test cases in a parallel fashion, default cache recreation is subject to data race.

PR #488, opened 2025-06-17
Initial table copy

nbiscaro:
## Summary

Implements initial table copy to support creating columnstore tables from existing rowstore. 

There are two main components needed to support this
1) Copying the initial data:

When adding the table to the replication, we create a dedicated `PostgresSource` to use for the copy. We initialize this by getting the active slot (that was created by the replication connection) and exporting the current snapshot. In a separate task we get the `table_copy_stream` and read each `TableRow`. For each row in the stream, we create an `TableEvent::Append` to send over the tables `event_sender`. We also add an `is_copied: true` flag so the table handler knows not to buffer this event. When the copy completes we send a `FinishTableCopy`. 
(There are definitely some TODOs around optimizing this copy but these will be done in a follow up PR. The goal here is to get this working functionally). 

2) Buffering ongoing CDC events:

To avoid dropping any CDC events while waiting for the initial copy to finish, we must buffer these events and only apply them when the copy is finished. Conveniently, the logic we already have to support uncommitted streaming transactions is already very similar (they are almost functionally the same idea). We can reuse this logic with some slight modifications to also add support for buffering events during an initial table copy. 

The idea is as follows:
In the case of streaming transactions, not much needs to change. We already isolate these to their own mem slice and key them by `xact_id`. Really the only difference is that on commit, we don't actually want to commit these events since its possible the initial copy is still ongoing. In this case we call `prepare_commit` which does everything commit does up until the point of actually applying the changes to the snapshot. Instead, we take this `TransactionStreamCommit` and store it in a `initial_copy_buffered_commits`. Then, when we receive a `FinishTableCopy` we apply all buffered commits.

In the case of non-streaming transactions, we simply assign `INITIAL_COPY_XACT_ID` as the xact_id and treat it as a streamed transaction identical to the above.  

Another notable change -- for uninitialized values such as the `cached_lsn`, `last_read_lsn`, and `snapshot_lsn` we use `u64::Max` to indicate this instead of 0. The reason is that when initial copy completes its possible to have a table with data at LSN 0. 

## Related Issues

## Changes

- 
- 
- 

## Checklist

- [ ] Code builds correctly


src/moonlink/src/storage/iceberg/iceberg_table_event_manager.rs
```diff
@@ -42,7 +42,7 @@ impl IcebergTableEventManager {
     /// Drop an iceberg table.
     pub async fn drop_table(&mut self) -> Result<()> {
         self.table_event_tx
-            .send(TableEvent::DropTable)
+            .send(TableEvent::DropIcebergTable)
```
dentiny: Do you mind changing it back? I did something unrelated to iceberg.

zhousun:
two bigger comments:
1. How do we handle transactional consistency between copy and streaming records? Do we need some LSN coordination?
2. We can't apply deletes from ongoing cdc stream before initial copy is done.

nbiscaro:
> two bigger comments:
> 
> 1. How do we handle transactional consistency between copy and streaming records? Do we need some LSN coordination?
> 2. We can't apply deletes from ongoing cdc stream before initial copy is done.

1. When the replication slot is created for the connection we record the snapshot. When the copy happens we set the transaction to use this snapshot. So COPY contains all pre-snapshot rows, anything after should be buffered. 

2. Deletes during the initial copy are now buffered in the streaming `mem_slice` as `bufferered_deletes`. We apply the deletes when applying the rest of the commit when finalizing the initial copy.

src/moonlink/src/storage/mooncake_table.rs
```diff
@@ -867,6 +881,62 @@ impl MooncakeTable {
...
+
+        let commits = self
+            .initial_copy_buffered_commits
+            .drain(..)
+            .collect::<Vec<_>>();
+
+        self.in_initial_copy = false;
+        // Second: apply all buffered streaming transaction commits
+        for commit in commits {
+            let _ = self.commit_transaction_stream(commit).await;
```
dentiny, line 917: we should check returned result?

src/moonlink/src/table_handler.rs
```diff
@@ -194,29 +195,40 @@ impl TableHandler {
...
-                                    if table.should_transaction_flush(xact_id) {
-                                        if let Err(e) = table.flush_transaction_stream(xact_id).await {
-                                            error!(error = %e, "flush failed in append");
-                                        }
+                        TableEvent::Append { is_copied, row, xact_id } => {
+                            if is_copied || (!table.is_in_initial_copy() && xact_id.is_none()) {
+                                if let Err(e) = table.append(row) {
+                                    warn!(error = %e, "failed to append row");
+                                }
+                                if table.should_flush() {
```
dentiny, line 203: Why do we keep flushing with LSN 0 ??
It will cause all later snapshot to have LSN 0, for non initial copy appends

dentiny: To make it worse, flush LSN is recorded as data file flush LSN, which would regress after this PR

dentiny: I think for flush there're two invariants we should keep:
- Flush when the table is at a clean state, otherwise iceberg snapshot would be at an inconsistent state
- Flush LSN never regresses

dentiny: And for the current implementation, flushing in the middle of initial copy also doesn't work, because snapshot logic assume "only flush at a clean state".

In detail, when persist to iceberg, we only check flush LSN (and assume it's committed). So it would lead to partial write to iceberg, which puts our persistence layer in an inconsistent state.

dentiny: I plan to remove the whole flush logic here for both initial copy and normal non-streaming write.

If you want periodical flush, why not using streaming write buffer?

src/moonlink/src/storage/mooncake_table.rs
```diff
@@ -867,6 +881,62 @@ impl MooncakeTable {
...
+        self.table_snapshot_watch_sender.clone()
+    }
+
+    /// Enter initial copy mode. Subsequent CDC events will be
+    /// buffered in a dedicated streaming memslice until
+    /// `finish_initial_copy` is called.
+    /// In this case of a streaming transaction, we simply use the already provided `xact_id` to identify the transaction. In the case of non-streaming, we use `INITIAL_COPY_XACT_ID` to identify the transaction.
+    /// All commits are buffered and deferred until initial copy finishes.
+    pub fn start_initial_copy(&mut self) {
+        self.in_initial_copy = true;
```
dentiny, line 895: Shouldn't be an invariant check that 
```
in_initial_copy == true
initial_copy_buffered_commits is empty
```

src/moonlink/src/storage/mooncake_table.rs
```diff
@@ -867,6 +881,62 @@ impl MooncakeTable {
...
+        self.initial_copy_buffered_commits.clear();
+    }
+
+    /// Exit initial copy mode and commit all buffered changes.
+    pub async fn finish_initial_copy(&mut self) -> Result<()> {
+        if !self.in_initial_copy {
```
dentiny, line 901: curious any situation we could have `finish_initial_copy` called when not in initial copy state?

src/moonlink_backend/tests/test_initial_copy.rs
```diff
@@ -0,0 +1,660 @@
...
+                guard.database_id,
+                TABLE_ID,
+                DST_URI.to_string(),
+                format!("public.{table_name}"),
+                SRC_URI.to_string(),
+            )
+            .await
+            .unwrap();
+
+        // Delete one of the rows while copy is executing
```
dentiny, line 396: curious how do we make sure the copy is ongoing?

src/moonlink/src/storage/mooncake_table.rs
dentiny: If you commit multiple times, like this, snapshot flush LSN will be reset for a few times, which is also a bug.

Issue #471, opened 2025-06-16
iceberg: update namespace and table name for iceberg table


dentiny:
## Summary

```sh
root@fa1b818bb651:/var/lib/postgresql/data/mooncake# tree
.
├── default
│   ├── indicator.text
│   └── public.r
│       ├── data
│       │   ├── 019779de-c946-7362-96a8-69021b40bfa1-hash-index-v1-puffin.bin
│       │   ├── data-019779de-c925-75a0-b24f-c7a4a4fbc6bd.parquet
│       │   └── index_block_019779de-c92e-7862-9f10-02588504c9d6.bin
│       └── metadata
│           ├── 019779de-c948-7731-919a-d1ec5759ffff-m0.avro
│           ├── 019779de-c94e-7690-a08f-2001927bda63-m0.avro
│           ├── snap-2710404655640231827-0-019779de-c948-7731-919a-d1ec5759ffff.avro
│           ├── v0.metadata.json
│           ├── v1.metadata.json
│           └── version-hint.text
├── public.r
│   ├── data-019779de-c925-75a0-b24f-c7a4a4fbc6bd.parquet
│   └── index_block_019779de-c92e-7862-9f10-02588504c9d6.bin
└── var
    └── lib
        └── postgresql
            └── data
                └── mooncake

11 directories, 12 files
```
Current filesystem layout for the iceberg table, I hard-coded them at the second week, need to revisit.
For example, I need to include database name, schema name, and column store table name.

dentiny:
Already addressed.

PR #499, opened 2025-06-18
[storage] Update iceberg namespace to pg db schema

dentiny:
## Summary

Use db schema name as iceberg namespace, instead of hard-coded "default".

## Related Issues

Related to https://github.com/Mooncake-Labs/moonlink/issues/471

## Checklist

- [ ] Code builds correctly


dentiny:
Actually it's worth an official documentation, where and when an iceberg snapshot is created.

PR #508, opened 2025-06-18
[storage] No drop when ongoing snapshot

dentiny:
## Summary

With the newly added test, current implementation would fail with
```sh
thread 'table_handler::tests::test_iceberg_drop_table_with_data' panicked at src/moonlink/src/storage/mooncake_table.rs:960:9:
assertion failed: self.iceberg_table_manager.is_some()
```
This PR makes sure only drop table when no ongoing mooncake and iceberg snapshot.

## Checklist

- [x] Code builds correctly

### DIFF
--- a/src/moonlink/src/table_handler.rs
+++ b/src/moonlink/src/table_handler.rs
@@ -169,13 +169,19 @@ impl TableHandler {
 
         // Used to clean up mooncake table status, and send completion notification.
         let drop_table = async |table: &mut MooncakeTable| {
-            let res = table.shutdown().await;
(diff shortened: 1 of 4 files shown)


dentiny:
The dropped code coverage complained by codecov is error handling on drop table
image
I actually have a mock-based failure test before this PR:
https://github.com/Mooncake-Labs/moonlink/blob/681a6861c76f19d71cd17fca940770a41bc56b28/src/moonlink/src/table_handler/tests.rs#L1028-L1066

Issue #530, opened 2025-06-20
cache: unreference and delete cache on drop


dentiny:
## Summary

I need to not only unreference, but also delete cache entries on table drop.

PR #541, opened 2025-06-21
[storage] Integrate file index into object storage cache

dentiny:
## Summary

This PR is the final large piece for cache integration into moonlink, which hooks file indices.
Similar to the previous ones of data files and puffin files,
- It's imported into cache when snapshot eventloop
- File indices never gets evicted before deletion
- It's unreferenced and deleted after usage, like compaction and index merge

## Related Issues

Closes https://github.com/Mooncake-Labs/moonlink/issues/530
Closes https://github.com/Mooncake-Labs/moonlink/issues/543

## Checklist

- [x] Code builds correctly


src/moonlink/src/storage/compaction/compactor.rs
```diff
@@ -311,6 +311,7 @@ impl CompactionBuilder {
 
     /// Perform a compaction operation, and get the result back.
     #[tracing::instrument(name = "compaction_build", skip_all)]
+    #[allow(clippy::mutable_key_type)]
```
dentiny, line 314: I will make another PR to work this out.

Issue #546, opened 2025-06-22
[ISSUE] Data file being GC'ed too early


dpxcc:
## Summary

Briefly describe the task, feature, or problem.

https://github.com/Mooncake-Labs/pg_mooncake/blob/main/tests/pg_regress/sql/partitioned_table.sql was regressed that moonlink can return non-existing data files

This bug can likely be reproduced with simple UPDATE/DELETE, without partitioned tables

## Context

Why is this issue important? Link any relevant discussions, designs, or documents.

```sql
pg_mooncake (pid: 43760) =# CREATE EXTENSION pg_mooncake;
CREATE EXTENSION
pg_mooncake (pid: 43760) =# CREATE TABLE r (a int PRIMARY KEY, b TEXT) PARTITION BY RANGE (a);
CREATE TABLE r1 PARTITION OF r FOR VALUES FROM (0) TO (50);
CREATE TABLE r2 PARTITION OF r FOR VALUES FROM (50) TO (100);
CALL mooncake.create_table('c', 'r');
INSERT INTO r SELECT g, 'val_' || g from generate_series(0, 99) g;
DELETE FROM r WHERE a > 90;
SELECT count(*) FROM c WHERE a < 25;
SELECT count(*) FROM c WHERE a < 75;
SELECT count(*) FROM c WHERE a < 25 OR a > 75;
CREATE TABLE
CREATE TABLE
CREATE TABLE
CALL
INSERT 0 100
DELETE 9
 count 
-------
    25
(1 row)

ERROR:  (PGDuckDB/Duckdb_ExecCustomScan_Cpp) IO Error: Cannot open file "/tmp/moonlink_temp_file/inmemory_public.r_16863_37161112.parquet": No such file or directory
ERROR:  (PGDuckDB/Duckdb_ExecCustomScan_Cpp) IO Error: Cannot open file "/tmp/moonlink_temp_file/inmemory_public.r_16863_37161112.parquet": No such file or directory
```

## Tasks

- [ ] 

## Notes

Any extra notes or details that might be useful.

dpxcc:
Bisect to https://github.com/Mooncake-Labs/moonlink/commit/250b517b7c0a954191eb8b3c3b36216e1b948ace

dentiny:
I will take a look today.

dentiny:
I understand what's the issue: 
[...]

PR #550, opened 2025-06-22
[storage] Cache read state instead of snapshot read output

dentiny:
## Summary

The bug is: at `scan_table` every time we create a new `ReadState`, which deletes all temporary files after this particular variable goes out of scope.

Tested by pg_mooncake regression test:
```sql
PASS setup 24ms
PASS partitioned_table 663ms
PASS sanity 566ms
passed=3, failed=0
```

## Related Issues

Closes https://github.com/Mooncake-Labs/moonlink/issues/546

## Checklist

- [x] Code builds correctly

### DIFF
--- a/src/moonlink/src/union_read/read_state_manager.rs
+++ b/src/moonlink/src/union_read/read_state_manager.rs
@@ -25,7 +25,15 @@ impl ReadStateManager {
         let (table_snapshot, table_snapshot_watch_receiver) = table.get_state_for_reader();
         ReadStateManager {
             last_read_lsn: AtomicU64::new(0),
-            last_read_state: RwLock::new(Arc::new(ReadState::default())),
+            last_read_state: RwLock::new(Arc::new(ReadState::new(
+                /*data_files=*/ Vec::new(),
+                /*puffin_cache_handles=*/ Vec::new(),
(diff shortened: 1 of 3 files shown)


src/moonlink/src/union_read/read_state_manager.rs
```diff
@@ -145,12 +145,12 @@ impl ReadStateManager {
                     current_snapshot_lsn
                 };
 
-            let read_output = table_state_snapshot.request_read().await?;
+            let snapshot_read_output = table_state_snapshot.request_read().await?;
 
             self.last_read_lsn.store(effective_lsn, Ordering::Release);
-            *last_read_snapshot_output_guard = Arc::new(read_output);
+            *last_read_state_guard = snapshot_read_output.take_as_read_state().await;
```
dentiny, line 159: I recall why I cache `ReadOutput` instead of `ReadState` in the previous PR: `take_as_read_state` could involve IO operation if cache miss, it's not good because happen in a critical section, but not a problem for now because currently assume never goes out of disk and never do proactive eviction.

dentiny: it might be ok, because likely different `scan_table`'s asking for the same LSN

Issue #516, opened 2025-06-19
cache: avoid storing files twice for local filesystem


dentiny:
## Summary

This is a feature / issue reported by our user, which I will prioritize this week.

Basically we store files twice on local filesystem, one for iceberg snapshot, another for write/read cache, which is definitely not ideal for local filesystem use case.

My plan is to special handle local filesystem.
It depends on object storage cache fully integrated with moonlink.

PR #565, opened 2025-06-23
[storage] [1/N] Object storage cache support local filesystem optimization

dentiny:
## Summary

This PR is the first one to implement local filesystem file storage/cache optimization, to avoid double storage.
The overall design is:
- Optimization lies in caching layer, to avoid "check and optimize" logic scattered everywhere
- Optimization happens at import and fetch cache entries

One design design decision I made is:
- The ideal scenario is to add another pending state to replace local file cache with remote one, but that complicates state machine a lot, for example, we need to handle concurrent deletion and usage
- So for simplicity, the replacement only attempts once
- Extra local files could be deleted via compaction and cache eviction, so local filesystem is still shrinkable

I tested with pg_mooncake regression test:
```sh
PASS setup 28ms
PASS partitioned_table 651ms
PASS sanity 566ms
passed=3, failed=0
```

I tested with pg_mooncake SQL statements:
```sql
pg_mooncake (pid: 53328) =# DROP EXTENSION pg_mooncake CASCADE;
DROP TABLE r;                                                                                                                                       
CREATE EXTENSION pg_mooncake; 
INSERT INTO r (a, b) 
SELECT gs AS a, 'val_' || gs AS b FROM generate_series( (SELECT COALESCE(MAX(a), 0) + 1 FROM r), (SELECT COALESCE(MAX(a), 0) + 500000 FROM r) ) AS gs; 
SELECT count(*) AS total_rows FROM c;
DROP EXTENSION
DROP TABLE
CREATE EXTENSION
CREATE TABLE
CALL
INSERT 0 500000
 total_rows 
------------
     500000
(1 row)

pg_mooncake (pid: 53328) =# DROP EXTENSION IF EXISTS pg_mooncake CASCADE;
DROP TABLE IF EXISTS r;
CREATE EXTENSION pg_mooncake;
SELECT count(*) FROM c;
SELECT count(*) FROM c;
DROP EXTENSION
DROP TABLE
CREATE EXTENSION
CREATE TABLE
CALL
INSERT 0 250000
 count  
--------
 250000
(1 row)

DELETE 125000
 count  
--------
 125000
(1 row)
```

## Related Issues

Related to https://github.com/Mooncake-Labs/moonlink/issues/516

## Checklist

- [x] Code builds correctly

### DIFF
--- a/src/moonlink/src/storage/cache/object_storage/state_tests.rs
+++ b/src/moonlink/src/storage/cache/object_storage/state_tests.rs
@@ -118,6 +118,7 @@ async fn test_cache_2_requested_to_read_with_sufficient_space() {
     let mut cache = ObjectStorageCache::new(ObjectStorageCacheConfig {
         max_bytes: CONTENT.len() as u64,
         cache_directory: cache_file_directory.path().to_str().unwrap().to_string(),
+        optimize_local_filesystem: false,
     });
 
     // Import into cache first.
@@ -268,6 +269,7 @@ async fn test_cache_2_new_entry_with_sufficient_space() {
     let mut cache = ObjectStorageCache::new(ObjectStorageCacheConfig {
         max_bytes: (CONTENT.len() * 2) as u64,
         cache_directory: cache_file_directory.path().to_str().unwrap().to_string(),
+        optimize_local_filesystem: false,
     });
 
     // Import the first cache file.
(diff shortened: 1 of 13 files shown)


src/moonlink/src/storage/cache/object_storage/state_tests.rs
```diff
@@ -16,13 +17,14 @@
 /// (1) + create mooncake snapshot => (2)
 /// (1) + requested to read + sufficient space => (3)
 /// (2) + requested to read + sufficient space => (3)
-/// (2) + requested to read + insufficient space => (2)
```
dentiny, line 19: This is an invalid state!

PR #566, opened 2025-06-23
fix multi column key

zhousun:
## Summary
The code was calling equals_record_batch_at_offset with a different input 
(the function is expecting full row and caller calls with extracted values).

Fix the mismatch.
Also fix a small perf inefficiency (we don't need to read record batch when the key is stored in memory)
Also modified test config so we will hit the case.

## Related Issues

## Changes

- 
- 
- 

## Checklist

- [ ] Code builds correctly

### DIFF
--- a/src/moonlink/src/row/moonlink_row.rs
+++ b/src/moonlink/src/row/moonlink_row.rs
@@ -18,23 +18,19 @@ impl MoonlinkRow {
         Self { values }
     }
 
-    /// Apply identity properties projection for the current row.
-    fn project<'a>(&'a self, identity: &IdentityProp) -> Vec<&'a RowValue> {
+    fn is_extracted_identity_row(&self, identity: &IdentityProp) -> bool {
         match identity {
-            IdentityProp::SinglePrimitiveKey(idx) => vec![&self.values[*idx]],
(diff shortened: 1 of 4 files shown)


src/moonlink/src/row/moonlink_row.rs
```diff
@@ -18,23 +18,22 @@ impl MoonlinkRow {
...
         match identity {
-            IdentityProp::SinglePrimitiveKey(idx) => vec![&self.values[*idx]],
-            IdentityProp::Keys(indices) => indices.iter().map(|i| &self.values[*i]).collect(),
-            IdentityProp::FullRow => self.values.iter().collect(),
+            IdentityProp::SinglePrimitiveKey(_) => {
+                panic!("Single primitive key should not need any identity check")
+            }
+            IdentityProp::Keys(indices) => self.values.len() == indices.len(),
+            IdentityProp::FullRow => {
+                assert_eq!(self.values.len(), self.values.len());
```
dentiny: what's the purpose for the assertion?

zhousun:
We were calling equals_record_batch_at_offset with wrong input
(the function is expecting full row and caller calls with extracted values).

Fix the mismatch.
Also fix a small perf inefficiency (we don't need to read record batch when the key is stored in memory)

Modified test config so we will hit the case.

Issue #378, opened 2025-06-04
[ISSUE] Report iceberg LSN back to pg_mooncake


nbiscaro:
## Summary

Connect up the iceberg snapshot LSN and stop sending 0 back. 

## Context

Why is this issue important? Link any relevant discussions, designs, or documents.

## Tasks

- [ ] 

## Notes

Any extra notes or details that might be useful.

PR #621, opened 2025-06-27
Report iceberg flush LSN on status update

nbiscaro:
## Summary

Adds a watch channel for flush LSN updates in the event manager so that the replication layer can monitor when data is safely persisted to Iceberg. 

Modified the table handler to forward each snapshot’s flush LSN through this channel whenever an Iceberg snapshot completes. 

Updated the replication event loop to track flush LSNs from all tables and send the minimum confirmed value back to Postgres. 

Also removed the temporary fix of sending back the latest commit or keepalive lsn from the sink which now returns nothing. 

## Related Issues

Closes https://github.com/Mooncake-Labs/moonlink/issues/378

## Changes

- 
- 
- 

## Checklist

- [ ] Code builds correctly

### DIFF
--- a/src/moonlink/src/table_handler/tests.rs
+++ b/src/moonlink/src/table_handler/tests.rs
@@ -1015,6 +1015,117 @@ async fn test_multiple_snapshot_requests() {
     assert_eq!(visited, [true, true]);
 }
 
+/// Test that flush_lsn correctly reflects LSN ordering for batch operations
+#[tokio::test]
+async fn test_flush_lsn_ordering() {
+    let temp_dir = tempdir().unwrap();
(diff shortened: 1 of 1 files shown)


src/moonlink/src/table_handler.rs
```diff
@@ -386,6 +389,7 @@ impl TableHandler {
                             match iceberg_snapshot_result {
                                 Ok(snapshot_res) => {
                                     let iceberg_flush_lsn = snapshot_res.flush_lsn;
+                                    let _ = iceberg_event_sync_sender.flush_lsn_tx.send(iceberg_flush_lsn);
```
dentiny, line 392: qq: why not check the return value?

PR #647, opened 2025-06-30
Fix inconsistency between delete lsn & commit lsn

zhousun:
## Summary

Previously we expect delete_lsn > previous_commit_lsn and delete_lsn == current_commit_lsn.

Where both are not guaranteed.

Changed to
delete_lsn >= previous_commit_lsn, delete_lsn < current_commit_lsn

So commit will commit all deletes whose delete_lsn < commit_lsn.
deletes can see all previous_commits whose lsn <= delete_lsn

## Related Issues

## Changes

- 
- 
- 

## Checklist

- [ ] Code builds correctly

### DIFF
--- a/src/moonlink/src/storage/mooncake_table/snapshot.rs
+++ b/src/moonlink/src/storage/mooncake_table/snapshot.rs
@@ -156,13 +156,7 @@ impl SnapshotTableState {
     ) -> HashMap<MooncakeDataFileRef, BatchDeletionVector> {
         let mut aggregated_deletion_logs = HashMap::new();
         for cur_deletion_log in self.committed_deletion_log.iter() {
-            ma::assert_le!(
-                cur_deletion_log.lsn,
-                self.current_snapshot.snapshot_version,
-                "Committed deletion log {:?} is later than current snapshot LSN {}",
-                cur_deletion_log,
-                self.current_snapshot.snapshot_version
-            );
+            ma::assert_le!(cur_deletion_log.lsn, self.current_snapshot.snapshot_version);
             if cur_deletion_log.lsn >= flush_lsn {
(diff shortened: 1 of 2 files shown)


src/moonlink/src/storage/mooncake_table/snapshot.rs
```diff
@@ -195,8 +201,13 @@ impl SnapshotTableState {
 
         let old_committed_deletion_logs = std::mem::take(&mut self.committed_deletion_log);
         for cur_deletion_log in old_committed_deletion_logs.into_iter() {
-            ma::assert_le!(cur_deletion_log.lsn, self.current_snapshot.snapshot_version);
-            if cur_deletion_log.lsn > flush_point_lsn {
+            assert!(
```
dentiny: any chance to use more assertion? so you don't need to print the value out yourselve

src/moonlink/src/storage/mooncake_table/snapshot.rs
```diff
@@ -156,8 +156,14 @@ impl SnapshotTableState {
...
-            ma::assert_le!(cur_deletion_log.lsn, self.current_snapshot.snapshot_version);
-            if cur_deletion_log.lsn > flush_lsn {
+            ma::assert_le!(
+                cur_deletion_log.lsn,
+                self.current_snapshot.snapshot_version,
+                "Committed deletion log {:?} is later than current snapshot LSN {}",
+                cur_deletion_log,
+                self.current_snapshot.snapshot_version
+            );
+            if cur_deletion_log.lsn >= flush_lsn {
```
dentiny, line 160: curious why cannot we persist deletion log of `flush_lsn`?

Issue #665, opened 2025-07-01
recovery: resume CDC stream after recovery


dentiny:
## Summary

After https://github.com/Mooncake-Labs/moonlink/pull/657, mooncake/iceberg table is resumed and could take read requests, the next big missing piece is resuming cdc stream.

Issue #653, opened 2025-07-01
recovery: perform recovery operation when initializing moonlink backend


dentiny:
## Summary

Moonlink backend takes a mapping from source database uri to metadata storage database uri.
- connect to metadata storage with later
- check whether schema "mooncake" exists or not, if not, the db is not managed by mooncake
- load metadata (including table config, src table name, etc) from metadata store
- create mooncake and iceberg table
- re-establish replication if necessary

PR #670, opened 2025-07-02
[storage] Add unit test for operate on recovered table

dentiny:
## Summary

Followup PR for https://github.com/Mooncake-Labs/moonlink/pull/657
This PR is basically a no-op change, which adds more testing for recovery, to make sure mooncake tables work as usual after recovery.

## Related Issues

Closes https://github.com/Mooncake-Labs/moonlink/issues/653
Closes https://github.com/Mooncake-Labs/moonlink/issues/665

## Checklist

- [x] Code builds correctly

### DIFF
--- a/src/moonlink_connectors/src/replication_connection.rs
+++ b/src/moonlink_connectors/src/replication_connection.rs
@@ -41,6 +41,7 @@ pub enum Command {
     Shutdown,
 }
 
+/// Manages replication for table(s) within a database.
 pub struct ReplicationConnection {
     uri: String,
     table_base_path: String,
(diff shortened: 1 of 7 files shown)


src/moonlink_connectors/src/replication_connection.rs
```diff
@@ -150,12 +152,6 @@ impl ReplicationConnection {
                 table_name
             ))
             .await?;
-        self.postgres_client
```
dentiny, line 153: It's duplicate logic with `alter_table_replica_identity`.

Issue #686, opened 2025-07-03
recovery: handle duplicate records


dentiny:
## Summary

Chances are that:
- Mooncake table performs an iceberg snapshot
- pg_mooncake / moonlink crash before sending back confirmed flush LSN to postgres

When we do recovery, 
- Iceberg snapshot will be loaded into mooncake table
- Persisted cdc events are resent via cdc stream

PR #689, opened 2025-07-03
[storage] Discard already persisted content after recovery

dentiny:
## Summary

Moonlink sends back persisted flush LSN to pg, so WAL could be recycled.
However, chances are that before pg handles confirmed LSN, moonlink / pg_mooncake already crashes.

After load snapshot from iceberg table, we are able to recover up to flush LSN, but due to the situation described above, pg's still possible to resent persisted content which we need to handle.

In this PR, I discard already persisted content by checking LSN for 
- LSN for non-streaming writes in begin message
- LSN for stream write in stream commit message

Testing doc: https://docs.google.com/document/d/1Go02j9Nv2HrhABM-GBpBA4lPACJG3twC58J8pnKHAmA/edit?usp=sharing

## Related Issues

Closes https://github.com/Mooncake-Labs/moonlink/issues/686

## Checklist

- [x] Code builds correctly

### DIFF
--- a/src/moonlink/src/table_handler.rs
+++ b/src/moonlink/src/table_handler.rs
@@ -205,7 +205,7 @@ impl TableHandler {
                         // ==============================
                         //
                         TableEvent::Begin { lsn } => {
-                            current_non_streaming_txn_kept = lsn <= initial_persistence_lsn;
+                            current_non_streaming_txn_kept = lsn > initial_persistence_lsn;
                         }
                         TableEvent::Append { is_copied, row, xact_id } => {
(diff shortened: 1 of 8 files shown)


src/moonlink/src/table_handler.rs
```diff
@@ -193,7 +204,13 @@ impl TableHandler {
                         // Replication events
                         // ==============================
                         //
+                        TableEvent::Begin { lsn } => {
```
dentiny, line 207: One thing I missed in the PR is to hook begin event with cdc stream.

Issue #695, opened 2025-07-03
recovery: race condition between add schema to cdc stream in-memory state & pg resend


dentiny:
## Summary

During recovery, there're a few things to setup for postgres:
- re-establish cdc stream publication
- add table schema to cdc stream in-memory state, so later updates could recognize
- re-process unconfirmed cdc content, which is received from cdc stream

There's no guarantee on the order of (2) and (3), which means:
- if (3) happens before (2), all updates will suffer missing schema error

Issue #603, opened 2025-06-25
repl: moonlink errors with "missing schema for replication event"


dentiny:
## Summary

I tested repeated with the following command:
```sql
DROP EXTENSION IF EXISTS pg_mooncake CASCADE;
DROP TABLE IF EXISTS r;
CREATE EXTENSION pg_mooncake;
CREATE TABLE r (a int primary key, b text);
CALL mooncake.create_table('c', 'r');
INSERT INTO r SELECT a, 'val_' || a FROM generate_series(1, 10000000) AS a;
SELECT count(*) FROM c;
DELETE FROM r WHERE a % 2 = 0;
SELECT count(*) FROM c;
```
And I sometimes get 
```sh
2025-06-25T23:50:06.813673Z  WARN replication_event_loop: moonlink_connectors::replication_connection: missing schema for replication event
2025-06-25T23:50:06.813681Z  WARN replication_event_loop: moonlink_connectors::replication_connection: missing schema for replication event
```
and it never stops.

nbiscaro:
for later reference: noticing I get the same error in testing when the replication slot is not properly terminated and dropped. This may be a result of incomplete shutdown process.

PR #696, opened 2025-07-03
[postgres] Buffer replication messages when missing table schema

dentiny:
## Summary

As described at https://github.com/Mooncake-Labs/moonlink/issues/695, we currently suffer race condition on recovery (or moonlink restart). In short words, cdc events could be emitted before table schema has been set, leading to missing schema error.

This PR handles certain case with buffering replication messages in a buffer, and emit when corresponding schema arrives.

I tested manually with pg_mooncake:
- Write a few rows to mooncake table, and don't trigger iceberg snapshot, so everything's unconfirmed
- Shutdown everything, make sure all rows are replayed and inserted into the mooncake table

## Related Issues

Closes https://github.com/Mooncake-Labs/moonlink/issues/695
Closes https://github.com/Mooncake-Labs/moonlink/issues/603

## Checklist

- [x] Code builds correctly

### DIFF
--- a/src/moonlink_connectors/src/pg_replicate/conversions/cdc_event.rs
+++ b/src/moonlink_connectors/src/pg_replicate/conversions/cdc_event.rs
@@ -133,19 +133,17 @@ impl CdcEventConverter {
     fn get_column_schemas(
         table_id: SrcTableId,
         table_schemas: &HashMap<SrcTableId, TableSchema>,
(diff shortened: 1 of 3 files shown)


dentiny:
Copy from Zhou:
> we have very similar requirement for initial copy but solved in totally different way

I could followup with necessary changes.

Issue #692, opened 2025-07-03
bug: LSN assertion failure for initial copy


dentiny:
## Summary

Cannot reliably reproduce, but in a certain chance.
```sh
thread 'table_handler::tests::test_initial_copy_basic' panicked at src/moonlink/src/storage/mooncake_table/snapshot.rs:1061:17:
assertion failed: `(left < right)`
  left: `10`,
 right: `10`
```

dentiny:
This is a bug for initial copy:
https://github.com/Mooncake-Labs/moonlink/blob/8f853c8619f1807246b5d6eb14dae73a91851a27/src/moonlink/src/storage/mooncake_table.rs#L922-L927
we need to make sure there's at most one mooncake snapshot ongoing, otherwise bad things could happen.
For example, each mooncake snapshot comes with its own LSN, concurrent snapshot could bring LSN down.

Issue #688, opened 2025-07-03
bug: snapshot LSN assertion failure


dentiny:
## Summary

Error message:
```sh
thread 'tokio-runtime-worker' panicked at moonlink/src/moonlink/src/storage/mooncake_table/snapshot.rs:1061:17:
assertion failed: `(left < right)`
  left: `2459562880`,
 right: `2459562880`
stack backtrace:
   0: rust_begin_unwind
             at /rustc/05f9846f893b09a1be1fc8560e33fc3c815cfecb/library/std/src/panicking.rs:695:5
   1: core::panicking::panic_fmt
             at /rustc/05f9846f893b09a1be1fc8560e33fc3c815cfecb/library/core/src/panicking.rs:75:14
   2: more_asserts::inner::assert_failed_impl
             at /usr/local/cargo/registry/src/index.crates.io-1949cf8c6b5b557f/more-asserts-0.3.1/src/inner.rs:60:9
   3: more_asserts::inner::assert_failed_nomsg_impl
             at /usr/local/cargo/registry/src/index.crates.io-1949cf8c6b5b557f/more-asserts-0.3.1/src/inner.rs:25:5
   4: more_asserts::__private::assert_failed_nomsg
             at /usr/local/cargo/registry/src/index.crates.io-1949cf8c6b5b557f/more-asserts-0.3.1/src/lib.rs:76:9
   5: moonlink::storage::mooncake_table::snapshot::SnapshotTableState::update_snapshot::{{closure}}
             at /workspaces/pg_mooncake/moonlink/src/moonlink/src/storage/mooncake_table/snapshot.rs:1061:17
   6: moonlink::storage::mooncake_table::MooncakeTable::create_snapshot_async::{{closure}}
             at /workspaces/pg_mooncake/moonlink/src/moonlink/src/storage/mooncake_table.rs:1101:14
   7: <tracing::instrument::Instrumented<T> as core::future::future::Future>::poll
```

dentiny:
Only met the issue once for >100 times, sql I executed:
```sql
pg_mooncake (pid: 355348) =# DROP EXTENSION IF EXISTS pg_mooncake CASCADE;
DROP TABLE IF EXISTS c1, c2, r1, r2; 
CREATE EXTENSION pg_mooncake;
pg_mooncake (pid: 355348) =# CREATE TABLE r1 (a int primary key, b text);
CALL mooncake.create_table('c1', 'r1');
[...]

PR #699, opened 2025-07-03
Add checks for special case flush lsn

nbiscaro:
## Summary

As of https://github.com/Mooncake-Labs/moonlink/pull/488 we need to support the case where the `flush_lsn` is 0, since at the time of initial copy the table will contain data at LSN 0. There are some assertions that don't currently support this and are causing failures. We should add this special case to their checks.

## Related Issues

Closes https://github.com/Mooncake-Labs/moonlink/issues/692
Closes https://github.com/Mooncake-Labs/moonlink/issues/688

## Changes

- 
- 
- 

## Checklist

- [ ] Code builds correctly

### DIFF
--- a/src/moonlink/src/storage/mooncake_table.rs
+++ b/src/moonlink/src/storage/mooncake_table.rs
@@ -546,7 +546,7 @@ impl MooncakeTable {
         if contains_new_writes(iceberg_snapshot_res) {
             assert!(
                 self.last_iceberg_snapshot_lsn.is_none()
-                    || self.last_iceberg_snapshot_lsn.unwrap() < flush_lsn,
+                    || self.last_iceberg_snapshot_lsn.unwrap() < flush_lsn || flush_lsn == 0,
                 "Last iceberg snapshot LSN is {:?}, flush LSN is {:?}, imported data file number is {}, imported puffin file number is {}",
                 self.last_iceberg_snapshot_lsn,
                 flush_lsn,
--- a/src/moonlink/src/storage/mooncake_table/snapshot.rs
+++ b/src/moonlink/src/storage/mooncake_table/snapshot.rs
@@ -1057,7 +1057,10 @@ impl SnapshotTableState {
         // Assert and update flush LSN.
         if let Some(new_flush_lsn) = task.new_flush_lsn {
             // Assert flush LSN doesn't regress, if not force snapshot.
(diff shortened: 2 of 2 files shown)


src/moonlink/src/storage/mooncake_table.rs
```diff
@@ -546,7 +546,7 @@ impl MooncakeTable {
         if contains_new_writes(iceberg_snapshot_res) {
             assert!(
                 self.last_iceberg_snapshot_lsn.is_none()
-                    || self.last_iceberg_snapshot_lsn.unwrap() < flush_lsn,
+                    || self.last_iceberg_snapshot_lsn.unwrap() < flush_lsn || flush_lsn == 0,
```
dentiny, line 549: I don't think this is the correct way, you could still suffer two concurrent mooncake snapshots.
one for initial copy, another for later changes

dentiny: For example, one detailed state to set is flush LSN.
If you have two mooncake snapshot ongoing, it's possible to have
- snapshot A: initial copy mooncake snapshot ongoing, which indicates LSN 1
- snapshot B: you make some changes, and gets reflected to mooncake snapshot which indicates LSN 2

It's possible to have snapshot B handled before snapshot A, since both run at background threads
=> you suffer LSN regression

src/moonlink/src/storage/mooncake_table/snapshot.rs
```diff
@@ -1057,7 +1057,10 @@ impl SnapshotTableState {
         // Assert and update flush LSN.
         if let Some(new_flush_lsn) = task.new_flush_lsn {
             // Assert flush LSN doesn't regress, if not force snapshot.
-            if self.current_snapshot.data_file_flush_lsn.is_some() && !opt.force_create {
+            if self.current_snapshot.data_file_flush_lsn.is_some()
+                && !opt.force_create
+                && new_flush_lsn != 0
```
dentiny, line 1062: If you take a look at the error message here: https://github.com/Mooncake-Labs/moonlink/issues/688
the failed assertion is not of 0 LSN
```sh
thread 'tokio-runtime-worker' panicked at moonlink/src/moonlink/src/storage/mooncake_table/snapshot.rs:1061:17:
assertion failed: `(left < right)`
  left: `2459562880`,
 right: `2459562880`
```

PR #701, opened 2025-07-03
Rework mooncake snapshot for only once

dentiny:
Reverts Mooncake-Labs/moonlink#699

This is the reason why current implementation only works for one ongoing snapshot: https://github.com/Mooncake-Labs/moonlink/pull/699#discussion_r2183947525

And we don't need special handling for initial copy situation because it's always the first to create mooncake snapshot and iceberg snapshot (if applicable).
All assertions in the reverted PR should pass without problem.


dentiny:
CC @nbiscaro

Issue #716, opened 2025-07-04
[ISSUE] Insert 100K rows crashes


dpxcc:
## Summary

Briefly describe the task, feature, or problem.

```sql
CREATE EXTENSION pg_mooncake;
CREATE TABLE r (a int primary key, b text);
CALL mooncake.create_table('c', 'r');
INSERT INTO r SELECT g, 'val_' || g FROM generate_series(1, 100000) g;
```

```bash
thread 'tokio-runtime-worker' panicked at moonlink/src/moonlink/src/storage/mooncake_table/snapshot.rs:1061:17:
assertion failed: `(left < right)`
  left: `0`,
 right: `0`
stack backtrace:
   0: __rustc::rust_begin_unwind
             at /rustc/17067e9ac6d7ecb70e50f92c1944e545188d2359/library/std/src/panicking.rs:697:5
   1: core::panicking::panic_fmt
             at /rustc/17067e9ac6d7ecb70e50f92c1944e545188d2359/library/core/src/panicking.rs:75:14
   2: more_asserts::inner::assert_failed_impl
             at /usr/local/cargo/registry/src/index.crates.io-1949cf8c6b5b557f/more-asserts-0.3.1/src/inner.rs:60:9
   3: more_asserts::inner::assert_failed_nomsg_impl
             at /usr/local/cargo/registry/src/index.crates.io-1949cf8c6b5b557f/more-asserts-0.3.1/src/inner.rs:25:5
   4: more_asserts::__private::assert_failed_nomsg
             at /usr/local/cargo/registry/src/index.crates.io-1949cf8c6b5b557f/more-asserts-0.3.1/src/lib.rs:76:9
   5: moonlink::storage::mooncake_table::snapshot::SnapshotTableState::update_snapshot::{{closure}}
             at /workspaces/pg_mooncake/moonlink/src/moonlink/src/storage/mooncake_table/snapshot.rs:1061:17
   6: moonlink::storage::mooncake_table::MooncakeTable::create_snapshot_async::{{closure}}
             at /workspaces/pg_mooncake/moonlink/src/moonlink/src/storage/mooncake_table.rs:1102:14
   7: <tracing::instrument::Instrumented<T> as core::future::future::Future>::poll
             at /usr/local/cargo/registry/src/index.crates.io-1949cf8c6b5b557f/tracing-0.1.41/src/instrument.rs:321:9
   8: <core::pin::Pin<P> as core::future::future::Future>::poll
[...]

PR #717, opened 2025-07-04
[storage] Fix flush in non-streaming flush

dentiny:
## Summary

This PR removes flush in non-streaming write situations.

 There're two invariants for snapshot / table handler:
- Flush when the table is at a clean state
- Flush LSN only progresses and never regresses

Flush operation in the middle of a non-streaming transaction breaks both.

Flush inside of initial copy also breaks, but should be handled differently, because
- Normal non-streaming writes only happens for small volume of updates
- But initial copy could have indefinitely large number of updates, which cannot buffer all of them in memory.

As for initial copy appends, I considered a few possible fixes:
1. Flush LSN affects iceberg snapshot, but not mooncake snapshot; so an easy fix is to explicitly skip iceberg snapshot during initial copy operation. The code change is the least (3 LOC), but hard to reason.
2. Modify invariants for snapshot creation. Similar to https://github.com/Mooncake-Labs/moonlink/pull/699, update the invariants and special handle (aka, skip) initial copy.
3. (selected fix) Use streaming write, instead of non-streaming one to take initial copy writes, which works well under current assumptions.

I prefer (3) over (2) because (3) scopes all initial copy logic in table handler, instead of populating it elsewhere like mooncake snapshot / iceberg snapshots.

CC @nbiscaro 

All unit tests / integration tests have passed, and the failed sql statements linked in the issue work without issues.

Tested with pg_mooncake regression test
```sh
PASS setup 24ms
PASS partitioned_table 742ms
PASS sanity 733ms
passed=3, failed=0
```

## Related Issues

Closes https://github.com/Mooncake-Labs/moonlink/issues/716

## Checklist

- [x] Code builds correctly

### DIFF
--- a/src/moonlink/src/table_handler.rs
+++ b/src/moonlink/src/table_handler.rs
@@ -240,14 +240,11 @@ impl TableHandler {
                                 if let Err(e) = table.append(row) {
                                     warn!(error = %e, "failed to append row");
                                 }
-                                if table.should_flush() {
-                                    if let Err(e) = table.flush(0).await {
-                                        warn!(error = %e, "flush failed in append");
-                                    }
-                                }
+                                // TODO: optimize initial copy performance; current implement would buffer everything in memory.
                                 continue;
                             }
 
+                            // If mooncake is at initial copy state, buffer all changes to stream batches, which get applied when initial copy finished.
                             let xid = xact_id.unwrap_or(INITIAL_COPY_XACT_ID);
 
                             if let Err(e) = table.append_in_stream_batch(row, xid) {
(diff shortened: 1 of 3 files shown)


src/moonlink/src/table_handler.rs
```diff
@@ -240,14 +240,11 @@ impl TableHandler {
                                 if let Err(e) = table.append(row) {
                                     warn!(error = %e, "failed to append row");
                                 }
-                                if table.should_flush() {
```
dentiny, line 243: @nbiscaro 
I think this PR change could be a regression for initial copy, because we buffer all rows in the memory rather than flushing to disk.

There're two ways to resolve I think:
- Break the current invariant, to which doesn't persist iceberg snapshot for initial copy
- Or we could use streaming write to hold initial copied rows.

Issue #697, opened 2025-07-03
storage: index build is using async writer


zhousun:
## Summary

Index build is very very slow due to async functions.

## Context

Why is this issue important? Link any relevant discussions, designs, or documents.

## Tasks

- [ ] 

## Notes

Any extra notes or details that might be useful.

PR #721, opened 2025-07-05
[storage] Implement async bitwriter

dentiny:
## Summary

Currently we're using an async bitwriter, whichi is inperformant in a few aspects:
- Bits are written and flushed in bytes granularity, which means every bytes we write creates a background thread, and a syscall;
- All operations are async, which means writing a single bit involves tokio runtime rescheduling.

To resolve these two problems:
- I use a buffer to hold bits, so they're flushed every 4KiB
- Write interfaces are designed in cooperative mode, which means normal write is completely sync and only involves in-memory operations; flush operations should be manually called by users, based on the return value of write interfaces.

A few key data structures, like endian and bitqueue are still borrowed from async bitwriter; I would like to avoid bit manipulation ourselves. We could followup with dependency optimization later, to write our own and complete remove the  3rd party crate.
In the next PR, I will integrate with index builder and do a benchmark.

## Related Issues

Related to https://github.com/Mooncake-Labs/moonlink/issues/697

## Checklist

- [x] Code builds correctly

### DIFF
--- a/src/moonlink/src/storage/async_bitwriter.rs
+++ b/src/moonlink/src/storage/async_bitwriter.rs
@@ -1,0 +1,295 @@
+use more_asserts as ma;
+use std::io;
+use tokio::io::{AsyncWrite, AsyncWriteExt};
+use tokio_bitstream_io::{BitQueue, Endianness, Numeric};
+
+/// Buffer size, which controls the flush threshold.
+const BUFFER_SIZE: usize = 4096;
+
+/// [`BitWriter`] implement asynchronous write for bits; to avoid excessive runtime rescheduling caused by `await`,
+/// it takes cooperative interface, which means it requires users to explicit flush.
+#[allow(dead_code)]
+pub struct BitWriter<W: AsyncWrite + Unpin + Send + Sync, E: Endianness> {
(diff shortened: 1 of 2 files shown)


src/moonlink/src/storage/async_bitwriter.rs
```diff
@@ -0,0 +1,295 @@
+use more_asserts as ma;
+use std::io;
+use tokio::io::{AsyncWrite, AsyncWriteExt};
+use tokio_bitstream_io::{BitQueue, Endianness, Numeric};
+
+/// Buffer size, which controls the flush threshold.
+const BUFFER_SIZE: usize = 4096;
+
+/// [`BitWriter`] implement asynchronous write for bits; to avoid excessive runtime rescheduling caused by `await`,
```
dentiny, line 9: I found my latest doc change is not pushed successfully, will followup:
```rs
/// [`BitWriter`] implement asynchronous write for bits;
/// - To avoid excessive runtime rescheduling caused by `await`, it takes cooperative interface, which means it requires users to explicit flush.
/// - To avoid excessive syscalls, bits are buffered in-memory until it reaches flush threshold.
```

Issue #739, opened 2025-07-07
storage: reflect data compaction persistence into snapshot


dentiny:
## Summary

As titled, fix persistence results for data compaction.

PR #741, opened 2025-07-07
[storage] Reflect iceberg persistence result into snapshot

dentiny:
## Summary

This PR is a bug fix, that current implementation doesn't reflect iceberg persistence to mooncake snapshot.

## Related Issues

Closes https://github.com/Mooncake-Labs/moonlink/issues/739

## Checklist

- [ ] Code builds correctly

### DIFF
--- a/src/moonlink/src/storage/iceberg/compaction_tests.rs
+++ b/src/moonlink/src/storage/iceberg/compaction_tests.rs
@@ -274,15 +274,34 @@ async fn check_snapshot_reflects_persistence_for_compaction(
     table_data_file: &MooncakeDataFileRef,
     table_file_indices: Vec<FileIndex>,
 ) {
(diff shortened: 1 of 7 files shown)


src/moonlink/src/storage/mooncake_table/table_operation_test_utils.rs
```diff
@@ -260,6 +260,7 @@ async fn sync_mooncake_snapshot_and_create_new_by_iceberg_payload(
         skip_file_indices_merge: true,
         skip_data_file_compaction: true,
     }));
+    sync_mooncake_snapshot(table, receiver).await;
```
dentiny, line 263: This is a bug in the unit test, which could lead to flaky test, since we're not properly block wait mooncake snapshot completion.

PR #748, opened 2025-07-07
Deterministically add schemas before starting replication

nbiscaro:
## Summary

When recovering a table it's currently possible to receive CDC events before the table is configured on the moonlink side. Instead of buffering events until moonlink catches up, we can stall replication until all tables are configured. 

We do this by first calling `add_table` for each of the tables we want to add or recover. We keep track of each `src_uri` and once table addition is complete we start the replication for each connection. 

## Related Issues

## Changes

- 
- 
- 

## Checklist

- [ ] Code builds correctly

### DIFF
--- a/src/moonlink_backend/src/lib.rs
+++ b/src/moonlink_backend/src/lib.rs
@@ -91,8 +91,8 @@ where
         let table_id = mooncake_table_id.get_table_id_value();
 
         // Add mooncake table to replication, and create corresponding mooncake table.
+        let mut manager = self.replication_manager.write().await;
         let moonlink_table_config = {
-            let mut manager = self.replication_manager.write().await;
             manager
                 .add_table(
                     &src_uri,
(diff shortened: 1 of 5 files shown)


src/moonlink_backend/src/lib.rs
```diff
@@ -103,6 +103,7 @@ where
                 )
                 .await?
         };
+        manager.start_replication(&src_uri).await?;
```
dentiny, line 106: We don't need to hold the lock when accessing metadata storage?

src/moonlink_backend/src/recovery_utils.rs
```diff
@@ -89,6 +90,12 @@ where
 
         // Place into metadata store clients map.
         recovered_metadata_stores.insert(D::from(database_id), metadata_store_accessor);
+
+        unique_uris.insert(cur_metadata_store_uri);
+    }
+
+    for uri in unique_uris.into_iter() {
```
dentiny, line 97: I think this is not correct, metadata store is not source db.

PR #756, opened 2025-07-08
[storage] Filesystem accessor support copy

dentiny:
## Summary

The minimum object storage feature we need to support is copy local write-through cache to object storage.
I will integrate the API to iceberg table manager in the next PR.

## Checklist

- [x] Code builds correctly


src/moonlink/src/storage/filesystem/s3/tests.rs
```diff
@@ -0,0 +1,37 @@
...
+    let actual_content = filesystem_accessor
+        .read_object(&dst_filepath)
+        .await
+        .unwrap();
+    assert_eq!(actual_content, TEST_CONTEST);
+
+    // Delete all objects within the bucket.
+    filesystem_accessor.remove_directory("/").await.unwrap();
+    // Clean up test bucket.
+    delete_test_s3_bucket(bucket.clone()).await.unwrap();
```
dentiny, line 36: minio allows deletion without deleting all objects in the bucket, but fake GCS server doesn't.

PR #760, opened 2025-07-08
Add bug report title

dentiny:
## Summary

chatgpt told me title and description is necessary for templates, our current setup seems to be broken.
Unfortunately I don't have a way to test, since forked repo doesn't allow creating issues :(

## Checklist

- [ ] Code builds correctly


dentiny:
OK now we have bug report template:
image

Issue #743, opened 2025-07-07
[ISSUE] Buffer events for init copy


nbiscaro:
## Summary

Instead of trying to apply events on the fly we should simply buffer during initial copy. 

## Context

Why is this issue important? Link any relevant discussions, designs, or documents.

## Tasks

- [ ] 

## Notes

Any extra notes or details that might be useful.

PR #762, opened 2025-07-08
Buffer events during initial copy

nbiscaro:
## Summary

Rather than trying to apply the cdc events on the fly to an isolated mem slice, we simply buffer the `TableEvents` and replay them on `FinishInitialCopy`. This avoids difficult reasoning around edge cases and better aligns with Moonlink owning the WAL. For simplicity we buffer these in memory, but will eventually write these out to disk. 

## Related Issues

Closes https://github.com/Mooncake-Labs/moonlink/issues/743

## Changes

- Buffer incoming table events during initial copy
- Refactor event loop for easy replay
- Remove `prepare_stream_commit` and commit directly

## Checklist

- [ ] Code builds correctly

### DIFF
--- a/src/moonlink/src/storage/mooncake_table.rs
+++ b/src/moonlink/src/storage/mooncake_table.rs
@@ -55,9 +55,7 @@ use tokio::sync::mpsc::Sender;
 use tokio::sync::{watch, RwLock};
 use tracing::info_span;
 use tracing::Instrument;
-use transaction_stream::{
-    TransactionStreamCommit, TransactionStreamOutput, TransactionStreamState,
-};
+use transaction_stream::{TransactionStreamOutput, TransactionStreamState};
 
 /// Special transaction id used for initial copy append operation.
 pub(crate) const INITIAL_COPY_XACT_ID: u32 = u32::MAX - 1;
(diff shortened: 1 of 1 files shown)


src/moonlink/src/table_handler.rs
```diff
@@ -587,6 +457,189 @@ impl TableHandler {
...
+            TableEvent::StreamFlush { xact_id } => {
+                if let Err(e) = table.flush_transaction_stream(xact_id).await {
+                    error!(error = %e, "stream flush failed");
+                }
+            }
+            _ => {}
+        }
+    }
+
+    fn to_discard(lsn: u64, initial_persistence_lsn: Option<u64>) -> bool {
```
dentiny, line 620: I think next time when refactor it's good not to remove the comment...
I don't think the `to_discard` naming is clear enough alone...

nbiscaro: Then we should rename `to_discard` to be more clear.

dentiny: In the context of usage, it's clear; but reading the function alone is not, that's why I added a comment.
I think it's the responsibility for refactor-er to rename or leave comment?

You made the decision to discard it, I think that's for a reason?

PR #767, opened 2025-07-08
[storage] [1/N] Copy object from remote to local

dentiny:
## Summary

It's need for object storage cache on cache miss, and recovery.

## Checklist

- [x] Code builds correctly


src/moonlink/src/storage/filesystem/accessor/base_filesystem_accessor.rs
```diff
@@ -41,4 +43,7 @@ pub trait BaseFileSystemAccess: std::fmt::Debug + Send + Sync {
 
     /// Copy from local file [`src`] to remote file [`dst`].
     async fn copy_from_local_to_remote(&self, src: &str, dst: &str) -> Result<()>;
+
+    /// Copy from remote file [`src`] to local file [`dst`].
+    async fn copy_from_remote_to_local(&self, src: &str, dst: &str) -> Result<()>;
```
dentiny, line 48: I separate the copy feature into two APIs because opendal works on "root directory" concept.
To implement a generic copy function, you need two operators, each with different root.

PR #793, opened 2025-07-10
[storage] [1/N] Refactor state machine tests

dentiny:
## Summary

This PR is the first part of state-machine-based unit test refactoring, which aims to extract common functions to (1) prepare initial state, and (2) validate end state.

To reduce change set (it's already a non-trivial PR), I start with the largest test file, and this PR include half of the tests.
I will try to put separate PRs to refactor tests by tests.

## Checklist

- [x] Code builds correctly

### DIFF
--- a/src/moonlink/src/storage/mooncake_table/data_file_state_tests.rs
+++ b/src/moonlink/src/storage/mooncake_table/data_file_state_tests.rs
@@ -569,72 +569,31 @@ async fn test_3_read_and_read_over_and_pinned_3_without_local_filesystem_optimiz
         &temp_dir, /*optimize_local_filesystem=*/ false,
     );
 
-    let (mut table, mut table_notify) =
-        prepare_test_disk_file_for_read(&temp_dir, cache.clone(), use_batch_write).await;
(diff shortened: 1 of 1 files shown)


src/moonlink/src/storage/mooncake_table/data_file_state_tests.rs
```diff
@@ -141,79 +142,212 @@ async fn test_shutdown_table(#[case] use_batch_write: bool) {
...
-    #[case] use_batch_write: bool,
-) {
-    let temp_dir = tempfile::tempdir().unwrap();
-    let mut cache = create_infinite_object_storage_cache(&temp_dir, optimize_local_filesystem);
+/// Prepare state 3 for read usage: remote, local, in use
+///
+/// # Arguments
+///
+/// * read_then_snapshot: to reach state "remote" and "in use", we could either "persist then read", or "read then persist".
+async fn prepare_state_3(
```
dentiny, line 154: "remote, local, in use" is actually a complicated state:
- "use then persist" vs "persist and use" leads to different behavior
- whether enable local filesystem optimization decided whether to delete local files after persistence

Issue #787, opened 2025-07-09
pg: logging flooded with "missing schema" error


dentiny:
## Summary

Could be a bug.

My logging is again flooded with 
```sh
2025-07-09T18:24:33.980866Z  WARN replication_event_loop: moonlink_connectors::replication_connection: missing schema for replication event
2025-07-09T18:24:33.980870Z  WARN replication_event_loop: moonlink_connectors::replication_connection: missing schema for replication event
2025-07-09T18:24:33.980874Z  WARN replication_event_loop: moonlink_connectors::replication_connection: missing schema for replication event
2025-07-09T18:24:33.980878Z  WARN replication_event_loop: moonlink_connectors::replication_connection: missing schema for replication event
```
Repro:
```sql
pg_mooncake (pid: 24597) =# DROP EXTENSION IF EXISTS pg_mooncake CASCADE;
DROP TABLE IF EXISTS r;
CREATE EXTENSION pg_mooncake;
CREATE TABLE r (a int primary key, b text);
CALL mooncake.create_table('c', 'r');
INSERT INTO r SELECT a, 'val_' || a FROM generate_series(1, 2500) AS a;
NOTICE:  drop cascades to table c
DROP EXTENSION
DROP TABLE
CREATE EXTENSION
CREATE TABLE
CALL
INSERT 0 2500
```

Run the above SQL repeatedly without restart devcontainer.

dentiny:
moonlink commit: cd71d2671513eb865bbd2214de173f1c044ed917
pg_mooncake commit: d65268bc0ed7745d0fddfcbfae2215813181663b

nbiscaro:
This is a consequence of replaying the wal from the confirmed flush LSN. You can see that the events are for older table ids that have since been dropped.

PR #808, opened 2025-07-11
Remove warn for missing schema

nbiscaro:
## Summary

When replaying from `confirmed_flush_lsn` it's possible and valid to replay events from tables that have since been dropped. We should drop the `warn!` to avoid spamming. 

## Related Issues

Closes https://github.com/Mooncake-Labs/moonlink/issues/787

## Changes

- 
- 
- 

## Checklist

- [ ] Code builds correctly

### DIFF
--- a/src/moonlink_connectors/src/replication_connection.rs
+++ b/src/moonlink_connectors/src/replication_connection.rs
@@ -568,7 +568,6 @@ async fn run_event_loop(
 
                 match event_result {
                     Err(CdcStreamError::CdcEventConversion(CdcEventConversionError::MissingSchema(_))) => {
-                        warn!("missing schema for replication event");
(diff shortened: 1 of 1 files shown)


src/moonlink_connectors/src/replication_connection.rs
```diff
@@ -568,7 +568,6 @@ async fn run_event_loop(
 
                 match event_result {
                     Err(CdcStreamError::CdcEventConversion(CdcEventConversionError::MissingSchema(_))) => {
-                        warn!("missing schema for replication event");
```
dentiny, line 571: nit: it would be nice if we could leave a comment next time :(
people without context could be confused why we don't handle missing schema

PR #816, opened 2025-07-11
Async flush non streaming

nbiscaro:
## Summary

Spawn table flush as background task and send `TableEvent::FlushResult` on complete. There are some extra edge cases that need to be handled for the streaming case, so this is still handled by `flush_mem_slice`. We will unify these in a follow up PR. 

## Related Issues

## Changes

- Flush non streaming events in background
- Send `TableEvent::FlushResult`
- Add `set_flush_result` to set disk slices on complete
- Remove .await and update callers

## Checklist

- [ ] Code builds correctly

### DIFF
--- a/src/moonlink/src/storage/mooncake_table.rs
+++ b/src/moonlink/src/storage/mooncake_table.rs
@@ -45,7 +45,7 @@ use crate::NonEvictableHandle;
 use arrow::record_batch::RecordBatch;
 use arrow_schema::Schema;
 use delete_vector::BatchDeletionVector;
-pub(crate) use disk_slice::DiskSliceWriter;
+pub use disk_slice::DiskSliceWriter;
 use mem_slice::MemSlice;
 pub(crate) use snapshot::{PuffinDeletionBlobAtRead, SnapshotTableState};
 use std::collections::{HashMap, HashSet};
@@ -56,8 +56,8 @@ use table_snapshot::{IcebergSnapshotImportResult, IcebergSnapshotIndexMergeResul
 use tokio::sync::mpsc::Receiver;
 use tokio::sync::mpsc::Sender;
 use tokio::sync::{watch, RwLock};
-use tracing::info_span;
 use tracing::Instrument;
+use tracing::{error, info_span};
 use transaction_stream::{TransactionStreamOutput, TransactionStreamState};
 
 /// Special transaction id used for initial copy append operation.
@@ -692,6 +692,24 @@ impl MooncakeTable {
         self.next_snapshot_task.data_compaction_result = data_compaction_res;
     }
 
+    /// Set flush result, which will append the flushed disk slice into current task.
+    pub(crate) fn set_flush_result(
+        &mut self,
+        flush_result: Option<Result<DiskSliceWriter>>,
+        lsn: u64,
+    ) {
+        self.next_snapshot_task.new_flush_lsn = Some(lsn);
+        match flush_result {
+            Some(Ok(disk_slice)) => {
+                self.next_snapshot_task.new_disk_slices.push(disk_slice);
+            }
+            Some(Err(err)) => {
+                error!(error = ?err, "failed to flush mem slice");
+            }
+            None => {}
+        }
+    }
+
     /// Get iceberg snapshot flush LSN.
     pub(crate) fn get_iceberg_snapshot_lsn(&self) -> Option<u64> {
         self.last_iceberg_snapshot_lsn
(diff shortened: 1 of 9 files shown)


dentiny (commented):
Question: if you flush at commit, and there's deletion records in the transaction, only deletion records but not data files are persisted into iceberg?

src/moonlink/src/storage/mooncake_table.rs
```diff
@@ -827,15 +839,42 @@ impl MooncakeTable {
...
+                    path,
+                    batches,
+                    Some(lsn),
+                    next_file_id,
+                    index,
+                    parquet_flush_threshold_size,
+                );
+                let res = disk_slice.write().await;
+                table_notify_tx
+                    .send(TableEvent::FlushResult {
```
dentiny, line 877: Flush LSN controls what to persist, you should place it at result, rather than L834

src/moonlink/src/storage/iceberg/compaction_tests.rs
```diff
@@ -836,11 +836,11 @@ async fn test_compaction_2_3_1() {
     // Delete two rows and commit/flush/persist into iceberg.
     table.delete(rows[0].clone(), /*lsn=*/ 2).await; // Belong to the first data file.
     table.commit(/*lsn=*/ 3);
-    table.flush(/*lsn=*/ 3).await.unwrap();
+    table.flush(/*lsn=*/ 3).unwrap();
 
     table.delete(rows[2].clone(), /*lsn=*/ 4).await; // Belong to the second data file.
     table.commit(/*lsn=*/ 5);
-    table.flush(/*lsn=*/ 5).await.unwrap();
+    table.flush(/*lsn=*/ 5).unwrap();
```
dentiny: These are all interface change + behavior changes, and all UT assumes sync flush.

I strongly suggest for UT, it might be easier to provide a util function, which does (1) async flush; (2) background flush; (3) block wait flush completion; (4) set flush result.

Example: https://github.com/Mooncake-Labs/moonlink/blob/bc5e30263391b0d9b5d2189d4a112db6e245ac14/src/moonlink/src/storage/mooncake_table/table_operation_test_utils.rs#L239-L264

src/moonlink/src/storage/mooncake_table.rs
```diff
@@ -818,7 +830,7 @@ impl MooncakeTable {
     // This function
     // - tracks all record batches by current snapshot task
     // - persists all full batch records to local filesystem
-    pub async fn flush(&mut self, lsn: u64) -> Result<()> {
+    pub fn flush(&mut self, lsn: u64) -> Result<()> {
```
dentiny, line 831: Comment is not accurate any more.

> persists all full batch records to local filesystem

dentiny:
Another issue
- in the current implementation, all flush lsn only increases but not decreases
- but after we move flush io operation into background, later flush could finish and set flush lsn first

dentiny:
> Another issue
> 
> * in the current implementation, all flush lsn only increases but not decreases
> * but after we move flush io operation into background, later flush could finish and set flush lsn first

An example
- append row 1, commit and flush with lsn 1
- append row 2, commit and flush with lsn 2

So now we have two background flushes ongoing, it's possible to have the second flush finish first, and place the disk slide into mooncake, but not the first flush

src/moonlink/src/storage/mooncake_table.rs
nbiscaro: This is now set in `set_flush_result`

src/moonlink/src/storage/iceberg/compaction_tests.rs
nbiscaro: Resolved

src/moonlink/src/storage/mooncake_table.rs
nbiscaro: This still functionally does the exact same thing

nbiscaro:
> > Another issue
> > 
> > * in the current implementation, all flush lsn only increases but not decreases
> > * but after we move flush io operation into background, later flush could finish and set flush lsn first
> 
> An example
> 
> * append row 1, commit and flush with lsn 1
> * append row 2, commit and flush with lsn 2
> 
> So now we have two background flushes ongoing, it's possible to have the second flush finish first, and place the disk slide into mooncake, but not the first flush

Discussed offline, out of order is fine.

dentiny (commented):
Two things:
- when I request change, I mean I spot some severe problems, so I expect a review before merge. Usually non admib people are not able to merge w/ pending request to change.
- i don't see any new unit test added,. Ino you need at least two tests: flush failure cases, multiple flush operations not finishing in sequential order.

src/moonlink/src/storage/mooncake_table.rs
```diff
@@ -810,24 +828,61 @@ impl MooncakeTable {
     // This function
     // - tracks all record batches by current snapshot task
     // - persists all full batch records to local filesystem
-    pub async fn flush(&mut self, lsn: u64) -> Result<()> {
-        self.next_snapshot_task.new_flush_lsn = Some(lsn);
+    pub fn flush(&mut self, lsn: u64) -> Result<()> {
+        let table_notify_tx = self.table_notify.as_ref().unwrap().clone();
 
         if self.mem_slice.is_empty() {
+            self.next_snapshot_task.new_flush_lsn = Some(lsn);
```
dentiny, line 835: I don't understand this part as well, essentially you're setting flush LSN for twice, which prevents sanity check.

PR #822, opened 2025-07-11
TableHandler refactoring before enabling alter table 

zhousun:
## Summary

## Related Issues

## Changes

- 
- 
- 

## Checklist

- [ ] Code builds correctly

### DIFF
--- a/src/moonlink/src/table_handler.rs
+++ b/src/moonlink/src/table_handler.rs
@@ -32,14 +32,14 @@ pub struct EventSyncSender {
 }
 
 #[derive(PartialEq)]
-pub enum SpecialTableState {
+enum SpecialTableState {
     Normal,
     InitialCopy,
     _AlterTable,
     DropTable,
 }
 
-pub struct TableHandlerState {
+struct TableHandlerState {
     // cached table states
     //
     // Initial persisted LSN.
@@ -78,7 +78,7 @@ pub struct TableHandlerState {
 }
 
 impl TableHandlerState {
-    pub fn new() -> Self {
+    fn new() -> Self {
         Self {
             iceberg_snapshot_result_consumed: true,
             iceberg_snapshot_ongoing: false,
@@ -93,7 +93,7 @@ impl TableHandlerState {
         }
     }
 
-    pub fn update_table_lsns(&mut self, event: &TableEvent) {
+    fn update_table_lsns(&mut self, event: &TableEvent) {
         if event.is_ingest_event() {
             match event {
                 TableEvent::Commit { lsn, .. } => {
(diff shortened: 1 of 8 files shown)


src/moonlink/src/table_handler.rs
```diff
@@ -31,6 +31,150 @@ pub struct EventSyncSender {
     pub flush_lsn_tx: watch::Sender<u64>,
 }
 
+#[derive(PartialEq)]
+pub enum SpecialTableState {
```
dentiny: curious will it be used outside? 
i.e. I would imagine it to be "mooncake table" or "moonlink" internal thing, I think it should be `pub(crate)` rather than `pub`.

src/moonlink/src/table_handler.rs
```diff
@@ -31,6 +31,150 @@ pub struct EventSyncSender {
...
+    Normal,
+    InitialCopy,
+    _AlterTable,
+    DropTable,
+}
+
+pub struct TableHandlerState {
```
dentiny: Same question here, table handler is the internal eventloop implementation.
Also for all `TableHandlerState` member functions.

src/moonlink/src/table_handler.rs
```diff
@@ -262,48 +350,47 @@ impl TableHandler {
...
                             // Fast-path: no other concurrent events, directly clean up states and ack back.
-                            if !mooncake_snapshot_ongoing && !iceberg_snapshot_ongoing {
+                            if !table_handler_state.mooncake_snapshot_ongoing && !table_handler_state.iceberg_snapshot_ongoing {
                                 drop_table(&mut table, event_sync_sender).await;
                                 return;
                             }
 
                             // Otherwise, leave a drop marker to clean up states later.
-                            drop_table_requested = true;
+                            table_handler_state.start_drop_table();
```
dentiny: Hmm, do you think it's better to call it `request_drop_table` or `mark_drop_table`? `start_drop_table` makes me feel "drop table" operation will be started async-ly or sync-ly.

src/moonlink/src/table_handler.rs
```diff
@@ -262,48 +350,47 @@ impl TableHandler {
...
+                        }
+                        TableEvent::AlterTable { column_to_drop } => {
```
dentiny:
```suggestion
                        TableEvent::AlterTable { columns_to_drop } => {
```

src/moonlink/src/table_handler.rs
dentiny: My humble guess is you want to keep naming consistency for "drop table" and "initial copy",
but to me, "drop table" is a verb + noun, while "initial copy" is more like a "state".

src/moonlink_connectors/src/pg_replicate/moonlink_sink.rs
```diff
@@ -260,6 +262,38 @@ impl Sink {
                     relation_name = relation_body.name().unwrap_or("unknown"),
                     "Relation"
                 );
+                let src_table_id = relation_body.rel_id();
+                let cache_entry = self.relation_cache.get_mut(&src_table_id);
+                if let Some(cache_entry) = cache_entry {
+                    if cache_entry.columns().len() != relation_body.columns().len() {
+                        let event_sender = self.event_senders.get(&src_table_id);
+                        if let Some(event_sender) = event_sender {
+                            assert!(cache_entry.columns().len() > relation_body.columns().len());
```
dentiny, line 227:
```suggestion
                            ma::assert_gt!(cache_entry.columns().len(), relation_body.columns().len());
```

zhousun: so far pg_replicate is not using ma, so i am not changing it.
may revisit later.

src/moonlink/src/table_handler.rs
```diff
@@ -31,6 +31,150 @@ pub struct EventSyncSender {
...
+
+    fn update_table_lsns(&mut self, event: &TableEvent) {
+        if event.is_ingest_event() {
+            match event {
+                TableEvent::Commit { lsn, .. } => {
+                    self.latest_commit_lsn = Some(*lsn);
+                    self.table_consistent_view_lsn = Some(*lsn);
+                }
+                _ => {
+                    self.table_consistent_view_lsn = None;
```
dentiny, line 104: table consistent view LSN is used to indicate whether we could safely flush, so for table maintenance events like force snapshot, index merge, etc; it should keep unchanged

PR #823, opened 2025-07-11
[minor] Provide a sync flush API for testing

dentiny:
## Summary

This PR adds a new API to flush synchronously, so when flush becomes async, we only have to modify one test util function, but not everywhere.

## Checklist

- [x] Code builds correctly

### DIFF
--- a/src/moonlink/src/storage/mooncake_table/table_operation_test_utils.rs
+++ b/src/moonlink/src/storage/mooncake_table/table_operation_test_utils.rs
@@ -14,6 +14,21 @@ use crate::{MooncakeTable, SnapshotReadOutput};
 use crate::{ReadState, Result};
 
 /// ===============================
+/// Flush
+/// ===============================
+///
+/// Flush mooncake and block wait its completio.
+#[cfg(test)]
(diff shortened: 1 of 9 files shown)


src/moonlink/src/storage/mooncake_table/table_operation_test_utils.rs
```diff
@@ -13,6 +13,21 @@ use crate::table_notify::TableEvent;
...
+/// ===============================
+///
+/// Flush mooncake and block wait its completio.
+#[cfg(test)]
+pub(crate) async fn flush_table_and_sync(
+    table: &mut MooncakeTable,
+    _receiver: &mut Receiver<TableEvent>,
+    lsn: u64,
+) -> Result<()> {
+    // TODO(Nolan): Use receiver to block wait until table flush finishes, and set .
```
dentiny: @nbiscaro I imagine in your PR, this will be your only line of change, sth like
```rs
table.async_flush(lsn).await;
let flush_res = wait_for_completion(receiver).await;
table.set_flush_result(flush_res);
```

PR #862, opened 2025-07-15
Fix test dependency leak

dentiny:
## Summary

There's a dependency leak issue for iceberg-rust, I put a PR for it and got merged.

## Checklist

- [x] Code builds correctly

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1921,7 +1921,7 @@ dependencies = [
 [[package]]
 name = "iceberg"
 version = "0.5.1"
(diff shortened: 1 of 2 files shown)


Cargo.lock
```diff
@@ -1946,7 +1946,6 @@ dependencies = [
  "fnv",
  "futures",
  "itertools 0.13.0",
- "mockall",
```
dentiny, line 1949: `mockall` shouldn't appear in production code

Issue #447, opened 2025-06-12
index merge: provide a command to perform index merge


dentiny:
## Summary

Expose a function to perform index merge forcefully (similar to `force_iceberg_snapshot`).
- Allow users to checkin manually
- Easy for SQL-based testing to deterministically check index merge behavior

PR #865, opened 2025-07-15
[storage] Implement force index merge

dentiny:
## Summary

This PR implement force index merge; but it's worth noticing the function returns right after index merge completes, rather change persisted to iceberg table, worth a second thought.

## Related Issues

Closes https://github.com/Mooncake-Labs/moonlink/issues/447

## Checklist

- [x] Code builds correctly

### DIFF
--- a/src/moonlink/src/table_handler/test_utils.rs
+++ b/src/moonlink/src/table_handler/test_utils.rs
@@ -18,7 +18,7 @@ use iceberg::io::FileRead;
 use parquet::arrow::arrow_reader::ParquetRecordBatchReaderBuilder;
 use std::sync::Arc;
 use tempfile::{tempdir, TempDir};
-use tokio::sync::{mpsc, oneshot, watch};
+use tokio::sync::{broadcast, mpsc, oneshot, watch};
 
 /// Creates a `MoonlinkRow` for testing purposes.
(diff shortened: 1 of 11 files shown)


src/moonlink_backend/src/lib.rs
```diff
@@ -73,6 +73,22 @@ where
         Ok(())
     }
 
+    /// Perform an index merge operation, return when it completes.
+    /// Notice: the function will be returned right after index merge results reflected to mooncake snapshot, instead of being persisted into iceberg.
```
dentiny: I didn't return after iceberg snapshot because index merge + snapshot involves multiple iterations;
if people want to see how index merge results are reflected at iceberg, they could create another force snapshot command.

I will reconsider the behavior, it's not friendly for testing at least.

src/moonlink/src/table_handler/test_utils.rs
```diff
@@ -216,6 +221,24 @@ impl TestEnvironment {
         self.send_event(TableEvent::StreamAbort { xact_id }).await;
     }
 
+    /// Force an index meger, and block wait its completion.
+    pub async fn force_index_merge_and_sync(&self) {
```
dentiny, line 225: Just realized using `TableEventManager` for synchronization is much easier, will change in the followup.

Issue #858, opened 2025-07-14
[BUG] Our of order snapshots


nbiscaro:
## Description

Another issue

in the current implementation, all flush lsn only increases but not decreases
but after we move flush io operation into background, later flush could finish and set flush lsn first

An example

append row 1, commit and flush with lsn 1
append row 2, commit and flush with lsn 2
So now we have two background flushes ongoing, it's possible to have the second flush finish first, and place the disk slide into mooncake, but not the first flush

## Steps to Reproduce

1. 
2. 
3. 

## Version

- moonlink version (if applicable):
- pg_mooncake version (if applicable):

## Expected Behavior

What did you expect to happen?

## Actual Behavior

What actually happened?

## Additional Context

Any other information (logs, screenshots, configs) that might help.

dentiny:
Thanks for the issue!
I would suggest to add new unit tests [here](https://github.com/Mooncake-Labs/moonlink/blob/main/src/moonlink/src/storage/iceberg/tests.rs).
I'm still in the process of test refactoring, so if you find it hard to add new tests, let me know and I will refactor ASAP and unblock you. :)

Issue #857, opened 2025-07-14
[ISSUE] Add additional unit testing for async flush


nbiscaro:
## Summary

@dentiny 
I don't see any new unit test added,. Ino you need at least four tests: flush failure cases, multiple flush operations not finishing in sequential order, a row is committed but not flushed, a row deleted references to an unflushed row

## Context

Why is this issue important? Link any relevant discussions, designs, or documents.

## Tasks

- [ ] 

## Notes

Any extra notes or details that might be useful.

PR #868, opened 2025-07-15
Async flush snapshot order

nbiscaro:
## Summary

Given flush is now async it's possible for LSNs to complete out of order. This is fine in the mooncake snapshot case but we need to guarantee order when creating iceberg snapshot. 

To do this we maintain a set of pending flush LSNs and only allow iceberg snapshot if there are no pending flush LSNs < requested LSN. 

This needs more integration/unit testing and will be done as part of https://github.com/Mooncake-Labs/moonlink/issues/857, but would like to get in for correctness. 

## Related Issues

Closes https://github.com/Mooncake-Labs/moonlink/issues/858

## Changes

- 
- 
- 

## Checklist

- [ ] Code builds correctly

### DIFF
--- a/src/moonlink/src/storage/mooncake_table.rs
+++ b/src/moonlink/src/storage/mooncake_table.rs
@@ -48,7 +48,7 @@ use delete_vector::BatchDeletionVector;
 pub use disk_slice::DiskSliceWriter;
 use mem_slice::MemSlice;
 pub(crate) use snapshot::{PuffinDeletionBlobAtRead, SnapshotTableState};
-use std::collections::{HashMap, HashSet};
+use std::collections::{BTreeSet, HashMap, HashSet};
 use std::path::PathBuf;
 use std::sync::Arc;
 use table_snapshot::{IcebergSnapshotImportResult, IcebergSnapshotIndexMergeResult};
@@ -499,6 +499,9 @@ pub struct MooncakeTable {
 
     /// Table notifier, which is used to sent multiple types of event completion information.
     table_notify: Option<Sender<TableEvent>>,
+
+    // LSNs of pending flush operations.
+    pending_flush_lsns: BTreeSet<u64>,
 }
 
 impl MooncakeTable {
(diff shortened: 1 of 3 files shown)


dentiny (commented):
I think we both agree we should add some unit tests for iceberg snapshot?

dentiny:
> This needs more integration testing and will be done as part of https://github.com/Mooncake-Labs/moonlink/issues/857, but would like to get in for correctness.

Don't think that is the correct way.
To fix the problem, you should revert pr first.

src/moonlink/src/storage/mooncake_table.rs
```diff
@@ -840,6 +844,7 @@ impl MooncakeTable {
     // - tracks all record batches by current snapshot task
     // - persists all full batch records to local filesystem
     pub fn flush(&mut self, lsn: u64) -> Result<()> {
+        self.pending_flush_lsns.insert(lsn);
```
dentiny, line 847: Check return value

nbiscaro: What should we do with the return value?

dentiny: Check whether its a new value or an existing one

dentiny: Essentially you need to sanity check whether we have duplicate flush LSN

src/moonlink/src/storage/mooncake_table.rs
```diff
@@ -499,6 +499,9 @@ pub struct MooncakeTable {
 
     /// Table notifier, which is used to sent multiple types of event completion information.
     table_notify: Option<Sender<TableEvent>>,
+
+    // LSNs of pending flush operations.
+    pending_flush_lsns: BTreeSet<u64>,
```
dentiny, line 504: why do we put half of the flush logic in table handler, and another half in mooncake table?

dentiny: I think it's better to make sure the buffer flush logic is constraint in one place, either table or handler but not both.
It's good for coding understanding and for unit testing.

dentiny:
image
?????

Issue #448, opened 2025-06-12
compaction: provide a command to perform data file compaction


dentiny:
## Summary

Expose a function to perform data file compaction forcefully (similar to `force_iceberg_snapshot`).
- Allow users to checkin manually
- Easy for SQL-based testing to deterministically check compaction behavior

PR #873, opened 2025-07-15
[storage] Implement force data compaction

dentiny:
## Summary

Add a command to force data compaction, so we could test on pg_mooncake side.

## Related Issues

Closes https://github.com/Mooncake-Labs/moonlink/issues/448

## Checklist

- [x] Code builds correctly

### DIFF
--- a/src/moonlink/src/table_handler/test_utils.rs
+++ b/src/moonlink/src/table_handler/test_utils.rs
@@ -226,11 +226,18 @@ impl TestEnvironment {
         self.send_event(TableEvent::StreamAbort { xact_id }).await;
     }
 
-    /// Force an index meger, and block wait its completion.
+    /// Force an index merge operation, and block wait its completion.
     pub async fn force_index_merge_and_sync(&self) {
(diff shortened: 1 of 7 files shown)


src/moonlink/src/table_handler/test_utils.rs
```diff
@@ -221,13 +226,20 @@ impl TestEnvironment {
...
-    /// Force an index meger, and block wait its completion.
+    /// Force an index merge operation, and block wait its completion.
     pub async fn force_index_merge_and_sync(&self) {
         self.send_event(TableEvent::ForceIndexMerge).await;
         let mut index_merge_completion_rx = self.index_merge_completion_tx.subscribe();
         index_merge_completion_rx.recv().await.unwrap();
     }
 
+    /// Force a data compaction operation, and block wait its completion.
+    pub async fn force_data_compaction_and_sync(&self) {
```
dentiny, line 237: I actually considered using `TableEventManager` for command initiation and synchronization, the blocker I met is, `initial_xxx_command` takes a mutable reference which makes it hard to do concurrent testing...

PR #887, opened 2025-07-16
[storage] [1/N] Implement full table maintenance

dentiny:
## Summary

This PR is the first part for full table maintenance. TLDR:
- I treat full table maintenance as a special form of data compaction, since essentially there's no compaction difference
- As team discussed offline, force maintenance commands will only be executed for debugging, testing and admin purpose, so I disallow concurrent invocation for implementation simplicity

## Checklist

- [x] Code builds correctly

### DIFF
--- a/src/moonlink/src/table_handler.rs
+++ b/src/moonlink/src/table_handler.rs
@@ -1,6 +1,6 @@
 use crate::event_sync::EventSyncSender;
 use crate::storage::mooncake_table::DataCompactionResult;
-use crate::storage::mooncake_table::MaintainanceOption;
+use crate::storage::mooncake_table::MaintenanceOption;
 use crate::storage::mooncake_table::SnapshotOption;
 use crate::storage::mooncake_table::INITIAL_COPY_XACT_ID;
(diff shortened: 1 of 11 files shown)


src/moonlink/src/table_handler.rs
```diff
@@ -418,14 +427,25 @@ impl TableHandler {
                                 table_handler_state.pending_force_snapshot_lsns.entry(requested_lsn).or_default().push(tx);
                             }
                         }
-                        // Branch to trigger a force index merge request.
-                        TableEvent::ForceIndexMerge => {
+                        // Branch to trigger a force regular index merge request.
+                        TableEvent::ForceRegularIndexMerge => {
                             // TODO(hjiang): Handle cases where there're not enough file indices to merge.
-                            table_handler_state.force_index_merge_requested = true;
+                            assert_eq!(table_handler_state.index_merge_request_status, MaintenanceRequestStatus::Unrequested);
```
dentiny, line 433: These force table maintenance commands are only for testing, debugging and admin usage, so for implenentation simplicity I disallow concurrent usage. Could add back if necessary.

Issue #884, opened 2025-07-16
recovery: use sqlite for metadata storage on local deployment


dentiny:
## Summary

So we don't need to rely on postgres for user production usage.

PR #897, opened 2025-07-17
[storage] [4/N] Keep metadata storage integration with backend

dentiny:
## Summary

Second PR for rewriting metadata integration.

## Related Issues

Related to https://github.com/Mooncake-Labs/moonlink/issues/884

## Checklist

- [ ] Code builds correctly

### DIFF
--- a/src/moonlink_metadata_store/src/postgres/sql/create_secrets.sql
+++ b/src/moonlink_metadata_store/src/postgres/sql/create_secrets.sql
@@ -1,10 +1,10 @@
 -- SQL statements to store moonlink secret related fields.
 BEGIN;
 
-CREATE TABLE mooncake.secrets (
-    id SERIAL PRIMARY KEY,          -- unique row identifier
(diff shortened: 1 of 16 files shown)


src/moonlink_metadata_store/src/postgres/sql/create_secrets.sql
```diff
@@ -1,10 +1,10 @@
 -- SQL statements to store moonlink secret related fields.
 BEGIN;
 
-CREATE TABLE mooncake.secrets (
-    id SERIAL PRIMARY KEY,          -- unique row identifier
-    uid TEXT DEFAULT current_user,  -- user for the secret
-    oid oid,                        -- Mooncake table OID.
+CREATE TABLE secrets (
```
dentiny, line 4: Now the assumption for metadata storage has been completely changed:
- Previous assumption: pg is shared among (multiple) user and moonlink, so we need to apply security policy to make sure user-A is not able to access user-B's data;
- Current assumption: pg is owned by moonlink so no others could access it.

Issue #917, opened 2025-07-18
[ISSUE] Refactor postgres_source code


nbiscaro:
## Summary

This can be simplified and likely unnecessary connections being opened for ad-hoc queries. We can clean this up. 

## Context

Why is this issue important? Link any relevant discussions, designs, or documents.

## Tasks

- [ ] 

## Notes

Any extra notes or details that might be useful.

Issue #916, opened 2025-07-18
[ISSUE] Use binary format for initial copy


nbiscaro:
## Summary

Currently use text format, change to binary

## Context

Why is this issue important? Link any relevant discussions, designs, or documents.

## Tasks

- [ ] 

## Notes

Any extra notes or details that might be useful.

PR #918, opened 2025-07-18
Coordinate `pg_current_wal_lsn` with initial copy

nbiscaro:
## Summary

In order to better coordinate initial copy with incoming CDC events, we
1) Alter the publication with the new table
2) Fetch the `pg_current_wal_lsn` and start the `COPY TO` within the same transaction
3) Discard any events that have LSN < `pg_current_wal_lsn`

Since we add to the publication first and then start the copy we are at risk of having events both in the initial copy and sent as CDC event. We use the current_wal_lsn to dedup as it allows us to determine which events are already included in the initial copy. This was previously done in the reverse order, starting the initial copy and then altering the publication, however there is a window between when the copy starts and when the publication is altered where we may drop events. 

There are some other QOL improvements here such as a check for early return on initial copy (there are no rows). 

## Related Issues

Follow up: https://github.com/Mooncake-Labs/moonlink/issues/917
Follow up: https://github.com/Mooncake-Labs/moonlink/issues/916

## Changes

- Coordinate LSN with initial copy
- Early return if no rows

## Checklist

- [ ] Code builds correctly

### DIFF
--- a/src/moonlink_connectors/src/replication_connection.rs
+++ b/src/moonlink_connectors/src/replication_connection.rs
@@ -309,48 +309,54 @@ impl ReplicationConnection {
         )
         .await?;
 
-        // Add table to publication first to begin accumulating any cdc events.
-        // We can check where our initial copy started from and discard any rows we have already seen.
-        // We should add to publication regardless of whether we are recovering or not.
-        copy_source
-            .add_table_to_publication(&schema.table_name)
-            .await?;
-
         // Check if there are existing rows
         let row_count = copy_source.get_row_count(&schema.table_name).await?;
 
         // Only perform initial copy for new tables, not during recovery.
         // Early return if there are no rows to copy.
         if !is_recovery && row_count > 0 {
-            let stream = copy_source
-                .get_table_copy_stream(&schema.table_name, &schema.column_schemas)
-                .await?;
-            let start_lsn = stream.start_lsn.into();
-            if let Err(e) = event_sender_clone
-                .send(TableEvent::StartInitialCopy { start_lsn })
-                .await
-            {
+            if let Err(e) = event_sender_clone.send(TableEvent::StartInitialCopy).await {
                 error!(error = ?e, "failed to send StartInitialCopy event");
             }
 
+            // Alter the publication to add the table.
+            // Add table to publication first to begin accumulating any cdc events.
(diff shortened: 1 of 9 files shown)


src/moonlink_connectors/src/replication_connection.rs
```diff
@@ -315,28 +300,54 @@ impl ReplicationConnection {
...
 
-        // Only perform initial copy for new tables, not during recovery.
-        if !is_recovery {
-            // Create a dedicated source for the copy and register and snapshot the table.
-            let copy_source = PostgresSource::new(
-                &self.uri,
-                Some(self.slot_name.clone()),
-                TableNamesFrom::Vec(vec![schema.table_name.clone()]),
-            )
+        // Create a dedicated source for the copy
```
zhousun, line 303: i assume for non-replication we can skip these all together?

src/moonlink_connectors/src/replication_connection.rs
```diff
@@ -290,13 +282,6 @@ impl ReplicationConnection {
 
         let event_sender_clone = table_resources.event_sender.clone();
 
-        // Mark the table as being in initial copy, before receiving any events.
```
zhousun, line 293: It's important to startInitialcopoy before add_table_to_publication.

zhousun (commented):
otherwise lgtm

src/moonlink_connectors/src/replication_connection.rs
nbiscaro: Corrected sequence: 

send `StartInitialCopyEvent`
add table to publication with `ALTER PUB`

`BEGIN` a transaction on copy source connection
get `pg_current_wal_lsn`
Get the stream with `COPY OUT`
consume stream in while loop and send events
`COMMIT` the transaction

The ordering here is a little tricky: 
We need to send `StartInitialCopy` event BEFORE we `ALTER PUB` to ensure the table is in a blocking state
We need to get the stream and the current wal LSN in the same transaction
Once the stream copy starts, we can't `ALTER PUB` since `COPY TO` needs an `AccessShareLock` and `ALTER PUB` needs an `EXCLUSIVE LOCK` on both the publication and the table. 
This means we can't send the `start_lsn` in the `StartInitialCopy` event. This is fine, as it's really only necessary when applying the buffered events, so we can send it as part of `FinishInitialCopy`.

Issue #903, opened 2025-07-17
compaction: snapshot creation should consider index merge / compaction


dentiny:
## Summary

Assume such case:
- append, commit, flush row-1 with lsn 1
- append, commit, flush row-2 with lsn 2
- create iceberg snapshot, flush lsn = lsn 2
- after we do data compaction or index merge, when we do force snapshot, there's no way to force an iceberg snapshot, because the requested iceberg snapshot lsn has already been satisfied

dentiny:
For `force_snapshot`, the contract that we provide is: when user requests to have an iceberg snapshot, we block wait until iceberg snapshot flush lsn is >= requested

dentiny:
This is actually hard to do, the completed table maintenance results could be at multiple places
- buffered in snapshot task
- resides at current snapshot persistence buffer

It's hard to force snapshot call to guarantee both are cleared.

PR #919, opened 2025-07-18
[storage] Force table maintainence until persistence

dentiny:
## Summary

This PR changes the semantics for force table maintenance:
- Previously, it returns when maintenance results are placed back to snapshot buffer, so later query is possible to hit old snapshot;
- New behavior is to hold until changes are persisted into iceberg.

## Related Issues

Closes https://github.com/Mooncake-Labs/moonlink/issues/903

## Checklist

- [x] Code builds correctly

### DIFF
--- a/src/moonlink/src/table_handler.rs
+++ b/src/moonlink/src/table_handler.rs
@@ -54,7 +54,7 @@ enum MaintenanceRequestStatus {
 }
 
 #[derive(Clone, Debug, PartialEq)]
-enum MaintainanceRequestProcessStatus {
+enum MaintenanceProcessStatus {
     /// Force maintainence request is not being requested.
     Unrequested,
     /// Force maintainence request is being processed.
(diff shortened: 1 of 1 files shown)


src/moonlink/src/table_handler.rs
```diff
@@ -583,6 +703,17 @@ impl TableHandler {
                             table_handler_state.iceberg_snapshot_ongoing = false;
                             match iceberg_snapshot_result {
                                 Ok(snapshot_res) => {
+                                    // Update table maintainence operation status.
+                                    if !snapshot_res.index_merge_result.is_empty() && table_handler_state.table_maintenance_process_status == MaintainanceRequestProcessStatus::InPersist {
```
dentiny: TBD: simplify the verbose and length if-else logic

src/moonlink/src/table_handler.rs
```diff
@@ -597,10 +728,17 @@ impl TableHandler {
                                 Err(e) => {
                                     let err = Err(Error::IcebergMessage(format!("Failed to create iceberg snapshot: {e:?}")));
                                     if table_handler_state.has_pending_force_snapshot_request() {
-                                        if let Err(send_err) = table_handler_state.force_snapshot_completion_tx.send(Some(err)) {
+                                        if let Err(send_err) = table_handler_state.force_snapshot_completion_tx.send(Some(err.clone())) {
```
dentiny, line 721: TBD: add test case where iceberg snapshot fails for table maintenance result persistence

PR #935, opened 2025-07-19
[storage] Persist wal metadata at iceberg

dentiny:
## Summary

so on recovery we could load WAL metadata from iceberg; for flexibility I store a struct instead a single field.
Currently I store it in the table property, which is BAD; 
the ideal place should be snapshot summary, but due to the limitation of iceberg-rust I don't find better solution.

## Checklist

- [ ] Code builds correctly


dentiny:
CC @lohpaul9

PR #937, opened 2025-07-19
[storage] Iceberg schema evolution

dentiny:
## Summary

This PR implements schema evolution for iceberg. 
Since iceberg-rust doesn't support schema interface in transaction yet, I wrap everything into self-owned file catalog.

## Related Issues

Closes https://github.com/Mooncake-Labs/moonlink/issues/928

## Checklist

- [x] Code builds correctly

### DIFF
--- a/src/moonlink/src/storage/mooncake_table.rs
+++ b/src/moonlink/src/storage/mooncake_table.rs
@@ -1010,6 +1010,8 @@ impl MooncakeTable {
         );
     }
 
+    /// Update table schema to the provided [`updated_table_metadata`].
+    /// Notice: schema evolution performs IO operation in blocking style, and only returns when completion.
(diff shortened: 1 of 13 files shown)


src/moonlink/src/storage/mooncake_table.rs
```diff
@@ -966,6 +1010,24 @@ impl MooncakeTable {
         );
     }
 
+    /// Update table schema to the provided [`updated_table_metadata`].
+    /// Notice: schema evolution performs IO operation in blocking style, and only returns when completion.
+    #[allow(dead_code)]
+    pub(crate) async fn alter_table_schema(
```
dentiny, line 1016: @zhousun I assume this is the function you will integrate into table handler, iceberg schema evolution is implemented inside.

PR #957, opened 2025-07-22
[storage] Update iceberg update schema interface

dentiny:
## Summary

Current interface and semantics: users have to call `update_schema` in a blocking manner in foreground
New interface and semantics: schema change is considered as part of a normal iceberg persistence operation, requesting to update schema has nothing different than appending a row.

Expected usage:
```rs
table.update_schema(new_schema);
let iceberg_snapshot_payload = table.force_create_snapshot(...).await;
table.persist_iceberg_snapshot(iceberg_snapshot_payload);
// block wait its completion
```

There're two followups on my end:
- Add a unit test, which adds new rows after schema update
- Currently we can only perform schema evolution with new data files, we should also allow it with new content written

## Checklist

- [x] Code builds correctly

### DIFF
--- a/src/moonlink/src/storage/iceberg/file_catalog.rs
+++ b/src/moonlink/src/storage/iceberg/file_catalog.rs
@@ -676,7 +676,7 @@ impl SchemaUpdate for FileCatalog {
         &mut self,
         new_schema: IcebergSchema,
         table_ident: TableIdent,
-    ) -> IcebergResult<()> {
+    ) -> IcebergResult<Table> {
         let (_, old_metadata) = self.load_metadata(&table_ident).await?;
         let mut metadata_builder = old_metadata.into_builder(/*current_file_location=*/ None);
         metadata_builder = metadata_builder.add_current_schema(new_schema)?;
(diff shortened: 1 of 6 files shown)


src/moonlink/src/storage/mooncake_table/snapshot.rs
```diff
@@ -1019,10 +1026,10 @@ impl SnapshotTableState {
             file_indices_merge_payload = self.get_file_indices_to_merge(&opt.index_merge_option);
         }
 
-        // TODO(hjiang): Add whether to flush based on merged file indices.
+        // TODO(hjiang): When there's only schema evolution, we should also flush even no flush.
         if !opt.skip_iceberg_snapshot
             && self.current_snapshot.data_file_flush_lsn.is_some()
-            && (flush_by_new_files_or_maintainence || flush_by_deletion)
+            && (flush_by_new_files_or_maintainence || flush_by_deletion || flush_by_schema_change)
```
dentiny, line 1032: whenever there's schema evolution request, we should be able to create iceberg snapshot right away

Issue #968, opened 2025-07-22
wal / local disk wal


lohpaul9:
Write base class for WAL to be used with local disk

lohpaul9:
Closed by #969

PR #969, opened 2025-07-22
Wal prototype

lohpaul9:
## Summary
Adds WAL class, currently only enabled for local disk storage.

## Related Issues

#968 

## Changes
- Introduces new WAL events and WAL logic for inserting, flushing, deletion
- Converts between WAL events and supported TableEvents 

## Checklist

- [ X] Code builds correctly
- [ X] Tests have been added or updated

### DIFF
--- a/src/moonlink/src/storage.rs
+++ b/src/moonlink/src/storage.rs
@@ -9,7 +9,8 @@ pub(crate) mod mooncake_table;
 pub(crate) mod parquet_utils;
 pub(crate) mod path_utils;
 pub(crate) mod storage_utils;
-mod wal;
+#[allow(dead_code)]
+pub(crate) mod wal;
 
 pub use crate::event_sync::EventSyncReceiver;
 pub use cache::object_storage::cache_config::ObjectStorageCacheConfig;
--- a/src/moonlink/src/storage/iceberg/table_manager.rs
+++ b/src/moonlink/src/storage/iceberg/table_manager.rs
@@ -35,7 +35,7 @@ pub struct PersistenceResult {
 
 #[async_trait]
 #[cfg_attr(test, automock)]
-pub trait TableManager: Send {
+pub trait TableManager: Send + Sync {
     /// Write a new snapshot to iceberg table.
     /// It could be called for multiple times to write and commit multiple snapshots.
     ///
--- a/src/moonlink/src/storage/iceberg/table_manager.rs
+++ b/src/moonlink/src/storage/iceberg/table_manager.rs
@@ -35,7 +35,7 @@ pub struct PersistenceResult {
 
 #[async_trait]
 #[cfg_attr(test, automock)]
-pub trait TableManager: Send + Sync {
+pub trait TableManager: Send {
     /// Write a new snapshot to iceberg table.
     /// It could be called for multiple times to write and commit multiple snapshots.
     ///
--- a/src/moonlink/src/storage/wal.rs
+++ b/src/moonlink/src/storage/wal.rs
@@ -115,18 +115,18 @@ pub struct InMemWal {
     pub buf: Vec<WalEvent>,
     /// Tracks an LSN in case of a stream flush (or similar events) which has no accompanying LSN.
     /// A new instance will take the highest_lsn from the previous instance before it is populated.
-    pub highest_lsn: u64,
+    highest_lsn: u64,
 }
 
 impl InMemWal {
-    pub fn new(highest_lsn: u64) -> Self {
+    fn new(highest_lsn: u64) -> Self {
         Self {
             buf: Vec::new(),
             highest_lsn,
         }
     }
 
-    pub fn push(&mut self, table_event: &TableEvent) {
+    fn push(&mut self, table_event: &TableEvent) {
         let wal_event = WalEvent::new(table_event, self.highest_lsn);
         self.buf.push(wal_event);
 
@@ -140,10 +140,11 @@ impl InMemWal {
 
 /// Wal tracks both the in-memory WAL and the flushed WALs.
 /// Note that wal manager is meant to be used in a single thread. While
-/// persist and delete_files can be called concurrently, their results
-/// have to be handled serially.
+/// persist and delete_files can be called asynchronously, their results returned
+/// from those operations have to be handled serially.
 /// There is one instance of WalManager per table.
 pub struct WalManager {
+    /// In Mem Wal that gets appended to. When we need to flush, we call take on the buffer inside.
     pub in_mem_wal: InMemWal,
     /// The wal file numbers that are still live.
     live_wal_files_tracker: Vec<WalFileInfo>,
@@ -173,6 +174,7 @@ impl WalManager {
         format!("wal_{file_number}.json")
     }
 
+    /// Get the file info for the next wal file to be persisted.
     pub fn get_to_persist_wal_file_info(&self) -> WalFileInfo {
         WalFileInfo {
             file_number: self.curr_file_number,
@@ -180,6 +182,8 @@ impl WalManager {
         }
     }
 
+    /// Persist a series of wal events to the file system.
+    /// Should be called asynchronously after we call take on the in_mem_wal.
     pub async fn persist(
         file_system_accessor: Arc<dyn BaseFileSystemAccess>,
         wal_to_persist: &Vec<WalEvent>,
@@ -210,6 +214,7 @@ impl WalManager {
 
     /// Returns a list of files to be dropped, i.e. all files with highest_lsn < truncate_from_lsn
     /// List of files returned is sorted in ascending order of file number.
+    /// Should be called in preparation to asynchronously delete the files.
     pub fn get_files_to_truncate(&self, truncate_from_lsn: u64) -> Vec<WalFileInfo> {
         let last_truncate_idx = self
             .live_wal_files_tracker
@@ -227,6 +232,8 @@ impl WalManager {
         }
     }
 
+    /// Delete a list of wal files from the file system.
+    /// Should be called asynchronously after we call get_files_to_truncate.
     pub async fn delete_files(
         file_system_accessor: Arc<dyn BaseFileSystemAccess>,
         wal_file_numbers: &[WalFileInfo],
@@ -259,6 +266,7 @@ impl WalManager {
         }
     }
 
+    /// For now, we handle the persist and truncate results together.
     pub fn handle_completed_persist_and_truncate(
         &mut self,
         persist_and_truncate_result: &PersistAndTruncateResult,
--- a/src/moonlink/src/storage/wal.rs
+++ b/src/moonlink/src/storage/wal.rs
@@ -1,8 +1,4 @@
 pub(crate) mod wal_persistence_metadata;
-
-#[cfg(test)]
-mod test_utils;
-
 use crate::row::MoonlinkRow;
 use crate::storage::filesystem::accessor::base_filesystem_accessor::BaseFileSystemAccess;
 use crate::storage::filesystem::accessor::filesystem_accessor::FileSystemAccessor;
@@ -14,10 +10,9 @@ use futures::{future, StreamExt};
 use serde::{Deserialize, Serialize};
 use std::pin::Pin;
 use std::sync::Arc;
-use tokio::sync::Mutex;
 
 #[derive(Debug, Serialize, Deserialize, PartialEq)]
-enum WalEventEnum {
+pub(crate) enum WalEventEnum {
     Append {
         row: MoonlinkRow,
         xact_id: Option<u32>,
@@ -38,14 +33,25 @@ enum WalEventEnum {
     },
 }
 
+pub(crate) struct PersistAndTruncateResult {
+    pub file_persisted: Option<WalFileInfo>,
+    pub highest_deleted_file: Option<WalFileInfo>,
+}
+
+#[derive(Debug, Clone, PartialEq)]
+pub(crate) struct WalFileInfo {
+    pub file_number: u64,
+    pub highest_lsn: u64,
+}
+
 #[derive(Debug, Serialize, Deserialize, PartialEq)]
-struct WalEvent {
+pub(crate) struct WalEvent {
     pub lsn: u64,
     pub event: WalEventEnum,
 }
 
 impl WalEvent {
-    pub fn new_from_table_event(table_event: &TableEvent, last_highest_lsn: u64) -> Self {
+    pub fn new(table_event: &TableEvent, last_highest_lsn: u64) -> Self {
         // Try to get LSN from the table event, fallback to last_highest_lsn for events without LSN
         let lsn = table_event
             .get_lsn_for_ingest_event()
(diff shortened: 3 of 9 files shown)


src/moonlink/src/storage.rs
```diff
@@ -9,7 +9,8 @@ pub(crate) mod mooncake_table;
 pub(crate) mod parquet_utils;
 pub(crate) mod path_utils;
 pub(crate) mod storage_utils;
-mod wal;
+#[allow(dead_code)]
```
nbiscaro, line 12: why do we need this here?

src/moonlink/src/storage/iceberg/table_manager.rs
```diff
@@ -35,7 +35,7 @@ pub struct PersistenceResult {
 
 #[async_trait]
 #[cfg_attr(test, automock)]
-pub trait TableManager: Send {
+pub trait TableManager: Send + Sync {
```
dentiny: I don't think `IcebergTable` can be shared among multiple tasks/threads?

src/moonlink/src/storage/wal.rs
```diff
@@ -1 +1,307 @@
 pub(crate) mod wal_persistence_metadata;
+
+#[cfg(test)]
+mod test_utils;
```
nbiscaro: Place this at the bottom of the file

src/moonlink/src/storage/wal.rs
```diff
@@ -1 +1,307 @@
...
+
+/// Wal tracks both the in-memory WAL and the flushed WALs.
+/// The in-memory WAL may be modified concurrently with persist_and_truncate, but
+/// there is only meant to be one live call to persist_and_truncate at a time.
+pub struct Wal {
+    in_mem_wal: Mutex<InMemWal>,
+    /// The wal file numbers that are still live. Each entry is (file_number, highest_lsn within file).
+    /// Gets modified by persist and truncate, which are only called serially in the persist table handler.
+    live_wal_file_tracker: Mutex<Vec<(u64, u64)>>,
+    // Tracks the file number to be assigned to the next persisted wal file
```
dentiny: nit: `///` to keep consistent with comment above :)

src/moonlink/src/storage/wal.rs
```diff
@@ -1 +1,307 @@
...
+    Commit {
+        xact_id: Option<u32>,
+    },
+    StreamAbort {
+        xact_id: u32,
+    },
+    StreamFlush {
+        xact_id: u32,
+    },
+}
```
nbiscaro, line 35: Both `TableEvent::StreamFlush` and `TableEvent::Flush` seem to only be used in testing. These will eventually be removed as we shouldn't have test only events. Given this, I don't think you need to store `StreamFlush` events.

src/moonlink/src/storage/wal.rs
```diff
@@ -1 +1,307 @@
...
+use crate::Result;
+use futures::stream::{self, Stream};
+use futures::{future, StreamExt};
+use serde::{Deserialize, Serialize};
+use std::pin::Pin;
+use std::sync::Arc;
+use tokio::sync::Mutex;
+
+#[derive(Debug, Serialize, Deserialize, PartialEq)]
+enum WalEventEnum {
```
nbiscaro, line 16: Why are we not including LSN here?

src/moonlink/src/storage/wal.rs
```diff
@@ -1 +1,307 @@
...
+
+        // TODO(Paul): Implement streaming flush (if cross threshold, begin streaming write)
+    }
+
+    async fn take_and_replace(&self) -> (Option<Vec<WalEvent>>, u64) {
+        let mut in_mem_wal = self.in_mem_wal.lock().await;
+        if in_mem_wal.buf.is_empty() {
+            return (None, in_mem_wal.highest_lsn);
+        }
+        let mut new_buf = Vec::new();
```
dentiny:
```suggestion
        let new_buf = std::mem::take(&mut in_mem_wal.buf);
```

src/moonlink/src/storage/wal.rs
```diff
@@ -1 +1,307 @@
...
+    },
+    StreamFlush {
+        xact_id: u32,
+    },
+}
+
+#[derive(Debug, Serialize, Deserialize, PartialEq)]
+struct WalEvent {
+    pub lsn: u64,
+    pub event: WalEventEnum,
```
nbiscaro: Not every WalEvent has an LSN, for example `StreamAbort`. I think it make sense for the event itself to hold the LSN in this case.

src/moonlink/src/storage/wal.rs
nbiscaro, line 16: Missing `AlterTable`. 

This is a little different since AFAICT are not relying on the `Relation` message from postgres cdc, but I do think we would still need to record this in WAL. 

@zhousun can confirm.

src/moonlink/src/storage/wal/tests.rs
```diff
@@ -0,0 +1,496 @@
+use crate::storage::mooncake_table::test_utils::{test_row, TestContext};
```
nbiscaro, line 1: Rust convention is that unit tests should live at the bottom of the implementation .rs file. However, I don't think we are following this very closely in other areas of the repo so I'm hesitant to enforce it here. I'll leave it up to you as I'm personally fine with both.

src/moonlink/src/storage/wal.rs
```diff
@@ -1 +1,307 @@
...
+}
+
+impl WalEvent {
+    pub fn new_from_table_event(table_event: &TableEvent, last_highest_lsn: u64) -> Self {
+        // Try to get LSN from the table event, fallback to last_highest_lsn for events without LSN
+        let lsn = table_event
+            .get_lsn_for_ingest_event()
+            .unwrap_or(last_highest_lsn);
+
```
nbiscaro, line 60: Why does every table event need an LSN?

src/moonlink/src/storage/wal.rs
```diff
@@ -1 +1,307 @@
...
+}
+
+#[derive(Debug, Serialize, Deserialize, PartialEq)]
+struct WalEvent {
+    pub lsn: u64,
+    pub event: WalEventEnum,
+}
+
+impl WalEvent {
+    pub fn new_from_table_event(table_event: &TableEvent, last_highest_lsn: u64) -> Self {
```
nbiscaro: I think you can just call this `new()`

src/moonlink/src/storage/wal.rs
```diff
@@ -1 +1,307 @@
...
+            buf: Vec::new(),
+            highest_lsn,
+        }
+    }
+}
+
+/// Wal tracks both the in-memory WAL and the flushed WALs.
+/// The in-memory WAL may be modified concurrently with persist_and_truncate, but
+/// there is only meant to be one live call to persist_and_truncate at a time.
+pub struct Wal {
```
nbiscaro: Can we use a more informative name e.g `WalManager`?

src/moonlink/src/storage/wal.rs
```diff
@@ -1 +1,307 @@
...
+    in_mem_wal: Mutex<InMemWal>,
+    /// The wal file numbers that are still live. Each entry is (file_number, highest_lsn within file).
+    /// Gets modified by persist and truncate, which are only called serially in the persist table handler.
+    live_wal_file_tracker: Mutex<Vec<(u64, u64)>>,
+    // Tracks the file number to be assigned to the next persisted wal file
+    curr_file_number: Mutex<u64>,
```
nbiscaro: Do we need mutexes here? What is the case where we are reading/writing WAL for a table concurrently?

src/moonlink/src/storage/wal.rs
```diff
@@ -1 +1,307 @@
...
+    }
+
+    async fn take_and_replace(&self) -> (Option<Vec<WalEvent>>, u64) {
+        let mut in_mem_wal = self.in_mem_wal.lock().await;
+        if in_mem_wal.buf.is_empty() {
+            return (None, in_mem_wal.highest_lsn);
+        }
+        let mut new_buf = Vec::new();
+        std::mem::swap(&mut in_mem_wal.buf, &mut new_buf);
+        (Some(new_buf), in_mem_wal.highest_lsn)
```
nbiscaro: Use `take` instead of implementing yourself https://doc.rust-lang.org/beta/std/mem/fn.take.html

src/moonlink/src/storage/wal.rs
```diff
@@ -1 +1,307 @@
...
+
+    pub fn get_file_name(file_number: u64) -> String {
+        format!("wal_{file_number}.json")
+    }
+
+    #[cfg(test)]
+    /// Get the file system accessor for testing purposes
+    pub(crate) fn get_file_system_accessor(&self) -> Arc<dyn BaseFileSystemAccess> {
+        self.file_system_accessor.clone()
+    }
```
nbiscaro: Should go in test utils

src/moonlink/src/storage/wal.rs
```diff
@@ -1 +1,307 @@
...
+                            }
+                        };
+
+                        let table_events = filtered_wal_events
+                            .into_iter()
+                            .map(|wal| wal.into_table_event())
+                            .collect();
+                        Some((Ok(table_events), file_number + 1))
+                    }
+                    Err(e) => Some((Err(e), file_number + 1)),
```
nbiscaro, line 326: We deserialize from JSON -> WalEvent -> TableEvent. Can we deserialize directly into TableEvent here and avoid the extra hop?

src/moonlink/src/storage/wal.rs
dentiny: sidenote: do you have a way to stably flush? If not, flush might still be needed :(

src/moonlink/src/storage.rs
lohpaul9: most parts of wal (the WAL class, enum, etc) are only used in testing currently, will remove this in the next PR when it gets integrated to table handler

src/moonlink/src/storage/wal.rs
lohpaul9: Yes, but just for the WAL I wanted some way to order the events as they come in, so events that do not have an LSN will inherit the same LSN as the last event that came in. It makes the logic for tracking highest LSNs in files easier

src/moonlink/src/storage/wal.rs
lohpaul9: replied above

src/moonlink/src/storage/wal.rs
```diff
@@ -1 +1,307 @@
...
+                        if !exists {
+                            return None;
+                        }
+                    }
+                    Err(e) => {
+                        return Some((Err(e), file_number + 1));
+                    }
+                }
+
+                match file_system_accessor.read_object(&file_name).await {
```
dentiny, line 302: How big is the wal file? read_object reads the whole object at once

src/moonlink/src/storage/wal.rs
lohpaul9: The design is that persist and truncate is meant to be a long-running asynchronous operation, so the Wal class does undergo concurrent operations. We need to fulfill send + sync to send it around different threads. 

Technically, the only mutex we explicitly need is for in_mem_wal. The only real concurrency happens when we try to insert into the buffer and flush it at the same time. Therefore, live_wal_file_tracker and curr_file_number don't actually need mutexes, but I wrap vector and u64 with mutexes as a way to fulfil send + sync. The alternative i considered is UnsafeCell, but I don't think mutexes will add too much overhead here given that we don't ever compete for them.

src/moonlink/src/storage/wal.rs
lohpaul9: Its possible, but it would mean directly implementing some kind of de-serialization from JSON string to table event (might be even harder if eventually we use a different kind of serialization like bincode). I'm inclined to keep this for now and change it if we run into performance issues at recovery in the future.

src/moonlink/src/storage/wal.rs
lohpaul9: Will make change in future, opened issue

lohpaul9: #977

src/moonlink/src/storage/wal.rs
lohpaul9: it's about 500 MB, i'm going to implement this version first to read the entire file in, and then iterate on it if we have performance issues. This should also only be called on recovery

src/moonlink/src/storage/wal/tests.rs
lohpaul9: Ok. I think this unit test is getting quite long - will keep it here first

src/moonlink/src/storage/iceberg/table_manager.rs
lohpaul9: removed change

src/moonlink/src/storage/wal.rs
lohpaul9: Will make a clean-up change later to make sure we are including the correct events

src/moonlink/src/storage/wal.rs
lohpaul9: After discussion - WAL class will be not concurrent and kept in main table loop. asynch events will be triggered and any data needed will be passed in to them. Made changes as neccessary

src/moonlink/src/storage/wal.rs
```diff
@@ -1 +1,332 @@
...
+    },
+}
+
+pub(crate) struct PersistAndTruncateResult {
+    pub file_persisted: Option<WalFileInfo>,
+    pub highest_deleted_file: Option<WalFileInfo>,
+}
+
+#[derive(Debug, Clone, PartialEq)]
+pub(crate) struct WalFileInfo {
```
dentiny: qq: is this what you plan to store in persistence layer?

src/moonlink/src/storage/wal.rs
```diff
@@ -1 +1,332 @@
...
+    pub file_number: u64,
+    pub highest_lsn: u64,
+}
+
+#[derive(Debug, Serialize, Deserialize, PartialEq)]
+pub(crate) struct WalEvent {
+    pub lsn: u64,
```
dentiny: qq: why do we use `pub` for fields, but smaller visibility for struct `pub(crate)`?

src/moonlink/src/storage/wal.rs
```diff
@@ -1 +1,331 @@
...
+            TableEvent::StreamFlush { xact_id } => WalEventEnum::StreamFlush { xact_id: *xact_id },
+            _ => {
+                unimplemented!("Invalid table event for WAL: {:?}", table_event)
+            }
+        };
+
+        Self { lsn, event }
+    }
+
+    pub fn into_table_event(self) -> TableEvent {
```
dentiny: Also curious about this, `WalEvent` is with visibility `pub(crate)` but `into_table_event` is exposed as `pub`, wondering why is that?

src/moonlink/src/storage/wal.rs
lohpaul9: No, this is only used in memory right now. If you mean the iceberg metadata, for now it is just a single u64 of last flushed file number.

src/moonlink/src/storage/wal.rs
```diff
@@ -1 +1,331 @@
...
+            WalEventEnum::Commit { xact_id } => TableEvent::Commit {
+                xact_id,
+                lsn: self.lsn,
+            },
+            WalEventEnum::StreamAbort { xact_id } => TableEvent::StreamAbort { xact_id },
+            WalEventEnum::StreamFlush { xact_id } => TableEvent::StreamFlush { xact_id },
+        }
+    }
+}
+struct InMemWal {
```
dentiny: same question here, but `InMemWal` is private, curious why do we public fields?

src/moonlink/src/storage/wal.rs
```diff
@@ -1 +1,331 @@
...
+            highest_lsn,
+        }
+    }
+}
+
+/// Wal tracks both the in-memory WAL and the flushed WALs.
+/// Note that wal manager is meant to be used in a single thread. While
+/// persist and delete_files can be called concurrently, their results
+/// have to be handled serially.
+pub(crate) struct WalManager {
```
dentiny: qq: Is it per-table based, per-db based, or per-backend based?
(maybe worth a comment?)

src/moonlink/src/storage/wal.rs
```diff
@@ -1 +1,331 @@
...
+    in_mem_wal: InMemWal,
+    /// The wal file numbers that are still live.
+    live_wal_file_tracker: Vec<WalFileInfo>,
```
dentiny: nit: 
```suggestion
    live_wal_files_tracker: Vec<WalFileInfo>,
```

src/moonlink/src/storage/wal.rs
```diff
@@ -1 +1,331 @@
...
+        }
+    }
+
+    pub fn insert(&mut self, table_event: &TableEvent) {
+        let wal_event = WalEvent::new(table_event, self.in_mem_wal.highest_lsn);
+        self.in_mem_wal.buf.push(wal_event);
+
+        // Update highest_lsn if this event has a higher LSN
+        if let Some(lsn) = table_event.get_lsn_for_ingest_event() {
+            assert!(lsn >= self.in_mem_wal.highest_lsn);
```
dentiny:
```suggestion
            ma::assert_ge!(lsn, self.in_mem_wal.highest_lsn);
```
which prints out value if assertion fails

src/moonlink/src/storage/wal.rs
```diff
@@ -1 +1,331 @@
...
+        // TODO(Paul): Add a more robust constructor when implementing recovery
+        Self {
+            in_mem_wal: InMemWal::new(0),
+            live_wal_file_tracker: Vec::new(),
+            curr_file_number: 0,
+            file_system_accessor: Arc::new(FileSystemAccessor::new(config)),
+        }
+    }
+
+    pub fn insert(&mut self, table_event: &TableEvent) {
```
dentiny: nit (no strong preference): I usually see people using 
- `push`: append to end or prepend to front
- `insert`: add into a given position, with index provided

C++ vector reference:
- https://en.cppreference.com/w/cpp/container/vector/push_back.html
- https://en.cppreference.com/w/cpp/container/vector/insert.html

rust vector reference:
- https://doc.rust-lang.org/std/vec/struct.Vec.html#method.insert
- https://doc.rust-lang.org/std/vec/struct.Vec.html#method.push

src/moonlink/src/storage/wal.rs
```diff
@@ -1 +1,331 @@
...
+        wal_to_persist: &Vec<WalEvent>,
+        wal_file_info: &WalFileInfo,
+    ) -> Result<()> {
+        if !wal_to_persist.is_empty() {
+            let wal_json = serde_json::to_vec(&wal_to_persist).unwrap();
+
+            let wal_file_path = WalManager::get_file_name(wal_file_info.file_number);
+            file_system_accessor
+                .write_object(&wal_file_path, wal_json)
+                .await?;
```
dentiny, line 198: curious, does it happen in the background task, or in the eventloop, or anywhere?
mainly I would like to know if there're any feature requests for filesystem accessor

src/moonlink/src/storage/wal.rs
```diff
@@ -1 +1,331 @@
...
+        }
+        Ok(())
+    }
+
+    fn handle_complete_persist(&mut self, persist_and_truncate_result: &PersistAndTruncateResult) {
```
dentiny:
```suggestion
    fn handle_complete_persistence(&mut self, persist_and_truncate_result: &PersistAndTruncateResult) {
```

src/moonlink/src/storage/wal.rs
```diff
@@ -1 +1,331 @@
...
+        if let Some(file_persisted) = &persist_and_truncate_result.file_persisted {
+            assert!(file_persisted.file_number == self.curr_file_number);
```
dentiny:
```suggestion
            assert_eq!(file_persisted.file_number, self.curr_file_number);
```

src/moonlink/src/storage/wal.rs
```diff
@@ -1 +1,331 @@
...
+            // Add the persisted file to the live tracker
+            self.live_wal_file_tracker.push(file_persisted.clone());
+            self.curr_file_number += 1;
+        }
+    }
+
+    pub fn get_files_to_truncate(&self, truncate_from_lsn: u64) -> Vec<WalFileInfo> {
```
dentiny, line 218: nit: this function took me some time to read through, do you mind adding a short comment?
(not merge blocker)

src/moonlink/src/storage/wal.rs
lohpaul9: per table. added a comment

src/moonlink/src/storage/wal.rs
lohpaul9: re-worked private fields

src/moonlink/src/storage/wal.rs
lohpaul9: changed to push

src/moonlink/src/storage/wal.rs
lohpaul9: this happens as a background task. for now, shouldn't be any new feature requests for filesystem accessor

src/moonlink/src/storage/wal.rs
```diff
@@ -1 +1,331 @@
...
+                .iter()
+                .take(idx + 1)
+                .cloned()
+                .collect::<Vec<WalFileInfo>>()
+        } else {
+            Vec::new()
+        }
+    }
+
+    pub async fn delete_files(
```
dentiny, line 237: just for my own curiosity, when / where do you plan run delete files?

src/moonlink/src/storage/wal.rs
```diff
@@ -1 +1,331 @@
...
+        wal_file_numbers: &[WalFileInfo],
+    ) -> Result<()> {
+        let file_names = wal_file_numbers
+            .iter()
+            .map(|wal_file_info| WalManager::get_file_name(wal_file_info.file_number))
+            .collect::<Vec<String>>();
+        let delete_futures = file_names
+            .iter()
+            .map(|file_name| file_system_accessor.delete_object(file_name));
+        future::try_join_all(delete_futures).await?;
```
dentiny: is this error for join or for deletion?

src/moonlink/src/storage/wal.rs
lohpaul9: delete_files and persist are both run as asynch events spawned out of the table handler event loop

src/moonlink/src/storage/wal/tests.rs
```diff
@@ -0,0 +1,429 @@
...
+async fn test_wal_insert_persist_files() {
+    let context = TestContext::new("wal_persist");
+    let (mut wal, expected_events) = create_test_wal(&context).await;
+
+    // Persist and verify file number
+    wal.persist_and_truncate(None).await.unwrap();
+
+    // Check file exists and has content
+    let wal_file_path = context.path().join("wal_0.json");
+    assert!(wal_file_path.exists());
```
dentiny: it's a sync IO function?

lohpaul9: just returns bool

src/moonlink/src/storage/wal/tests.rs
```diff
@@ -0,0 +1,429 @@
...
+    let context = TestContext::new("wal_empty_persist");
+    let mut wal = WalManager::new(FileSystemConfig::FileSystem {
+        root_directory: context.path().to_str().unwrap().to_string(),
+    });
+
+    // Persist without any events
+    wal.persist_and_truncate(None).await.unwrap();
+
+    // No file should be created for empty WAL
+    let wal_file_path = context.path().join("wal_0.json");
```
dentiny: nit: we could check no file exists in the directory?
And `exists` seems to be sync IO call

src/moonlink/src/storage/wal/tests.rs
```diff
@@ -0,0 +1,429 @@
...
+        row: row.clone(),
+        xact_id: None,
+        lsn: 102,
+        is_copied: false,
+    };
+    wal.insert(&event3);
+    wal.persist_and_truncate(None).await.unwrap();
+
+    // Verify files exist
+    assert!(context.path().join("wal_0.json").exists());
```
dentiny: ditto

src/moonlink/src/storage/wal/tests.rs
```diff
@@ -0,0 +1,429 @@
...
+    assert!(context.path().join("wal_1.json").exists());
+    assert!(context.path().join("wal_2.json").exists());
+
+    // Use the new helper infrastructure to verify file contents
+    let expected_wal_events_0 = convert_to_wal_events_vector(vec![event1]);
```
dentiny: nit: could use vector and loop here :stuck_out_tongue_winking_eye:

src/moonlink/src/storage/wal.rs
```diff
@@ -1 +1,331 @@
...
+        // Update highest_lsn if this event has a higher LSN
+        if let Some(lsn) = table_event.get_lsn_for_ingest_event() {
+            assert!(lsn >= self.in_mem_wal.highest_lsn);
+            self.in_mem_wal.highest_lsn = lsn;
+        }
+
+        // TODO(Paul): Implement streaming flush (if cross threshold, begin streaming write)
+    }
+
+    fn take(&mut self) -> Vec<WalEvent> {
```
dentiny: nit: do you mind renaming the function a little? 
let's say `std::mem::take(x)`, it means take the ownership of `x` and replace with default.
On usage my first intuition of `self.take()` is `self`'s ownership shifted but we're only getting events here

src/moonlink/src/storage/wal.rs
dentiny, line 218: Need documentation that returns files are sorted

src/moonlink/src/storage/wal/tests.rs
dentiny: I mean, this is an async function, while `exists` is a sync IO call

src/moonlink/src/storage/wal.rs
nbiscaro: Does this need a function? We are wrapping take in take? I think you can just do `take(&mut self.in_mem_wal.buf)` at call site.

src/moonlink/src/storage/wal/test_utils.rs
```diff
@@ -0,0 +1,239 @@
...
+
+    /// Mock function for how this gets called in table handler. We retrieve some data from the wal,
+    /// asynchronously update the file system,
+    /// and then call handle_completed_persist_and_truncate to update the wal.
+    pub async fn persist_and_truncate(
+        &mut self,
+        last_iceberg_snapshot_lsn: Option<u64>,
+    ) -> Result<()> {
+        // handle the persist side
+        let wal_to_persist = self.take();
```
nbiscaro: wrt my previous comment, we should just use `take` directly instead of this custom function. I have no idea what this is "taking" just by seeing the call site.

src/moonlink/src/storage/wal.rs
```diff
@@ -1 +1,343 @@
...
+                            .collect();
+                        Some((Ok(table_events), file_number + 1))
+                    }
+                    Err(e) => Some((Err(e), file_number + 1)),
+                }
+            }
+        }))
+    }
+
+    pub fn recover_flushed_wals_flat(
```
dentiny, line 334: nit: do you mind adding a comment on `begin_from_lsn`? i.e. is it inclusive or exclusive?

src/moonlink/src/storage/wal/tests.rs
```diff
@@ -0,0 +1,429 @@
...
+        };
+        wal.push(&event);
+        wal.persist_and_truncate(None).await.unwrap();
+    }
+
+    // Truncate from LSN 102 (should delete files 0, 1 - files with LSN < 102)
+    wal.persist_and_truncate(Some(102)).await.unwrap();
+
+    // Verify files 0, 1 are deleted
+    assert!(!context.path().join("wal_0.json").exists());
```
dentiny: ditto

dentiny (commented):
current mooncake / snapshot place task spawn logic inside of mooncake.rs, instead of table handler:
https://github.com/Mooncake-Labs/moonlink/blob/58b6116feffa5fa9d870ff96b0201e26f3948543/src/moonlink/src/storage/mooncake_table.rs#L1035-L1043

https://github.com/Mooncake-Labs/moonlink/blob/58b6116feffa5fa9d870ff96b0201e26f3948543/src/moonlink/src/storage/mooncake_table.rs#L1270-L1278

https://github.com/Mooncake-Labs/moonlink/blob/58b6116feffa5fa9d870ff96b0201e26f3948543/src/moonlink/src/storage/mooncake_table.rs#L1073-L1087

https://github.com/Mooncake-Labs/moonlink/blob/58b6116feffa5fa9d870ff96b0201e26f3948543/src/moonlink/src/storage/mooncake_table.rs#L1110-L1122

I don't have strong opinion here, but better to keep the same coding standard.
That said, I'm ok if you prefer to put the task spawn at eventloop, but better to update current impl as well.

src/moonlink/src/storage/wal/tests.rs
lohpaul9: ahh okay, changed to use tokio::fs instead

src/moonlink/src/storage/wal/tests.rs
lohpaul9: changed to use tokio::fs

src/moonlink/src/storage/wal.rs
lohpaul9: changed to handle error for deletion

src/moonlink/src/storage/wal.rs
lohpaul9: removed wrapping function, let's do `take(&mut self.in_mem_wal.buf)`

src/moonlink/src/storage/wal/test_utils.rs
lohpaul9: Will just use `take(&mut self.in_mem_wal.buf)` instead

src/moonlink/src/storage/wal.rs
lohpaul9: rework field visibility

Issue #967, opened 2025-07-22
chaos: make table events serializable


dentiny:
## Summary

... otherwise cannot display it

PR #976, opened 2025-07-22
[storage] Add basic chaos test

dentiny:
## Summary

This PR introduces the very basic chaos test to moonlink eventloop, which is known to lack extensive testing and hard to test due to timing issues.
As of now, it has discovered two bugs with table handler.

A few other things included in the PR:
- Make table events clone-able, so it could be sent to the replay channel;
- Update display formatter, otherwise it's very verbose and spammy to print file index out.

The implementation here is very coarse, quite a few TODO items; I pause it because it doesn't make sense to keep adding new feature when we already detected a few issues.
- Add snapshot read, to check whether read content is correct
- Check iceberg snapshot persistence
- Add streaming events
- Add table maintenance events
- Add chaos test into CI pipeline

## Related Issues

Closes https://github.com/Mooncake-Labs/moonlink/issues/967

## Checklist

- [x] Code builds correctly

### DIFF
--- a/src/moonlink/src/table_handler/chaos_test.rs
+++ b/src/moonlink/src/table_handler/chaos_test.rs
@@ -14,8 +14,11 @@ use crate::ObjectStorageCache;
 use crate::TableEventManager;
 
 use rand::prelude::*;
-use rand::Rng;
+use rand::rngs::StdRng;
+use rand::{Rng, SeedableRng};
 use std::collections::VecDeque;
+use std::panic::{catch_unwind, AssertUnwindSafe};
+use std::time::{SystemTime, UNIX_EPOCH};
 use tempfile::{tempdir, TempDir};
 use tokio::sync::mpsc::{self, Sender};
 use tokio::sync::watch;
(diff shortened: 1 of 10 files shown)


src/moonlink/src/table_handler/chaos_test.rs
```diff
@@ -0,0 +1,239 @@
...
+struct TestEnvironment {
+    iceberg_temp_dir: TempDir,
+    cache_temp_dir: TempDir,
+    table_temp_dir: TempDir,
+    object_storage_cache: ObjectStorageCache,
+    read_state_manager: ReadStateManager,
+    table_event_manager: TableEventManager,
+    table_handler: TableHandler,
+    event_sender: Sender<TableEvent>,
+    event_replay_rx: mpsc::UnboundedReceiver<TableEvent>,
```
nbiscaro, line 142: Why are we using `Unbounded` here? How do we know this is safe?

dentiny: What's your concern?

dentiny: Fyi it's only used in testing env

Issue #934, opened 2025-07-19
iceberg: persist property at snapshot summary


dentiny:
## Summary

Current implementation adds key-value pair as table property, which is not good;
I did so because snapshot summary hasn't exposed an interface for iceberg-rust.
https://github.com/Mooncake-Labs/moonlink/blob/349f3b67f6700f84251c02c37b78001a6c0ec5fd/src/moonlink/src/storage/iceberg/iceberg_table_syncer.rs#L486-L489

PR #999, opened 2025-07-23
[storage] Fix snapshot summary properties

dentiny:
## Summary

Moonlink requires extra key-value pair storage to record its own metadata like flush LSN;
Current implementation is a hacky workaround: I set it in the table metadata, because iceberg-rust doesn't have better interface support.
I upstream a patch to allow append-only action to take properties, so now moonlink metadata is per-snapshot-based.

## Related Issues

Closes https://github.com/Mooncake-Labs/moonlink/issues/934

## Checklist

- [ ] Code builds correctly


src/moonlink/src/storage/iceberg/iceberg_table_syncer.rs
```diff
@@ -476,26 +476,33 @@ impl IcebergTableManager {
...
-            action = action.set(
+            snapshot_properties.insert(
                 MOONCAKE_WAL_METADATA.to_string(),
                 serde_json::to_string(&wal_metadata).unwrap(),
             );
         }
-        txn = action.apply(txn)?;
+
+        let mut txn = Transaction::new(self.iceberg_table.as_ref().unwrap());
+        let action = txn.fast_append();
```
dentiny, line 492: All updates should be placed under one transaction, otherwise `TableRequirement` check fails.

PR #1008, opened 2025-07-24
[storage] Rework recovery on streaming

dentiny:
## Summary

This PR rework recovery on streaming. 
Current implementation doesn't work for streaming write, whose LSN is only assigned at commit/abort.

This PR keeps everything unchanged for non-streaming write, but not discard streaming write events until commit/abort.

## Checklist

- [x] Code builds correctly

### DIFF
--- a/src/moonlink/src/table_handler/tests.rs
+++ b/src/moonlink/src/table_handler/tests.rs
@@ -1543,7 +1543,7 @@ async fn test_full_maintenance_with_sufficient_data_files() {
     assert_eq!(snapshot.indices.file_indices.len(), 1); // one compacted file index
 }
 
-/// Testing scenario: write operations no later than persisted LSN shall be discarded.
+/// Testing scenario: non-streaming write operations no later than persisted LSN shall be discarded.
 #[tokio::test]
 async fn test_discard_duplicate_writes() {
     let temp_dir = tempdir().unwrap();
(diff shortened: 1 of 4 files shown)


nbiscaro:
We should also refactor the `get_final_lsn` logic since this is only relevant for non-streaming tx.

src/moonlink/src/table_handler/tests.rs
```diff
@@ -1626,7 +1626,7 @@ async fn test_discard_duplicate_writes() {
         /*id=*/ 40,
         /*name=*/ "Dog",
         /*age=*/ 40,
-        /*lsn=*/ 25,
+        /*lsn=*/ 0,
```
dentiny, line 1629: Check LSN less than persistence LSN case.

dentiny:
> We should also refactor the get_final_lsn logic since this is only relevant for non-streaming tx.

Thanks for the review! I will followup with another PR today, not that familiar with pg code, might take more time. :(

nbiscaro:
> > We should also refactor the get_final_lsn logic since this is only relevant for non-streaming tx.
> 
> Thanks for the review! I will followup with another PR today, not that familiar with pg code, might take more time. :(

I mean the code added by you here https://github.com/Mooncake-Labs/moonlink/pull/690/files

dentiny:
> I mean the code added by you here

Yes.

dentiny:
> I mean the code added by you here https://github.com/Mooncake-Labs/moonlink/pull/690/files

And what worth refactor is https://github.com/Mooncake-Labs/moonlink/blob/1d0061a6d3c0194d4b6ed10bf7dd2febd895a1d9/src/moonlink_connectors/src/pg_replicate/moonlink_sink.rs#L17-L21, not `get_final_lsn`

Issue #1027, opened 2025-07-24
BUG: force index merge fails assertion


dentiny:
## Summary

```sh
thread 'tokio-runtime-worker' panicked at src/moonlink/src/table_handler.rs:225:29:
assertion `left == right` failed
  left: ForceRegular
 right: Unrequested
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

- default enabled compaction
- force index merge requested

PR #1028, opened 2025-07-24
[storage] Fix multi maintenance requests

dentiny:
## Summary

This issue is:
- It's possible to have multiple table maintenance requested, but none gets started;
- When the second request starts, we check current status is `UNREQUESTED`, which fails the assertion.

So in this PR, I directly acks back when there's already ongoing maintenance.

## Related Issues

Closes https://github.com/Mooncake-Labs/moonlink/issues/1027

## Checklist

- [x] Code builds correctly

### DIFF
--- a/src/moonlink/src/table_handler.rs
+++ b/src/moonlink/src/table_handler.rs
@@ -217,7 +217,7 @@ impl TableHandler {
                         // Branch to trigger a force regular index merge request.
                         TableEvent::ForceRegularIndexMerge => {
                             // TODO(hjiang): If there's already table maintenance ongoing, skip.
-                            if table_handler_state.table_maintenance_process_status.is_maintenance_ongoing() {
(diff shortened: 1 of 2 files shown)


src/moonlink/src/table_handler.rs
```diff
@@ -217,7 +217,7 @@ impl TableHandler {
                         // Branch to trigger a force regular index merge request.
                         TableEvent::ForceRegularIndexMerge => {
                             // TODO(hjiang): If there's already table maintenance ongoing, skip.
-                            if table_handler_state.table_maintenance_process_status.is_maintenance_ongoing() {
+                            if !table_handler_state.can_start_new_maintenance() {
```
dentiny, line 220: Cannot add reliable unit test due to concurrency :(
Checked by chaos test

Issue #1030, opened 2025-07-24
[ISSUE] Make batch IDs globally unique


nbiscaro:
## Summary

In order to support async flush we need to add batches from both streaming and non-streaming transactions to the snapshot. In this case we must ensure that batch ids are unique across all transactions for the lifetime of the table. 

## Context

Why is this issue important? Link any relevant discussions, designs, or documents.

## Tasks

- [ ] 

## Notes

Any extra notes or details that might be useful.

PR #1031, opened 2025-07-24
Unique batch id allocations per table

nbiscaro:
## Summary

In order to support async flush batches from both streaming and non-streaming transactions will need to be added to the snapshot. in order to support this, we must ensure batch ids are globally unique across all transactions. 

Previous behavior would start the `batch_ids` at 0 for each new mem slice. This means streaming transaction batch ids overlap with other streaming transactions and the main table mem slice. 

We do this by splitting the `u64` range in two, assigning streaming transactions the range `[0, 21^63)` and non-streaming `[2^63, 2^64)`. It's important that streaming batch id's are always strictly less than non-streaming batch id's since we use the commit point from the non-streaming mem slice to determine read visibility. As soon as a streaming transaction commits it will become visible since its always less than commit point.  

## Related Issues

Closes https://github.com/Mooncake-Labs/moonlink/issues/1030

## Changes

- 
- 
- 

## Checklist

- [ ] Code builds correctly

### DIFF
--- a/src/moonlink/src/storage/mooncake_table/batch_id_counter.rs
+++ b/src/moonlink/src/storage/mooncake_table/batch_id_counter.rs
@@ -4,8 +4,8 @@ use std::sync::Arc;
 /// Batch ID counter for the two-counter allocation strategy.
 ///
 /// The system uses two separate atomic counters to partition the 64-bit batch ID space:
-/// - **Streaming Counter**: Range 0 -> 2^32-1, used for streaming transactions
-/// - **Non-Streaming Counter**: Range 2^32+, used for regular operations
+/// - **Streaming Counter**: Range 0 -> 2^63-1, used for streaming transactions
+/// - **Non-Streaming Counter**: Range 2^63+, used for regular operations
 ///
 /// We give streaming batches the smaller range so that they are always behind the commit point, which points to the most recently added batch of the non-streaming batches.
 /// This ensures batch IDs are always monotonically increasing and unique across all transactions.
@@ -17,7 +17,7 @@ pub(super) struct BatchIdCounter {
 impl BatchIdCounter {
     pub fn new(is_streaming: bool) -> Self {
         Self {
-            counter: Arc::new(AtomicU64::new(if is_streaming { 0 } else { 1u64 << 32 })),
+            counter: Arc::new(AtomicU64::new(if is_streaming { 0 } else { 1u64 << 63 })),
             is_streaming,
         }
     }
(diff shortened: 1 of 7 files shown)


src/moonlink/src/storage/mooncake_table/batch_id_counter.rs
```diff
@@ -0,0 +1,178 @@
...
+        self.counter.load(Ordering::Relaxed)
+    }
+
+    pub fn next(&self) -> u64 {
+        let current = self.counter.load(Ordering::Relaxed);
+
+        // Check limits before incrementing
+        if self.is_streaming {
+            assert!(
+                current < (1u64 << 63),
```
dentiny, line 35: `ma::assert_lt!`

src/moonlink/src/storage/mooncake_table/batch_id_counter.rs
```diff
@@ -0,0 +1,178 @@
...
+                "Streaming batch ID counter overflow: exceeded 2^63-1"
+            );
+        } else {
+            assert!(
+                current < u64::MAX,
```
dentiny, line 40: `ma::assert_lt!`

src/moonlink/src/storage/mooncake_table/batch_id_counter.rs
```diff
@@ -0,0 +1,178 @@
...
+
+impl BatchIdCounter {
+    pub fn new(is_streaming: bool) -> Self {
+        Self {
+            counter: Arc::new(AtomicU64::new(if is_streaming { 0 } else { 1u64 << 63 })),
+            is_streaming,
+        }
+    }
+
+    pub fn load(&self) -> u64 {
```
dentiny, line 25: just for my own education, do you mind telling me why relaxed memory order is safe?

src/moonlink/src/storage/mooncake_table/batch_id_counter.rs
```diff
@@ -0,0 +1,178 @@
...
+        }
+
+        self.counter.fetch_add(1, Ordering::Relaxed)
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use std::thread;
```
dentiny, line 52: why there's no concurrency test?

src/moonlink/src/storage/mooncake_table/batch_id_counter.rs
nbiscaro: We know it’s safe to use Relaxed here because we’re only using the atomic counter for generating unique IDs via fetch_add (atomic read and store), and we’re not relying on it to synchronize visibility of any other memory. The atomic operation guarantees uniqueness due to atomicity, and Relaxed ordering is sufficient because we don’t need ordering or visibility guarantees.

https://doc.rust-lang.org/std/sync/atomic/enum.Ordering.html

dentiny: My only ask is to add a comment why it's safe, thanks.

PR #1036, opened 2025-07-25
[storage] Attach id to background events

dentiny:
## Summary

Certain bugs only surface after injecting 20K events, making it hard to debug with certain ways to "locate" the start and end for multiple concurrent background tasks.
Later event UUID is also helpful for observability, like logging and tracing.

## Checklist

- [ ] Code builds correctly


dentiny:
This PR is a no-op change which doesn't affect any functionality, the batch id issue is known and tracked separately.

Issue #1033, opened 2025-07-24
BUG: persisted deletion vector not found


dentiny:
## Summary

Error message:
```sh
thread 'tokio-runtime-worker' panicked at src/moonlink/src/storage/mooncake_table/snapshot_persistence.rs:202:76:
called `Option::unwrap()` on a `None` value
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

https://github.com/Mooncake-Labs/moonlink/blob/c0f8647f9b7ad0ed03e378c184e7e8c2618c7071/src/moonlink/src/storage/mooncake_table/snapshot_persistence.rs#L202

dentiny:
Error case:
- have two (data file, puffin) in disk files
- when puffin blob gets updated (due to deletion)
- iceberg snapshot payload is created
- data compaction request created and completed
- the (data file, puffin) gets replace by (compacted data file, None) due to compaction
- when iceberg persistence finishes, the data file already doesn't exist

dentiny:
fix: https://github.com/dentiny/moonlink/pull/43/commits/09f500dd83a2c8d34d14f376ab26b3ed97cd9e78

PR #1049, opened 2025-07-25
[storage] Allow puffin blob unfound at persistence

dentiny:
## Summary

See https://github.com/Mooncake-Labs/moonlink/issues/1033#issuecomment-3116730159 for details.
It's hard to do manual testing, since it requires two concurrent background threads, one for compaction, another for iceberg snapshot, issued and completed at certain timing.
Manual synchronization is also awkward, since if file handle not pinned, files will get deleted after compaction.

So I checked with chaos test.

## Related Issues

Closes https://github.com/Mooncake-Labs/moonlink/issues/1033

## Checklist

- [x] Code builds correctly

### DIFF
--- a/src/moonlink/src/storage/mooncake_table/snapshot_persistence.rs
+++ b/src/moonlink/src/storage/mooncake_table/snapshot_persistence.rs
@@ -200,18 +200,29 @@ impl SnapshotTableState {
         // Aggregate the evicted files to delete.
         let mut evicted_files_to_delete = vec![];
 
-        for (file_id, puffin_blob_ref) in puffin_blob_ref.into_iter() {
-            let entry = self.current_snapshot.disk_files.get_mut(&file_id).unwrap();
-            // Unreference and delete old cache handle if any.
-            let old_puffin_blob = entry.puffin_deletion_blob.take();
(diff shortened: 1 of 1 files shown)


src/moonlink/src/storage/mooncake_table/snapshot_persistence.rs
```diff
@@ -200,18 +200,29 @@ impl SnapshotTableState {
...
+                    let cur_evicted_files = old_puffin_blob
+                        .puffin_file_cache_handle
+                        .unreference_and_delete()
+                        .await;
+                    evicted_files_to_delete.extend(cur_evicted_files);
+                }
+                entry.puffin_deletion_blob = Some(puffin_blob_ref);
+            }
+            // The referenced data file has been removed, for example, by completed data compaction; directly discard the puffin blob.
+            // The committed deletion record included by the puffin deletion blob is still contained current snapshot's committed deletion records.
```
dentiny, line 218: If corresponding data files take part in compaction, committed deletion records will get remapped, and won't exist in `committed_deletion_logs_to_persist` (see https://github.com/Mooncake-Labs/moonlink/pull/1045)

Issue #1042, opened 2025-07-25
BUG: remap compacted deletion record


dentiny:
## Summary

Remapping committed deletion records should happen before prune by persistence, because compaction doesn't change flush LSN.
https://github.com/Mooncake-Labs/moonlink/blob/37c54a605781aee13e7c0ddb57d00b58504883fc/src/moonlink/src/storage/mooncake_table/snapshot.rs#L454-L455

dentiny:
fix: https://github.com/dentiny/moonlink/pull/43/commits/ee1fa488f9f5121ce8cda13e10ea1f12d03b982b

PR #1051, opened 2025-07-25
[storage] Fix committed deletion log remap

dentiny:
## Summary

This PR fixes bug: if we first prune committed deletion records, then do remap for compaction, committed records might have been pruned at remap, thus loss positional deletes at read.

## Related Issues

Closes https://github.com/Mooncake-Labs/moonlink/issues/1042

## Checklist

- [ ] Code builds correctly

### DIFF
--- a/src/moonlink/src/storage/mooncake_table/snapshot.rs
+++ b/src/moonlink/src/storage/mooncake_table/snapshot.rs
@@ -464,7 +464,15 @@ impl SnapshotTableState {
             .await;
         evicted_data_files_to_delete.extend(compaction_evicted_files);
 
+        // Remap uncommitted deletion logs.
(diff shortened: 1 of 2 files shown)


src/moonlink/src/storage/mooncake_table/snapshot.rs
```diff
@@ -464,7 +464,15 @@ impl SnapshotTableState {
             .await;
         evicted_data_files_to_delete.extend(compaction_evicted_files);
 
+        // Remap uncommitted deletion logs.
+        self.remap_uncommitted_deletion_logs_after_compaction(&mut task);
```
dentiny, line 468: one TODO is to combine two remap operations, after https://github.com/Mooncake-Labs/moonlink/pull/1045

PR #1086, opened 2025-07-28
wal redesign for streaming txn cases

### LINKED ISSUE #1055: wal / redesign for streaming xact
## Summary

Briefly describe the task, feature, or problem.

## Context

Why is this issue important? Link any relevant discussions, designs, or documents.

## Tasks

- [ ] 

## Notes

Any extra notes or details that might be useful.

lohpaul9:
Make changes to WAL design to handle streaming transactions

Closes #1055

### DIFF
--- a/src/moonlink/src/storage/wal.rs
+++ b/src/moonlink/src/storage/wal.rs
@@ -47,7 +47,7 @@ pub struct WalFileInfo {
 }
 
 impl WalEvent {
-    pub fn from_table_event(table_event: &TableEvent) -> Self {
+    pub fn new(table_event: &TableEvent) -> Self {
         match table_event {
             TableEvent::Append {
                 row,
@@ -267,7 +267,7 @@ impl WalManager {
 
     fn push(&mut self, table_event: &TableEvent) {
         // add to in_mem_buf
-        let wal_event = WalEvent::from_table_event(table_event);
+        let wal_event = WalEvent::new(table_event);
         self.in_mem_buf.push(wal_event);
 
         // Update highest_lsn if this event has a higher LSN
@@ -428,6 +428,8 @@ impl WalManager {
         });
     }
 
+    /// This cleans up any files that no longer need to be tracked, and also any xacts that no longer need to
+    /// be tracked as a result.
     fn handle_complete_truncate(&mut self, persist_and_truncate_result: &PersistAndTruncateResult) {
         if let Some(highest_deleted_file) = &persist_and_truncate_result.highest_deleted_file {
             self.live_wal_files_tracker.retain(|wal_file_info| {
@@ -441,8 +443,10 @@ impl WalManager {
         }
     }
 
-    /// For now, we handle the persist and truncate results together.
+    /// Should be called after a persist and truncate operation has completed. Updates
+    /// tracked files and transactions.
     pub fn handle_completed_persist_and_truncate(
+        // For now, we handle the persist and truncate results together.
         &mut self,
         persist_and_truncate_result: &PersistAndTruncateResult,
     ) {
--- a/src/moonlink/src/storage/wal.rs
+++ b/src/moonlink/src/storage/wal.rs
@@ -9,27 +9,28 @@ use futures::stream::{self, Stream};
 use futures::{future, StreamExt};
 use more_asserts::assert_ge;
 use serde::{Deserialize, Serialize};
+use std::collections::HashMap;
 use std::pin::Pin;
 use std::sync::Arc;
 
-#[derive(Debug, Serialize, Deserialize, PartialEq)]
-enum WalEventEnum {
+#[derive(Debug, Serialize, Deserialize, PartialEq, Clone)]
+pub enum WalEvent {
     Append {
         row: MoonlinkRow,
         xact_id: Option<u32>,
+        lsn: u64,
         is_copied: bool,
     },
     Delete {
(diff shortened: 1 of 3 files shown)


src/moonlink/src/storage/wal.rs
```diff
@@ -9,132 +9,142 @@ use futures::stream::{self, Stream};
...
     },
-    StreamFlush {
-        xact_id: u32,
-    },
 }
 
 pub struct PersistAndTruncateResult {
     file_persisted: Option<WalFileInfo>,
     highest_deleted_file: Option<WalFileInfo>,
+    associated_iceberg_snapshot_lsn: Option<u64>,
```
nbiscaro: nit: 
```suggestion
    iceberg_snapshot_lsn: Option<u64>,
```

src/moonlink/src/storage/wal.rs
```diff
@@ -9,132 +9,142 @@ use futures::stream::{self, Stream};
...
-    /// The in_mem_wal could have insertions done by incoming CDC events
-    pub buf: Vec<WalEvent>,
-    /// Tracks an LSN in case of a stream flush (or similar events) which has no accompanying LSN.
-    /// A new instance will take the highest_lsn from the previous instance before it is populated.
-    highest_lsn: u64,
+
+#[derive(Debug, Clone, PartialEq)]
+enum TransactionState {
+    // xact_ids are u32, so we can use u64 so we can have a
+    // larger number to represent the main transaction
```
nbiscaro: You can still use `u32` here and use `xact_id: 0` as your sentinel value: https://github.com/postgres/postgres/blob/d5b9b2d40262f57f58322ad49f8928fd4a492adb/src/include/access/transam.h#L31

src/moonlink/src/storage/wal.rs
```diff
@@ -9,132 +9,142 @@ use futures::stream::{self, Stream};
...
 }
-pub struct InMemWal {
-    /// The in_mem_wal could have insertions done by incoming CDC events
-    pub buf: Vec<WalEvent>,
-    /// Tracks an LSN in case of a stream flush (or similar events) which has no accompanying LSN.
-    /// A new instance will take the highest_lsn from the previous instance before it is populated.
-    highest_lsn: u64,
+
+#[derive(Debug, Clone, PartialEq)]
+enum TransactionState {
```
nbiscaro: `TransactionState` already exists. We should either unify these or rename, but don't overload: https://github.com/Mooncake-Labs/moonlink/blob/7310be1e60048c77554caff0a0f9d965ee7033d7/src/moonlink_connectors/src/pg_replicate/moonlink_sink.rs#L18

src/moonlink/src/storage/wal.rs
```diff
@@ -145,11 +155,22 @@ impl InMemWal {
...
+    highest_seen_lsn: u64,
+    /// The wal file numbers that are still live. Tracked in ascending order of file number.
     live_wal_files_tracker: Vec<WalFileInfo>,
-    /// Tracks the file number to be assigned to the next persisted wal file
+    /// Tracks the file number to be assigned to the next flushed file.
+    /// All events currently in the in_mem_buf will be flushed to a file with this file number.
     curr_file_number: u64,
+    /// Tracks any transactions that may not have been flushed to an iceberg snapshot yet,
+    /// and therefore need to live in the WAL.
+    active_transactions: HashMap<u32, TransactionState>,
```
nbiscaro: Can we unify these and just use the special sentinel value of `xact_id: 0` to identify the main transaction? (See above comment).

src/moonlink/src/storage/wal.rs
```diff
@@ -158,82 +179,211 @@ impl WalManager {
...
+    ///
+    /// A WAL file can only be dropped if all its events are captured in the iceberg snapshot,
+    /// as per the criteria above.
+    ///
     /// List of files returned is sorted in ascending order of file number.
     /// Should be called in preparation to asynchronously delete the files.
     pub fn get_files_to_truncate(&self, truncate_from_lsn: u64) -> Vec<WalFileInfo> {
-        let last_truncate_idx = self
-            .live_wal_files_tracker
+        let xacts_still_incomplete_after_truncate = self
```
nbiscaro, line 303: Same as above, do we need to process main vs streaming independently?

nbiscaro (commented):
Some smaller comments. 

One thing I don't see is updated logic for `send_status_update` and it's unclear which LSN will be used for this here. Is this to be done in a follow up PR?

src/moonlink/src/storage/wal.rs
lohpaul9: Thanks. Realized that actually this comment is outdated and I track main transaction separately, so i've since removed this comment

src/moonlink/src/storage/wal.rs
lohpaul9: Yes we need to process this separately. Over a span of WAL files we can be reasonably sure that only one of each streaming xact could exist at a time, but there could have been multiple main xact commits. Therefore, we need to keep the history of all these main xact commits in case the next iceberg snapshot happens to fall in between one of those many intervals.

lohpaul9:
> Some smaller comments.
> 
> One thing I don't see is updated logic for `send_status_update` and it's unclear which LSN will be used for this here. Is this to be done in a follow up PR?

send_status_update will be handled in a future PR #1090

Issue #1032, opened 2025-07-24
[ISSUE] Add both streaming and non-streaming batches to snapshot task


nbiscaro:
## Summary

Now that we have globally unique ids, we can add both streaming and non-streaming batches to the snapshot task. This doesn't have much value until flushes are done asynchronously, but change can be made separately to keep the async flush PR small. 

## Context

Why is this issue important? Link any relevant discussions, designs, or documents.

## Tasks

- [ ] 

## Notes

Any extra notes or details that might be useful.

PR #1100, opened 2025-07-29
Add stream state to snapshot task

nbiscaro:
## Summary

Removes the assumption that flush happens in the foreground. Adds state to `TransactionStreamState` or `next_snapshot_task` to ensure transaction visibility during flush. Note that flush is still done in the foreground here, this will be changed in a future PR.

In general, each time we flush we need to add the current batches and indices to the stream state. When a flush finishes, we remove this state. On commit we move any remaining batches and indices from the `TransactionStreamState` into the snapshot task to ensure it is visible. When the flush finishes, we add its disk slice to the snapshot task. 

## Related Issues

Closes https://github.com/Mooncake-Labs/moonlink/issues/1032

## Changes

- 
- 
- 

## Checklist

- [ ] Code builds correctly

### DIFF
--- a/src/moonlink/src/storage/mooncake_table/transaction_stream.rs
+++ b/src/moonlink/src/storage/mooncake_table/transaction_stream.rs
@@ -229,6 +229,8 @@ impl MooncakeTable {
         // Record abortion in snapshot task so we can remove any uncomitted deletions
         let stream_state = self.transaction_stream_states.get_mut(&xact_id).unwrap();
 
+        stream_state.status = TransactionStreamStatus::Aborted;
+
         // If there are no pending flushes, we can remove the stream state immediately
         // Otherwise, let `apply_stream_flush_result` handle the abortion
         if stream_state.pending_flush_count == 0 {
--- a/src/moonlink/src/storage/mooncake_table/transaction_stream.rs
+++ b/src/moonlink/src/storage/mooncake_table/transaction_stream.rs
@@ -18,8 +18,26 @@ pub(super) struct TransactionStreamState {
     local_deletions: Vec<ProcessedDeletionRecord>,
     pending_deletions_in_main_mem_slice: Vec<RawDeletionRecord>,
     index_bloom_filter: BloomFilter,
-    flushed_file_index: MooncakeIndex,
+    /// Both in memory and on disk indices for this transaction.
+    stream_indices: MooncakeIndex,
     flushed_files: hashbrown::HashMap<MooncakeDataFileRef, DiskFileEntry>,
+    new_record_batches: Vec<(u64, Arc<RecordBatch>)>,
+    status: TransactionStreamStatus,
+    /// LSN of the commit for this transaction.
+    /// Required during `apply_stream_flush_result` to update the writer lsn of the disk slice.
+    commit_lsn: Option<u64>,
+    /// Number of pending flushes for this transaction.
+    /// Only safe to remove transaction stream state when there are no pending flushes.
+    pending_flush_count: u32,
+}
+
+/// Determines the state of a transaction stream.
+/// Transaction can be safely removed when it is no longer `Pending` and has no pending flushes.
+#[derive(PartialEq)]
+pub enum TransactionStreamStatus {
+    Pending,
+    Committed,
+    Aborted,
 }
 
 pub enum TransactionStreamOutput {
@@ -74,8 +92,12 @@ impl TransactionStreamState {
             local_deletions: Vec::new(),
             pending_deletions_in_main_mem_slice: Vec::new(),
             index_bloom_filter: BloomFilter::with_num_bits(1 << 24).expected_items(1_000_000),
-            flushed_file_index: MooncakeIndex::new(),
+            stream_indices: MooncakeIndex::new(),
             flushed_files: hashbrown::HashMap::new(),
+            new_record_batches: Vec::new(),
+            status: TransactionStreamStatus::Pending,
+            commit_lsn: None,
+            pending_flush_count: 0,
         }
     }
 }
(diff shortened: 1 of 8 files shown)


src/moonlink/src/storage/mooncake_table/transaction_stream.rs
```diff
@@ -205,94 +227,233 @@ impl MooncakeTable {
...
-            let index = disk_slice.take_index();
-            if let Some(index) = index {
-                stream_state.flushed_file_index.insert_file_index(index);
+            return;
+        }
+
+        // Transaction aborted while stream was flushing. Remove stream state and do nothing.
+        // Drop the stream state since the transaction is over.
+        if stream_state.status == TransactionStreamStatus::Aborted {
+            if stream_state.pending_flush_count == 0 {
```
zhousun, line 379: We need to delete the file in this case?

src/moonlink/src/storage/mooncake_table/transaction_stream.rs
```diff
@@ -154,7 +172,7 @@ impl MooncakeTable {
                 return;
             }
             // Delete from stream flushed files
-            let matches = stream_state.flushed_file_index.find_record(&record).await;
+            let matches = stream_state.stream_indices.find_record(&record).await;
             if !matches.is_empty() {
                 for loc in matches {
                     let RecordLocation::DiskFile(file_id, row_id) = loc else {
```
zhousun: match == Memory is no longer 'unexpected'

src/moonlink/src/storage/mooncake_table/transaction_stream.rs
```diff
@@ -205,94 +223,234 @@ impl MooncakeTable {
...
+        // Drop the stream state since the transaction is over.
+        if stream_state.status == TransactionStreamStatus::Aborted {
+            if stream_state.pending_flush_count == 0 {
+                self.transaction_stream_states.remove(&xact_id);
             }
-            return Ok(());
+            return;
+        }
+
+        // Append state so we can find deletes during tx
```
zhousun, line 385: Where do we 'remap_deletion_if_needed' ?

src/moonlink/src/storage/mooncake_table/transaction_stream.rs
nbiscaro: Yes, although we don't do this in the current implementation either. If we want to implement a cleanup process for aborted stream files we should do this in a separate PR.

src/moonlink/src/storage/mooncake_table/transaction_stream.rs
nbiscaro: Yes, good call. 

Added support for deleting from in memory stream state. Also added unit test for this case.

src/moonlink/src/storage/mooncake_table/transaction_stream.rs
nbiscaro: Part of above implementation.

src/moonlink/src/storage/mooncake_table/transaction_stream.rs
zhousun: kk

zhousun:
lgtm. 
I didn't read super carefully especially the tests, since I assume turning on 'async flush' might change a lot of them

src/moonlink/src/storage/mooncake_table/transaction_stream.rs
```diff
@@ -153,34 +172,76 @@ impl MooncakeTable {
...
+                            {
+                                continue;
+                            }
+                            // Push the deletion record to stream state
+                            stream_state.local_deletions.push(ProcessedDeletionRecord {
+                                pos: loc,
+                                lsn: record.lsn,
+                            });
+                            // Mark the row as deleted in the batch
+                            stream_state
```
dentiny, line 209: Where did you initialize the [`deletions`] for the batch record?

src/moonlink/src/storage/mooncake_table/transaction_stream.rs
```diff
@@ -153,34 +172,76 @@ impl MooncakeTable {
...
-                        disk_file_entry.batch_deletion_vector.delete_row(row_id);
-                        return;
+                    match loc {
+                        RecordLocation::MemoryBatch(batch_id, row_id) => {
+                            let batch = stream_state
+                                .new_record_batches
+                                .get(&batch_id)
+                                .expect("Attempting to delete batch that doesn't exist");
+
+                            if batch.deletions.is_deleted(row_id) {
```
dentiny, line 186: curious in which case would it happen? does it mean a row is deleted for multiple times?

src/moonlink/src/storage/mooncake_table/transaction_stream.rs
```diff
@@ -205,94 +266,247 @@ impl MooncakeTable {
...
+        // We filter here since in the stream case we delete from the current mem slice directly instead of adding to `new_deletions`
+        let (_, mut batches, index) = stream_state.mem_slice.drain()?;
+        for batch in batches.iter_mut() {
+            let filtered_batch = batch.batch.get_filtered_batch()?;
+            if let Some(filtered_batch) = filtered_batch {
+                stream_state.new_record_batches.insert(
+                    batch.id,
+                    InMemoryBatch {
+                        data: Some(Arc::new(filtered_batch)),
+                        deletions: BatchDeletionVector::default(),
```
dentiny, line 311: Why not setting num rows? wouldn't check `is_deleted` also be false?

dentiny: You even didn't check `delete_row` return...

src/moonlink/src/storage/mooncake_table/transaction_stream.rs
```diff
@@ -205,94 +266,247 @@ impl MooncakeTable {
...
+        // Add state from current mem slice to stream first
+        let (_, mut batches, index) = stream_state.mem_slice.drain()?;
+        for batch in batches.iter_mut() {
+            let filtered_batch = batch.batch.get_filtered_batch()?;
+            if let Some(filtered_batch) = filtered_batch {
+                stream_state.new_record_batches.insert(
+                    batch.id,
+                    InMemoryBatch {
+                        data: Some(Arc::new(filtered_batch)),
+                        deletions: BatchDeletionVector::default(),
```
dentiny, line 454: Same question here, we need to sync

PR #1122, opened 2025-07-30
Wal table handler integration

### LINKED ISSUE #972: wal / integrate WAL to flush in table handler
## Summary
Success state: WAL should be running in the event loop and flushing should work correctly.

lohpaul9:
Adds WAL logic to persist and truncate in WAL handler, based on latest iceberg snapshot LSN.
Will add in tracking of latest WAL file in each iceberg snapshot for recovery in later PR. 

Closes #972


src/moonlink/src/table_handler.rs
```diff
@@ -56,13 +56,20 @@ impl TableHandler {
...
+        let event_sender_for_periodical_wal = event_sender.clone();
         let periodic_event_handle = tokio::spawn(async move {
             let mut periodic_snapshot_interval = tokio::time::interval(Duration::from_millis(500));
             let mut periodic_force_snapshot_interval =
                 tokio::time::interval(Duration::from_secs(300));
+            let mut periodic_wal_interval = tokio::time::interval(Duration::from_millis(500));
 
             loop {
                 tokio::select! {
+                    _ = periodic_wal_interval.tick() => {
```
dentiny, line 70: do you mind putting it to the bottom? the comment at L73 is supposed to cover all timer cases

src/moonlink/src/storage/mooncake_table.rs
```diff
@@ -1328,6 +1350,48 @@ impl MooncakeTable {
             .await
             .unwrap();
     }
+
+    pub(crate) fn persist_and_update_wal(&mut self) -> bool {
```
dentiny: What's the meaning of return value?

dentiny: and must use annotation

src/moonlink/src/storage/mooncake_table.rs
```diff
@@ -1328,6 +1350,48 @@ impl MooncakeTable {
...
+        let latest_iceberg_snapshot_lsn = self.get_iceberg_snapshot_lsn();
+        let files_to_truncate =
+            if let Some(latest_iceberg_snapshot_lsn) = latest_iceberg_snapshot_lsn {
+                self.wal_manager
+                    .get_files_to_truncate(latest_iceberg_snapshot_lsn)
+            } else {
+                vec![]
+            };
+        let file_to_persist = self.wal_manager.take_for_next_file();
```
dentiny: I would appreciate if you could rename `take_for_next_file`, it's hard for me to grasp.

src/moonlink/src/storage/mooncake_table.rs
```diff
@@ -1328,6 +1350,48 @@ impl MooncakeTable {
...
+                )
+                .await;
+            });
+            true
+        } else {
+            false
+        }
+    }
+
+    pub(crate) fn handle_completed_persist_and_truncate(
```
dentiny:
```suggestion
    pub(crate) fn handle_completed_persistence_and_truncate(
```

src/moonlink/src/storage/wal.rs
```diff
@@ -402,6 +435,67 @@ impl WalManager {
...
+        // Create result and notify
+        let persist_and_truncate_result = result.map(|_| {
+            PersistAndTruncateResult::new(
+                file_to_persist.map(|(_, wal_file_info)| wal_file_info),
+                files_to_delete.last().cloned(),
+                iceberg_snapshot_lsn,
+            )
+        });
+
+        table_notify
```
dentiny, line 504: how do you synchronize eventloop exit and WAL persist/truncate?
is it possible that eventloop has dropped, but WAL background task still ongoing, so you suffer `send` failure?

src/moonlink/src/storage/wal.rs
```diff
@@ -445,9 +539,14 @@ impl WalManager {
         // For now, we handle the persist and truncate results together.
         &mut self,
         persist_and_truncate_result: &PersistAndTruncateResult,
-    ) {
+    ) -> Option<u64> {
```
dentiny, line 558: doc on return

src/moonlink/src/storage/wal.rs
lohpaul9: Not quite sure if this is being handled any differently from the other mooncake table methods that run in the background, I think the event loop should wait for all event senders to drop before exiting?

https://github.com/Mooncake-Labs/moonlink/pull/1122/files/b1ed3407ae8c03d8a92c4c075aebc4b083d1f556#diff-0e6fbb4cc7dfc0978eaf954415d49e73ade29d3710905c9c22810a1af9d6acf4L491

There is a bug regarding wal_flush_lsn receivers being dropped in chaos test which results in a send error of wal_flush_lsn, i'm working on fixing that

src/moonlink/src/storage/mooncake_table.rs
lohpaul9: 'extract_next_persistent_file'

src/moonlink/src/storage/wal.rs
lohpaul9: wal_flush_lsn receiver bug should be fixed with latest commit

dentiny: > I think the event loop should wait for all event senders to drop before exiting

Before your PR drop doesn't know anything about WAL events, so you should add relevant logic?

src/moonlink/src/storage/mooncake_table.rs
```diff
@@ -1408,6 +1428,52 @@ impl MooncakeTable {
             .await
             .unwrap();
     }
+
+    /// Spawns a background task to persist and truncate WAL if needed.
+    /// Returns true if there was a task spawned, false otherwise.
+    pub(crate) fn persist_and_truncate_wal(&mut self) -> bool {
```
dentiny, line 1295: I strongly suggest annotate with `#[must_use]`, we suffer bugs in the past: https://github.com/Mooncake-Labs/moonlink/pull/701

lohpaul9: good point, thanks

src/moonlink/src/storage/mooncake_table.rs
```diff
@@ -957,6 +967,16 @@ impl MooncakeTable {
...
+    #[cfg(test)]
+    pub fn get_table_id(&self) -> u32 {
+        self.metadata.table_id
+    }
+
+    #[cfg(test)]
+    pub fn get_wal_manager(&self) -> &WalManager {
+        &self.wal_manager
+    }
+
```
nbiscaro: Can you impl these either at the bottom of the source file or in the test file altogether. Let's avoid mixing test code with production when possible.

dentiny:
@lohpaul9 You need to update https://github.com/Mooncake-Labs/moonlink/blob/dbc9c2074d1101994411e60e1457ac41d472d753/src/moonlink/src/table_handler/table_handler_state.rs#L318-L329
otherwise it's possible to drop table when WAL background flush ongoing.

dentiny:
and actually drop table when background task completed, example: https://github.com/Mooncake-Labs/moonlink/blob/dbc9c2074d1101994411e60e1457ac41d472d753/src/moonlink/src/table_handler.rs#L459-L463

src/moonlink/src/table_handler/test_utils.rs
```diff
@@ -312,6 +361,81 @@ impl TestEnvironment {
...
+    pub async fn check_wal_events(
+        &self,
+        should_contain_table_events: &[TableEvent],
+        should_not_contain_table_events: &[TableEvent],
+    ) {
+        let wal_events = self.get_wal_events().await;
+
+        assert_wal_events_contains(&wal_events, should_contain_table_events);
+        assert_wal_events_does_not_contain(&wal_events, should_not_contain_table_events);
+    }
```
nbiscaro, line 439: Why cant we assert `expected == actual` directly instead of using should contain and should not contain?

nbiscaro (commented):
Some smaller suggestions

src/moonlink/src/storage/wal.rs
nbiscaro: +1 here, I don't see how the case where we persist WAL but the event loop exits before the LSN propagates is handled. What happens in this case?

src/moonlink/src/table_handler/test_utils.rs
lohpaul9: in table handler tests, its difficult to have full control over the sequencing order between iceberg snapshot and wal persist (because they are sent on interval timers), so using some events as 'litmus test' on whether or not they were discarded or contained in the WAL is easier to reason with

src/moonlink/src/storage/mooncake_table.rs
```diff
@@ -754,6 +758,11 @@ impl MooncakeTable {
         let non_streaming_batch_id_counter = Arc::new(BatchIdCounter::new(false));
         let streaming_batch_id_counter = Arc::new(BatchIdCounter::new(true));
 
+        // TODO(Paul): Support object storage for WAL
+        let wal_manager_file_config =
+            WalManager::default_wal_storage_config(table_metadata.table_id, &table_metadata.path);
```
dentiny: I'm afraid it breaks recovery.

table path here is meant to store write-through cache, which recreates on recovery / initialization: https://github.com/Mooncake-Labs/moonlink/blob/0ea67cea1e8c1407487e2e5e1fba397d9be5d49d/src/moonlink_connectors/src/pg_replicate/table_init.rs#L64-L65

src/moonlink/src/event_sync.rs
```diff
@@ -29,6 +31,8 @@ pub struct EventSyncSender {
     pub force_snapshot_completion_tx: watch::Sender<Option<Result<u64>>>,
     /// Notifies when force table maintenance operation completes.
     pub table_maintenance_completion_tx: broadcast::Sender<Result<()>>,
+    /// Notifies when wal flush lsn advances. Will eventually replace flush_lsn_rx.
```
dentiny: Let's leave a TODO, or create an issue.

src/moonlink/src/storage/wal.rs
```diff
@@ -404,6 +437,67 @@ impl WalManager {
...
+            let file_system_accessor_persist = file_system_accessor.clone();
+
+            // Delete old WAL files
+            if !files_to_delete.is_empty() {
+                WalManager::delete_files(file_system_accessor, &files_to_delete).await?;
+            }
+
+            // Persist new WAL file
+            if let Some((wal_events, wal_file_info)) = &file_to_persist {
+                WalManager::persist(file_system_accessor_persist, wal_events, wal_file_info)
```
dentiny, line 487: the "delete all files" implementation should be placed into filesystem layer, instead of application layer

src/moonlink/src/table_handler/chaos_test.rs
```diff
@@ -499,6 +499,8 @@ struct TestEnvironment {
     replication_lsn_tx: watch::Sender<u64>,
     mooncake_table_metadata: Arc<TableMetadata>,
     iceberg_table_config: IcebergTableConfig,
+    // Currently unused but have to be kept so that the receiver is not dropped
+    _wal_flush_lsn_rx: watch::Receiver<u64>,
```
dentiny: No need for `_`, the whole struct has been marked as allow dead code.

dentiny: comment also not needed

dentiny: nit: place all `rx` together? (to L500)
same for L573, let's group all rx together

src/moonlink/src/storage/mooncake_table.rs
```diff
@@ -957,6 +967,16 @@ impl MooncakeTable {
         self.last_wal_persisted_metadata.clone()
     }
 
+    #[cfg(test)]
+    pub fn get_table_id(&self) -> u32 {
```
dentiny: make it a constant so we don't need the testing accessor?
https://github.com/Mooncake-Labs/moonlink/blob/dbc9c2074d1101994411e60e1457ac41d472d753/src/moonlink/src/table_handler/test_utils.rs#L154

src/moonlink/src/storage/mooncake_table.rs
```diff
@@ -957,6 +967,16 @@ impl MooncakeTable {
...
+        self.metadata.table_id
+    }
+
+    #[cfg(test)]
+    pub fn get_wal_manager(&self) -> &WalManager {
```
dentiny: is it used anywhere?

src/moonlink/src/storage/wal.rs
lohpaul9: will add, thanks for the catch

src/moonlink/src/storage/wal.rs
```diff
@@ -196,6 +238,11 @@ impl WalManager {
         format!("wal_{file_number}.json")
     }
 
+    /// Get the file system accessor for testing purposes
+    pub fn get_file_system_accessor(&self) -> Arc<dyn BaseFileSystemAccess> {
```
dentiny, line 251: If it's only for testing purpose, I personally prefer provide another factory function which takes an accessor.
The benefit is less exposed interface:
- It's easy to call `get_file_system_accessor` in production when it shouldn't (you didn't even mark it as test-only?)
- It's impossible to access and expose if we use factory function

src/moonlink/src/storage/wal.rs
```diff
@@ -442,14 +536,29 @@ impl WalManager {
...
+        persist_and_truncate_result
+            .file_persisted
+            .as_ref()
+            .map(|wal_file_info| wal_file_info.highest_lsn)
+    }
+
+    // ------------------------------
+    // Drop WAL files
+    // ------------------------------
+    pub async fn drop_wal(&mut self) -> Result<()> {
```
dentiny, line 572: if you put everything under a certain directory (current impl is write-through cache directory), why not delete the whole directory? it could also clean up possible orphaned files

src/moonlink/src/storage/mooncake_table.rs
lohpaul9: Good catch, will look into this

src/moonlink/src/storage/wal.rs
lohpaul9: will leave TODO and issue

src/moonlink/src/storage/wal.rs
lohpaul9: not just for testing purposes. we need to pass this to a background async function wal_persist_truncate_async

src/moonlink/src/storage/mooncake_table.rs
lohpaul9: removed

src/moonlink/src/storage/wal.rs
dentiny: I read it from your comment
```
/// Get the file system accessor for testing purposes
```

src/moonlink/src/storage/mooncake_table.rs
lohpaul9: will have to make more changes throughout the testing codebase to remove the testing accessor because it's tableid is declared as different values in different tests - might revisit this in a later change.

https://github.com/Mooncake-Labs/moonlink/blob/ab3b5aea1c12a75a3c126ce5b6314c16a27d7048/src/moonlink/src/table_handler/failure_tests.rs#L27

src/moonlink/src/storage/wal.rs
lohpaul9: cleaned up

src/moonlink/src/storage/wal.rs
lohpaul9: https://github.com/Mooncake-Labs/moonlink/issues/1145

src/moonlink/src/storage/wal.rs
```diff
@@ -8,11 +8,35 @@ use crate::table_notify::TableEvent;
...
+pub struct WalConfig {
+    accessor_config: AccessorConfig,
+}
+
+impl WalConfig {
+    pub fn default_wal_config_local(table_id: u32, base_path: &Path) -> WalConfig {
+        let wal_storage_config = StorageConfig::FileSystem {
+            root_directory: base_path
+                .join(DEFAULT_WAL_FOLDER)
+                .join(table_id.to_string())
```
dentiny: need to add database id as well

src/moonlink_connectors/src/pg_replicate/table_init.rs
```diff
@@ -70,6 +71,7 @@ pub async fn build_table_components(
         }),
     );
 
+    let wal_config = WalConfig::default_wal_config_local(table_id, &PathBuf::from(base_path));
```
dentiny: for the directory you use, you need to make sure it exists

dentiny: quite a few rust filesystem APIs fails with "directory or file not found" is parent folder doesn't exist

src/moonlink/src/storage/wal.rs
dentiny: example for iceberg: https://github.com/Mooncake-Labs/moonlink/blob/32b76bfd4a6e0b64e93af756c959e1b5bc36d694/src/moonlink_connectors/src/pg_replicate/table_init.rs#L67-L71

src/moonlink/benches/bench_write.rs
```diff
@@ -64,6 +64,7 @@ fn bench_write(c: &mut Criterion) {
                 };
                 let table_config =
                     MooncakeTableConfig::new(temp_dir.path().to_str().unwrap().to_string());
+                let wal_config = WalConfig::default_wal_config_local(1, temp_dir.path());
```
dentiny: use a constant to represent TABLE_ID

src/moonlink/src/storage/wal.rs
```diff
@@ -404,6 +443,67 @@ impl WalManager {
         Ok(())
     }
 
+    /// Persist a series of wal events to the file system.
+    /// Should be called asynchronously using the results of take_for_next_file.
```
dentiny: comment needs to be updated

src/moonlink_connectors/src/pg_replicate/table_init.rs
lohpaul9: https://github.com/Mooncake-Labs/moonlink/issues/1159

Will cover in this issue

dentiny: I think you should handle it in this PR, otherwise pg_mooncake could potentially crash

dentiny:
@lohpaul9 Forgot to mention, please consider schema evolution in for WAL entries (no need in this PR).

Issue #1121, opened 2025-07-30
wal / link iceberg snapshot with WAL file ID


lohpaul9:

lohpaul9:
closed by #1183

PR #1183, opened 2025-08-01
wal iceberg integration

lohpaul9:
integrates storing the oldest relevant WAL number into iceberg, and flushing any files that exist before that

### DIFF
--- a/src/moonlink/src/storage/mooncake_table.rs
+++ b/src/moonlink/src/storage/mooncake_table.rs
@@ -18,6 +18,7 @@ mod table_snapshot;
 pub mod table_status;
 pub mod table_status_reader;
 mod transaction_stream;
+use more_asserts as ma;
 
 use super::iceberg::puffin_utils::PuffinBlobRef;
 use super::index::{FileIndex, MemIndex, MooncakeIndex};
@@ -47,8 +48,10 @@ pub(crate) use crate::storage::mooncake_table::table_snapshot::{
 };
 use crate::storage::mooncake_table_config::MooncakeTableConfig;
 use crate::storage::storage_utils::{FileId, TableId};
-use crate::storage::wal::wal_persistence_metadata::WalPersistenceMetadata;
-use crate::storage::wal::{WalConfig, WalManager, WalPersistAndTruncateResult};
+use crate::storage::wal::wal_persistence_metadata::IcebergWalMetadata;
+use crate::storage::wal::{
+    IcebergSnapshotWalInfo, WalConfig, WalManager, WalPersistenceUpdateResult,
+};
 use crate::table_notify::{EvictedFiles, TableEvent};
 use crate::NonEvictableHandle;
 use arrow::record_batch::RecordBatch;
@@ -65,8 +68,8 @@ use table_snapshot::{IcebergSnapshotImportResult, IcebergSnapshotIndexMergeResul
 use tokio::sync::mpsc::Receiver;
 use tokio::sync::mpsc::Sender;
 use tokio::sync::{watch, RwLock};
-use tracing::info_span;
 use tracing::Instrument;
+use tracing::{error, info_span};
 use transaction_stream::{TransactionStreamOutput, TransactionStreamState};
 
 /// Special transaction id used for initial copy append operation.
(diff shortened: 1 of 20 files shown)


src/moonlink/src/storage/wal/wal_persistence_metadata.rs
```diff
@@ -4,7 +4,9 @@ use serde::{Deserialize, Serialize};
 /// WAL persistence metadata, which will be persisted into iceberg snapshot property.
 /// WARNING: it's supposed to be small.
 #[derive(Clone, Debug, Serialize, Deserialize, PartialEq)]
-pub(crate) struct WalPersistenceMetadata {
-    /// Persisted WAL file number.
-    pub(crate) persisted_file_num: u32,
+pub struct IcebergWalMetadata {
```
dentiny: mind renaming it? it reads to me, it's a wal metadata for iceberg

dentiny: We should consider supporting schema evolution here (no need in this PR)

src/moonlink/src/storage/mooncake_table/table_snapshot.rs
```diff
@@ -131,8 +131,6 @@ pub struct IcebergSnapshotPayload {
     pub(crate) uuid: uuid::Uuid,
     /// Flush LSN.
     pub(crate) flush_lsn: u64,
-    /// WAL persistence metadata.
```
dentiny, line 134: curious why this change? I put it in the payload, because it's of similar concept and purpose as `flush_lsn`
- both indicate persistence state
- both stored at iceberg snapshot metadata

lohpaul9: 1. Iceberg snapshot payload is created within .update_snapshot, which doesn't have access to the wal_manager. At the same time, the wal_manager requires flush_lsn, which is only known at update_snapshot. I didn't want to populate iceberg snapshot payload after .update_snapshot, so I thought somewhere else is better
2. I don't actually need to 'pass' it around so many places and so that would simplify the logic. Determining the wal persistence metadata exactly when we receive the IcebergSnapshotEvent simplifies the logic instead of trying to determine it elsewhere and passing it around through the snapshot paylaod

src/moonlink/src/storage/mooncake_table.rs
```diff
@@ -1164,23 +1167,37 @@ impl MooncakeTable {
             .unwrap();
     }
 
-    /// Spawns a background task to persist and truncate WAL if needed.
-    /// Returns true if there was a task spawned, false otherwise.
+    /// Uses the wal_persistence_metadata and iceberg LSN from the latest iceberg snapshot
```
dentiny: let's use 
- either both english words `wal persistence metadata` and `iceberg LSN`, 
- or both variable / type name, like `[wal_persistence_metadata]` and `[iceberg_flush_lsn]`

src/moonlink/src/storage/mooncake_table/table_snapshot.rs
dentiny: > Iceberg snapshot payload is created within .update_snapshot, which doesn't have access to the wal_manager.

schema update sets the value at table handler: https://github.com/Mooncake-Labs/moonlink/blob/6fcedd3547c66881629861a16dff1a885e755118/src/moonlink/src/table_handler.rs#L350

dentiny: not quite sure about (2)...
I prefer to put both in payload simply because they share similar purpose and same handling..

lohpaul9:
closes #1121

src/moonlink/src/storage/mooncake_table/table_snapshot.rs
lohpaul9: I've put it all in the iceberg snapshot payload. The wal metadata for this payload is now computed in mooncake table's create_snapshot_impl

lohpaul9: Thanks for the suggestion

Issue #234, opened 2025-05-23
[ISSUE] Dont await flush handle in event loop


nbiscaro:
## Summary

Currently we are awaiting the flush handle in the main event loop. This will block. 

We should create an mpsc channel that adds finished flush tasks. We will process these tasks in the main event loop. 

## Context

Why is this issue important? Link any relevant discussions, designs, or documents.

## Tasks

- [ ] 

## Notes

Any extra notes or details that might be useful.

PR #1204, opened 2025-08-04
[AsyncFlush 3/3] Use asynchronous flush

nbiscaro:
## Summary

Final PR for implementing asynchronous flush events for both streaming and non-streaming transactions.

Previous PRs: 
Use globally unique batch ids: https://github.com/Mooncake-Labs/moonlink/pull/1031
Add streaming and non-streaming state to same snapshot: https://github.com/Mooncake-Labs/moonlink/pull/1100

In this PR, we change all table flush to be asynchronous. We spawn the flush as background task and send `TableEvent::FlushResult` on complete, at which point we apply the results of the flush. Given we may have multiple concurrent flushes at once we track a queue of pending flush LSNs. Before iceberg snapshot we ensure there is no pending LSN less than the LSN we want to persist at. 

## Related Issues

Closes https://github.com/Mooncake-Labs/moonlink/issues/234

## Changes

- 
- 
- 

## Checklist

- [ ] Code builds correctly

### DIFF
--- a/src/moonlink/src/storage/mooncake_table.rs
+++ b/src/moonlink/src/storage/mooncake_table.rs
@@ -206,7 +206,8 @@ pub struct SnapshotTask {
     new_rows: Option<SharedRowBufferSnapshot>,
     new_mem_indices: Vec<Arc<MemIndex>>,
     /// Assigned (non-zero) after a commit event.
-    new_commit_lsn: u64,
+    /// Inherits the previous snapshot tasks commit LSN baseline on snapshot.
+    commit_lsn_baseline: u64,
     /// Assigned at a flush operation.
     new_flush_lsn: Option<u64>,
     new_commit_point: Option<RecordLocation>,
@@ -243,7 +244,7 @@ impl SnapshotTask {
             new_record_batches: Vec::new(),
             new_rows: None,
             new_mem_indices: Vec::new(),
-            new_commit_lsn: 0,
+            commit_lsn_baseline: 0,
             new_flush_lsn: None,
             new_commit_point: None,
             new_streaming_xact: Vec::new(),
@@ -289,7 +290,7 @@ impl SnapshotTask {
 
     pub fn should_create_snapshot(&self) -> bool {
         // If mooncake has new transaction commits.
-        self.new_commit_lsn > 0
+        self.commit_lsn_baseline > 0
             || self.force_empty_iceberg_payload
         // If mooncake table accumulated large enough writes.
             || !self.new_disk_slices.is_empty()
@@ -743,7 +744,7 @@ impl MooncakeTable {
     }
 
     pub fn commit(&mut self, lsn: u64) {
-        self.next_snapshot_task.new_commit_lsn = lsn;
+        self.next_snapshot_task.commit_lsn_baseline = lsn;
         self.next_snapshot_task.new_commit_point = Some(self.mem_slice.get_commit_check_point());
         assert!(
             self.next_snapshot_task.new_deletions.is_empty()
@@ -945,7 +946,8 @@ impl MooncakeTable {
         self.next_snapshot_task = SnapshotTask::new(self.metadata.config.clone());
         // Carry forward the commit baseline
         // This is important if we have a pending flush that will be added to the next snapshot task.
-        self.next_snapshot_task.new_commit_lsn = next_snapshot_task.new_commit_lsn;
+        // Otherwise, if we simply reset the `commit_lsn_baseline` to zero, the ongoing flush will finish and set `flush_lsn`. Then we have a snapshot task where `commit_lsn` < `flush_lsn` which breaks our invariant.
+        self.next_snapshot_task.commit_lsn_baseline = next_snapshot_task.commit_lsn_baseline;
 
         let cur_snapshot = self.snapshot.clone();
 
--- a/src/moonlink/src/storage/mooncake_table.rs
+++ b/src/moonlink/src/storage/mooncake_table.rs
@@ -453,8 +453,8 @@ pub struct MooncakeTable {
     /// WAL manager, used to persist WAL events.
     wal_manager: WalManager,
 
-    /// LSN of pending flushes.
-    pub pending_flush_lsns: BTreeSet<u64>,
+    /// LSN of ongoing flushes.
+    pub ongoing_flush_lsns: BTreeSet<u64>,
 }
 
 impl MooncakeTable {
@@ -550,7 +550,7 @@ impl MooncakeTable {
             last_iceberg_snapshot_lsn,
             table_notify: None,
             wal_manager,
-            pending_flush_lsns: BTreeSet::new(),
+            ongoing_flush_lsns: BTreeSet::new(),
         })
     }
 
@@ -877,7 +877,7 @@ impl MooncakeTable {
 
         let table_notify_tx = self.table_notify.as_ref().unwrap().clone();
 
-        if self.mem_slice.is_empty() || self.pending_flush_lsns.contains(&lsn) {
+        if self.mem_slice.is_empty() || self.ongoing_flush_lsns.contains(&lsn) {
             self.set_next_flush_lsn(lsn);
             tokio::task::spawn(async move {
                 table_notify_tx
@@ -907,7 +907,7 @@ impl MooncakeTable {
     }
 
     pub fn get_min_pending_flush_lsn(&self) -> u64 {
-        self.pending_flush_lsns
+        self.ongoing_flush_lsns
             .iter()
             .next()
             .copied()
@@ -916,14 +916,14 @@ impl MooncakeTable {
 
     pub fn insert_pending_flush_lsn(&mut self, lsn: u64) {
         assert!(
-            self.pending_flush_lsns.insert(lsn),
+            self.ongoing_flush_lsns.insert(lsn),
             "LSN {lsn} already in pending flush LSNs"
         );
     }
 
     pub fn remove_pending_flush_lsn(&mut self, lsn: u64) {
         assert!(
-            self.pending_flush_lsns.remove(&lsn),
+            self.ongoing_flush_lsns.remove(&lsn),
             "LSN {lsn} not found in pending flush LSNs"
         );
     }
--- a/src/moonlink/src/storage/mooncake_table.rs
+++ b/src/moonlink/src/storage/mooncake_table.rs
@@ -698,11 +698,6 @@ impl MooncakeTable {
         self.last_iceberg_snapshot_lsn
     }
 
-    #[cfg(test)]
-    pub(crate) fn set_iceberg_snapshot_lsn(&mut self, lsn: u64) {
-        self.last_iceberg_snapshot_lsn = Some(lsn);
-    }
-
     pub(crate) fn get_state_for_reader(
         &self,
     ) -> (Arc<RwLock<SnapshotTableState>>, watch::Receiver<u64>) {
--- a/src/moonlink/src/storage/mooncake_table.rs
+++ b/src/moonlink/src/storage/mooncake_table.rs
@@ -185,6 +185,12 @@ impl Snapshot {
     }
 }
 
+struct RecordBatchWithDeletionVector {
+    batch_id: u64,
+    record_batch: Arc<RecordBatch>,
+    deletion_vector: Option<BatchDeletionVector>,
+}
+
 #[derive(Default)]
 pub struct SnapshotTask {
     /// Mooncake table config.
@@ -196,7 +202,7 @@ pub struct SnapshotTask {
     disk_file_lsn_map: HashMap<FileId, u64>,
     new_deletions: Vec<RawDeletionRecord>,
     /// Pair of <batch id, record batch, optional deletion vector for streaming batches>.
-    new_record_batches: Vec<(u64, Arc<RecordBatch>, Option<BatchDeletionVector>)>,
+    new_record_batches: Vec<RecordBatchWithDeletionVector>,
     new_rows: Option<SharedRowBufferSnapshot>,
     new_mem_indices: Vec<Arc<MemIndex>>,
     /// Assigned (non-zero) after a commit event.
@@ -711,7 +717,11 @@ impl MooncakeTable {
         if let Some(batch) = self.mem_slice.append(lookup_key, row, identity_for_key)? {
             self.next_snapshot_task
                 .new_record_batches
-                .push((batch.0, batch.1, None));
+                .push(RecordBatchWithDeletionVector {
+                    batch_id: batch.0,
+                    record_batch: batch.1,
+                    deletion_vector: None,
+                });
         }
         Ok(())
     }
@@ -773,7 +783,11 @@ impl MooncakeTable {
         if let Some(batch) = new_batch {
             self.next_snapshot_task
                 .new_record_batches
-                .push((batch.0, batch.1, None));
+                .push(RecordBatchWithDeletionVector {
+                    batch_id: batch.0,
+                    record_batch: batch.1,
+                    deletion_vector: None,
+                });
         }
         self.next_snapshot_task.new_mem_indices.push(index.clone());
 
--- a/src/moonlink/src/storage/mooncake_table.rs
+++ b/src/moonlink/src/storage/mooncake_table.rs
@@ -898,6 +898,7 @@ impl MooncakeTable {
         Ok(())
     }
 
+    // Sets the flush LSN for the next iceberg snapshot. Note that we can only set the flush LSN if it's greater than the current min pending flush LSN. Otherwise, LSNs will be persisted to iceberg in the wrong order.
     fn set_next_flush_lsn(&mut self, lsn: u64) {
         let min_pending_lsn = self.get_min_pending_flush_lsn();
         if lsn < min_pending_lsn {
--- a/src/moonlink/src/storage/mooncake_table.rs
+++ b/src/moonlink/src/storage/mooncake_table.rs
@@ -58,7 +58,7 @@ use delete_vector::BatchDeletionVector;
 pub(crate) use disk_slice::DiskSliceWriter;
 use mem_slice::MemSlice;
 pub(crate) use snapshot::{PuffinDeletionBlobAtRead, SnapshotTableState};
-use std::collections::{HashMap, HashSet};
+use std::collections::{BTreeSet, HashMap, HashSet};
 use std::path::PathBuf;
 use std::sync::Arc;
 use table_snapshot::{IcebergSnapshotImportResult, IcebergSnapshotIndexMergeResult};
@@ -446,6 +446,9 @@ pub struct MooncakeTable {
 
     /// WAL manager, used to persist WAL events.
     wal_manager: WalManager,
+
+    /// LSN of pending flushes.
+    pub pending_flush_lsns: BTreeSet<u64>,
 }
 
 impl MooncakeTable {
@@ -541,6 +544,7 @@ impl MooncakeTable {
             last_iceberg_snapshot_lsn,
             table_notify: None,
             wal_manager,
+            pending_flush_lsns: BTreeSet::new(),
         })
     }
 
@@ -687,6 +691,11 @@ impl MooncakeTable {
         self.last_iceberg_snapshot_lsn
     }
 
+    #[cfg(test)]
+    pub(crate) fn set_iceberg_snapshot_lsn(&mut self, lsn: u64) {
+        self.last_iceberg_snapshot_lsn = Some(lsn);
+    }
+
     pub(crate) fn get_state_for_reader(
         &self,
     ) -> (Arc<RwLock<SnapshotTableState>>, watch::Receiver<u64>) {
@@ -754,56 +763,112 @@ impl MooncakeTable {
         self.mem_slice.get_num_rows() >= self.metadata.config.mem_slice_size
     }
 
-    /// Flush `mem_slice` into parquet files and return the resulting `DiskSliceWriter`.
-    ///
-    /// When `snapshot_task` is provided, new batches and indices are recorded so
-    /// that they can be included in the next snapshot.  The `lsn` parameter
-    /// specifies the commit LSN for the flushed data.  When `lsn` is `None` the
-    /// caller is responsible for setting the final LSN on the returned
-    /// `DiskSliceWriter`.
-    async fn flush_mem_slice(
-        mem_slice: &mut MemSlice,
-        metadata: &Arc<TableMetadata>,
-        next_file_id: u32,
-        lsn: Option<u64>,
-        snapshot_task: Option<&mut SnapshotTask>,
-    ) -> Result<DiskSliceWriter> {
+    /// Drains the current mem slice and prepares a disk slice for flushing.
+    /// Adds current mem slice batches and indices to `next_snapshot_task`.
+    fn prepare_disk_slice(&mut self, lsn: u64) -> Result<DiskSliceWriter> {
         // Finalize the current batch (if needed)
-        let (new_batch, batches, index) = mem_slice.drain().unwrap();
+        let (new_batch, batches, index) = self.mem_slice.drain()?;
 
         let index = Arc::new(index);
-        if let Some(task) = snapshot_task {
-            if let Some(batch) = new_batch {
-                task.new_record_batches.push((batch.0, batch.1, None));
-            }
-            task.new_mem_indices.push(index.clone());
+        if let Some(batch) = new_batch {
+            self.next_snapshot_task
+                .new_record_batches
+                .push((batch.0, batch.1, None));
         }
+        self.next_snapshot_task.new_mem_indices.push(index.clone());
 
-        let path = metadata.path.clone();
-        let parquet_flush_threshold_size = metadata.config.disk_slice_parquet_file_size;
+        let path = self.metadata.path.clone();
+        let parquet_flush_threshold_size = self.metadata.config.disk_slice_parquet_file_size;
+        let next_file_id = self.next_file_id;
+        self.next_file_id += 1;
 
-        let mut disk_slice = DiskSliceWriter::new(
-            metadata.schema.clone(),
+        let disk_slice = DiskSliceWriter::new(
+            self.metadata.schema.clone(),
             path,
             batches,
-            lsn,
+            Some(lsn),
             next_file_id,
             index,
             parquet_flush_threshold_size,
         );
 
-        disk_slice.write().await?;
         Ok(disk_slice)
     }
 
-    // UNDONE(BATCH_INSERT):
-    // Flush uncommitted batches from big batch insert, whether how much record batch there is.
-    //
-    // This function
-    // - tracks all record batches by current snapshot task
-    // - persists all full batch records to local filesystem
-    pub async fn flush(&mut self, lsn: u64) -> Result<()> {
-        // Sanith check flush LSN doesn't regress.
+    /// Flushes the disk slice for the transaction.
+    fn flush_disk_slice(
+        &mut self,
+        disk_slice: &mut DiskSliceWriter,
+        table_notify_tx: Sender<TableEvent>,
+        xact_id: Option<u32>,
+    ) {
+        let path = self.metadata.path.clone();
+        let schema = self.metadata.schema.clone();
+        let batches = disk_slice.take_input_batches();
+        let lsn = disk_slice.lsn();
+        let next_file_id = disk_slice.table_auto_incr_id;
+        let old_index = disk_slice.old_index().clone();
+        let parquet_flush_threshold_size = self.metadata.config.disk_slice_parquet_file_size;
+
+        if let Some(lsn) = lsn {
+            self.insert_pending_flush_lsn(lsn);
+        } else {
+            assert!(
+                xact_id.is_some(),
+                "LSN should be none for non streaming flush"
+            );
+        }
+
+        tokio::task::spawn(async move {
+            let mut disk_slice_clone = DiskSliceWriter::new(
+                schema,
+                path,
+                batches,
+                lsn,
+                next_file_id,
+                old_index,
+                parquet_flush_threshold_size,
+            );
+            let flush_result = disk_slice_clone.write().await;
+            match flush_result {
+                Ok(()) => {
+                    table_notify_tx
+                        .send(TableEvent::FlushResult {
+                            xact_id,
+                            flush_result: Some(Ok(disk_slice_clone)),
+                        })
+                        .await
+                        .unwrap();
+                }
+                Err(e) => {
+                    table_notify_tx
+                        .send(TableEvent::FlushResult {
+                            xact_id,
+                            flush_result: Some(Err(e)),
+                        })
+                        .await
+                        .unwrap();
+                }
+            }
+        });
+    }
+
+    /// Applies the result of a flush to the snapshot task.
+    /// Adds the disk slice to `next_snapshot_task`.
+    pub fn apply_flush_result(&mut self, disk_slice: DiskSliceWriter) {
+        let lsn = disk_slice
+            .lsn()
+            .expect("LSN should never be none for non streaming flush");
+        self.pending_flush_lsns.remove(&lsn);
+        self.set_next_flush_lsn(lsn);
+        self.next_snapshot_task.new_disk_slices.push(disk_slice);
+    }
+
+    /// Drains the current mem slice and create a disk slice.
+    /// Flushes the disk slice.
+    /// Adds the disk slice to `next_snapshot_task`.
+    pub fn flush(&mut self, lsn: u64) -> Result<()> {
+        // Sanity check flush LSN doesn't regress.
         assert!(
             self.next_snapshot_task.new_flush_lsn.is_none()
                 || self.next_snapshot_task.new_flush_lsn.unwrap() <= lsn,
@@ -811,25 +876,57 @@ impl MooncakeTable {
             self.next_snapshot_task.new_flush_lsn,
             lsn,
         );
-        self.next_snapshot_task.new_flush_lsn = Some(lsn);
 
-        if self.mem_slice.is_empty() {
+        let table_notify_tx = self.table_notify.as_ref().unwrap().clone();
+
+        if self.mem_slice.is_empty() || self.pending_flush_lsns.contains(&lsn) {
+            self.set_next_flush_lsn(lsn);
+            tokio::task::spawn(async move {
+                table_notify_tx
+                    .send(TableEvent::FlushResult {
+                        xact_id: None,
+                        flush_result: None,
+                    })
+                    .await
+                    .unwrap();
+            });
             return Ok(());
         }
 
-        let next_file_id = self.next_file_id;
-        self.next_file_id += 1;
-        let disk_slice = Self::flush_mem_slice(
-            &mut self.mem_slice,
-            &self.metadata,
-            next_file_id,
-            Some(lsn),
-            Some(&mut self.next_snapshot_task),
-        )
-        .await?;
-        self.next_snapshot_task.new_disk_slices.push(disk_slice);
+        let mut disk_slice = self.prepare_disk_slice(lsn)?;
+
+        self.flush_disk_slice(&mut disk_slice, table_notify_tx, None);
 
         Ok(())
+    }
+
+    fn set_next_flush_lsn(&mut self, lsn: u64) {
+        let min_pending_lsn = self.get_min_pending_flush_lsn();
+        if lsn < min_pending_lsn {
+            self.next_snapshot_task.new_flush_lsn = Some(lsn);
+        }
+    }
+
+    pub fn get_min_pending_flush_lsn(&self) -> u64 {
+        self.pending_flush_lsns
+            .iter()
+            .next()
+            .copied()
+            .unwrap_or(u64::MAX)
+    }
+
+    pub fn insert_pending_flush_lsn(&mut self, lsn: u64) {
+        assert!(
+            self.pending_flush_lsns.insert(lsn),
+            "LSN {lsn} already in pending flush LSNs"
+        );
+    }
+
+    pub fn remove_pending_flush_lsn(&mut self, lsn: u64) {
+        assert!(
+            self.pending_flush_lsns.remove(&lsn),
+            "LSN {lsn} not found in pending flush LSNs"
+        );
     }
 
     // Create a snapshot of the last committed version, return current snapshot's version and payload to perform iceberg snapshot.
@@ -843,14 +940,20 @@ impl MooncakeTable {
 
         // Re-initialize mooncake table fields.
         self.next_snapshot_task = SnapshotTask::new(self.metadata.config.clone());
+        // Carry forward the commit baseline
+        // This is important if we have a pending flush that will be added to the next snapshot task.
+        self.next_snapshot_task.new_commit_lsn = next_snapshot_task.new_commit_lsn;
 
         let cur_snapshot = self.snapshot.clone();
+
+        let has_pending_flushes = !self.pending_flush_lsns.is_empty();
         // Create a detached task, whose completion will be notified separately.
         tokio::task::spawn(
             Self::create_snapshot_async(
                 cur_snapshot,
                 next_snapshot_task,
                 opt,
+                has_pending_flushes,
                 self.table_notify.as_ref().unwrap().clone(),
             )
             .instrument(info_span!("create_snapshot_async")),
@@ -1112,13 +1215,14 @@ impl MooncakeTable {
         snapshot: Arc<RwLock<SnapshotTableState>>,
         next_snapshot_task: SnapshotTask,
         mut opt: SnapshotOption,
+        has_pending_flushes: bool,
         table_notify: Sender<TableEvent>,
     ) {
         let uuid = std::mem::take(&mut opt.uuid);
         let snapshot_result = snapshot
             .write()
             .await
-            .update_snapshot(next_snapshot_task, opt)
+            .update_snapshot(next_snapshot_task, opt, has_pending_flushes)
             .await;
         table_notify
             .send(TableEvent::MooncakeTableSnapshotResult {
--- a/src/moonlink/src/storage/mooncake_table.rs
+++ b/src/moonlink/src/storage/mooncake_table.rs
@@ -816,15 +816,7 @@ impl MooncakeTable {
         table_notify_tx: Sender<TableEvent>,
         xact_id: Option<u32>,
     ) {
-        let path = self.metadata.path.clone();
-        let schema = self.metadata.schema.clone();
-        let batches = disk_slice.take_input_batches();
-        let lsn = disk_slice.lsn();
-        let next_file_id = disk_slice.table_auto_incr_id;
-        let old_index = disk_slice.old_index().clone();
-        let parquet_flush_threshold_size = self.metadata.config.disk_slice_parquet_file_size;
-
-        if let Some(lsn) = lsn {
+        if let Some(lsn) = disk_slice.lsn() {
             self.insert_pending_flush_lsn(lsn);
         } else {
             assert!(
@@ -833,16 +825,8 @@ impl MooncakeTable {
             );
         }
 
+        let mut disk_slice_clone = disk_slice.clone();
         tokio::task::spawn(async move {
-            let mut disk_slice_clone = DiskSliceWriter::new(
-                schema,
-                path,
-                batches,
-                lsn,
-                next_file_id,
-                old_index,
-                parquet_flush_threshold_size,
-            );
             let flush_result = disk_slice_clone.write().await;
             match flush_result {
                 Ok(()) => {
--- a/src/moonlink/src/storage/mooncake_table.rs
+++ b/src/moonlink/src/storage/mooncake_table.rs
@@ -208,6 +208,9 @@ pub struct SnapshotTask {
     /// Assigned (non-zero) after a commit event.
     /// Inherits the previous snapshot tasks commit LSN baseline on snapshot.
     commit_lsn_baseline: u64,
+    /// Commit LSN baseline of the previous snapshot task.
+    /// We use this to determine if commit_lsn_baseline has been updated.
+    prev_commit_lsn_baseline: u64,
     /// Assigned at a flush operation.
     new_flush_lsn: Option<u64>,
     new_commit_point: Option<RecordLocation>,
@@ -232,6 +235,9 @@ pub struct SnapshotTask {
     /// ---- States have been recorded by mooncake snapshot, and persisted into iceberg table ----
     /// These persisted items will be reflected to mooncake snapshot in the next invocation of periodic mooncake snapshot operation.
     iceberg_persisted_records: IcebergPersistedRecords,
+
+    /// Minimum LSN of ongoing flushes.
+    min_ongoing_flush_lsn: u64,
 }
 
 impl SnapshotTask {
@@ -245,6 +251,7 @@ impl SnapshotTask {
             new_rows: None,
             new_mem_indices: Vec::new(),
             commit_lsn_baseline: 0,
+            prev_commit_lsn_baseline: 0,
             new_flush_lsn: None,
             new_commit_point: None,
             new_streaming_xact: Vec::new(),
@@ -257,6 +264,7 @@ impl SnapshotTask {
             data_compaction_result: DataCompactionResult::default(),
             // Iceberg persistence result.
             iceberg_persisted_records: IcebergPersistedRecords::default(),
+            min_ongoing_flush_lsn: u64::MAX,
         }
     }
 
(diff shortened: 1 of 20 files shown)


dentiny:
> thread 'tokio-runtime-worker' panicked at src/moonlink/src/storage/mooncake_table/snapshot_validation.rs:27:21:
assertion failed: `(left < right)`
  left: `11`,
 right: `11`

Looks like a real problem.

src/moonlink/src/table_notify.rs
```diff
@@ -170,6 +171,12 @@ pub enum TableEvent {
     PeriodicalPersistTruncateWalResult {
         result: Result<WalPersistAndTruncateResult>,
     },
+    FlushResult {
```
dentiny, line 184: Please define its own formatter, otherwise it's very spammy on logging.

dentiny:
FYI: I did a few fixes yesterday and today, suggest to rebase upon main.

dentiny:
Just fix a bug in table handler side, please rebase for chaos test stability. :)

src/moonlink/src/table_notify.rs
nbiscaro: Added its own debug formatter that only prints the batch ids in the disk slice

nbiscaro:
> > thread 'tokio-runtime-worker' panicked at src/moonlink/src/storage/mooncake_table/snapshot_validation.rs:27:21:
> > assertion failed: `(left < right)`
> > left: `11`,
> > right: `11`
> 
> Looks like a real problem.

Fixed this (amongst others) and added relevant unit tests to guard against the issue.

dentiny:
```sh
Test test_chaos_on_local_fs_with_index_merge is with random seed 1754428183492768295
Test test_chaos_on_local_fs_with_data_compaction is with random seed 1754428183492836537
Test test_chaos_on_local_fs_with_no_background_maintenance is with random seed 1754428183493092372

thread 'tokio-runtime-worker' panicked at src/moonlink/src/storage/mooncake_table/test_utils.rs:175:5:
assertion `left == right` failed
  left: [3, 7, 10, 13, 16, 18, 19, 22, 23, 26, 27, 28, 30, 32, 33, 34, 35, 36, 37, 38, 39, 40, 43, 44, 45, 46, 48, 49, 50]
 right: [3, 7, 10, 13, 16, 18, 19, 22, 23, 26, 27, 28, 30, 32, 33, 34, 35, 36, 37, 38, 39, 40, 43, 44, 45, 46, 49, 50]
```
I used the random seed to check in the main branch for a few, cannot repro, likely new issue introduce in this PR, but not existing ones.

dentiny:
```sh
thread 'tokio-runtime-worker' panicked at src/moonlink/src/storage/mooncake_table/snapshot.rs:893:21:
assertion failed: res
stack backtrace:
   0: __rustc::rust_begin_unwind
```
which comes from 
https://github.com/Mooncake-Labs/moonlink/blob/8bcd9ce5af9163a0a74e7c753099635676863070/src/moonlink/src/storage/mooncake_table/snapshot.rs#L887-L893, seems still an issue?

dentiny:
Please don't merge until we figure out the deletion vector issue: https://github.com/Mooncake-Labs/moonlink/pull/1243
Also suggest to rebase main branch, I add a few more sanity checks, thanks.

nbiscaro:
> Please don't merge until we figure out the deletion vector issue: #1243 Also suggest to rebase main branch, I add a few more sanity checks, thanks.

Issue is resolved here: https://github.com/Mooncake-Labs/moonlink/pull/1255

src/moonlink/src/table_handler/tests.rs
```diff
@@ -2118,3 +2103,35 @@ fn test_deletion_vector_capacity_exceeded_minimal() {
     // This should panic - trying to delete row_id=5 when capacity is only 3
     let _ = deletion_vector.delete_row(5);
 }
+
+#[tokio::test]
+async fn test_stream_delete_from_stream_memslice_row_debug() {
```
dentiny: What does it mean by `_debug`? Should we clean it up?

src/moonlink/src/table_handler/tests.rs
```diff
@@ -2118,3 +2103,35 @@ fn test_deletion_vector_capacity_exceeded_minimal() {
...
+
+    // Phase 3: Verify data is NOT visible before commit
+    env.set_table_commit_lsn(0);
+    env.set_replication_lsn(stream_commit_lsn + 5);
+    env.set_snapshot_lsn(0);
+    env.verify_snapshot(stream_commit_lsn, &[]).await;
+
+    // Phase 4: Commit the streaming transaction
+    env.stream_commit(stream_commit_lsn, xact_id).await;
+    println!("Debug: Verifying final state post-commit (should be empty)");
```
dentiny: Should we clean it up? I mean it says "debug".

src/moonlink/src/storage/mooncake_table.rs
```diff
@@ -687,6 +691,11 @@ impl MooncakeTable {
         self.last_iceberg_snapshot_lsn
     }
 
+    #[cfg(test)]
+    pub(crate) fn set_iceberg_snapshot_lsn(&mut self, lsn: u64) {
```
dentiny: Don't we already have a `set_iceberg_snapshot_res`?
And we could always set iceberg snapshot LSN at construction?

src/moonlink/src/storage/mooncake_table.rs
```diff
@@ -754,84 +763,172 @@ impl MooncakeTable {
...
         let index = Arc::new(index);
-        if let Some(task) = snapshot_task {
-            if let Some(batch) = new_batch {
-                task.new_record_batches.push((batch.0, batch.1, None));
-            }
-            task.new_mem_indices.push(index.clone());
+        if let Some(batch) = new_batch {
+            self.next_snapshot_task
+                .new_record_batches
+                .push((batch.0, batch.1, None));
```
dentiny: let me try again, do you mind we using struct instead of tuples? Struct provides field naming, which is much more readable.

src/moonlink/src/storage/mooncake_table.rs
```diff
@@ -754,84 +763,172 @@ impl MooncakeTable {
...
+            self.insert_pending_flush_lsn(lsn);
+        } else {
+            assert!(
+                xact_id.is_some(),
+                "LSN should be none for non streaming flush"
+            );
+        }
+
+        tokio::task::spawn(async move {
+            let mut disk_slice_clone = DiskSliceWriter::new(
```
dentiny: Curious why name it `clone`?

src/moonlink/src/storage/mooncake_table.rs
```diff
@@ -754,84 +763,172 @@ impl MooncakeTable {
...
+                lsn,
+                next_file_id,
+                old_index,
+                parquet_flush_threshold_size,
+            );
+            let flush_result = disk_slice_clone.write().await;
+            match flush_result {
+                Ok(()) => {
+                    table_notify_tx
+                        .send(TableEvent::FlushResult {
```
dentiny, line 844: (not sth we need to address now) If a disk write fails, do we leave orphaned disk files?

src/moonlink/src/storage/mooncake_table.rs
```diff
@@ -754,84 +763,172 @@ impl MooncakeTable {
...
+                            xact_id,
+                            flush_result: Some(Ok(disk_slice_clone)),
+                        })
+                        .await
+                        .unwrap();
+                }
+                Err(e) => {
+                    table_notify_tx
```
dentiny, line 852: You need to coordinate with drop table, otherwise send would fail.

dentiny: https://github.com/Mooncake-Labs/moonlink/blob/dff3672daad96bca992de37315fd19f529b844ad/src/moonlink/src/table_handler/table_handler_state.rs#L327-L341

src/moonlink/src/storage/mooncake_table.rs
```diff
@@ -754,84 +763,172 @@ impl MooncakeTable {
...
-        .await?;
-        self.next_snapshot_task.new_disk_slices.push(disk_slice);
+        let mut disk_slice = self.prepare_disk_slice(lsn)?;
+
+        self.flush_disk_slice(&mut disk_slice, table_notify_tx, None);
 
         Ok(())
     }
 
+    fn set_next_flush_lsn(&mut self, lsn: u64) {
```
dentiny: Need documentation or rename, not easy to tell.

src/moonlink/src/storage/mooncake_table.rs
```diff
@@ -754,84 +763,172 @@ impl MooncakeTable {
...
+        });
+    }
+
+    /// Applies the result of a flush to the snapshot task.
+    /// Adds the disk slice to `next_snapshot_task`.
+    pub fn apply_flush_result(&mut self, disk_slice: DiskSliceWriter) {
+        let lsn = disk_slice
+            .lsn()
+            .expect("LSN should never be none for non streaming flush");
+        self.pending_flush_lsns.remove(&lsn);
```
dentiny: Please check return value.

src/moonlink/src/storage/mooncake_table/transaction_stream.rs
```diff
@@ -343,36 +342,48 @@ impl MooncakeTable {
...
-        Ok(())
-    }
-
     /// Applies the result of a streaming flush to the stream state.
     /// Decrements the pending flush count for this transaction.
     /// Handles commit and abort cleanup.
     /// Removes in memory indices and record batches from the stream state.
     pub fn apply_stream_flush_result(&mut self, xact_id: u32, mut disk_slice: DiskSliceWriter) {
+        if let Some(lsn) = disk_slice.lsn() {
+            self.pending_flush_lsns.remove(&lsn);
```
dentiny: Please check return value.

src/moonlink/src/storage/mooncake_table.rs
```diff
@@ -446,6 +446,9 @@ pub struct MooncakeTable {
 
     /// WAL manager, used to persist WAL events.
     wal_manager: WalManager,
+
+    /// LSN of pending flushes.
+    pub pending_flush_lsns: BTreeSet<u64>,
```
dentiny: `pending` means not yet decided: https://www.merriam-webster.com/dictionary/pending
I'm not sure if it means (1) ongoing, (2) in flush queue, if we have one; (3) in-memory, will be flushed.
```suggestion
    pub ongoing_flush_lsns: BTreeSet<u64>,
```

src/moonlink/src/storage/mooncake_table.rs
```diff
@@ -843,14 +940,20 @@ impl MooncakeTable {
 
         // Re-initialize mooncake table fields.
         self.next_snapshot_task = SnapshotTask::new(self.metadata.config.clone());
+        // Carry forward the commit baseline
+        // This is important if we have a pending flush that will be added to the next snapshot task.
+        self.next_snapshot_task.new_commit_lsn = next_snapshot_task.new_commit_lsn;
```
dentiny: - You need to rename the variable, it's not only **new** LSNs any more
- Could you please elaborate what if we don't carry over? 

TBH
> This is important if we have a pending flush that will be added to the next snapshot task.

Tells me nothing

src/moonlink/src/storage/mooncake_table.rs
```diff
@@ -843,14 +940,20 @@ impl MooncakeTable {
...
 
         let cur_snapshot = self.snapshot.clone();
+
+        let has_pending_flushes = !self.pending_flush_lsns.is_empty();
```
dentiny:
```suggestion
        let has_ongoing_flushes = !self.ongoing_flush_lsns.is_empty();
```

src/moonlink/src/storage/mooncake_table.rs
```diff
@@ -754,84 +763,172 @@ impl MooncakeTable {
...
     }
 
+    fn set_next_flush_lsn(&mut self, lsn: u64) {
+        let min_pending_lsn = self.get_min_pending_flush_lsn();
+        if lsn < min_pending_lsn {
+            self.next_snapshot_task.new_flush_lsn = Some(lsn);
+        }
+    }
+
+    pub fn get_min_pending_flush_lsn(&self) -> u64 {
```
dentiny: nit: doc on fallback behavior

src/moonlink/src/storage/mooncake_table/disk_slice.rs
```diff
@@ -51,6 +53,16 @@ pub struct DiskSliceWriter {
     files: Vec<(MooncakeDataFileRef, DiskFileAttrs)>,
 }
 
+impl std::fmt::Debug for DiskSliceWriter {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        let batch_ids: Vec<u64> = self.batches.iter().map(|batch| batch.id).collect();
+        f.debug_struct("DiskSliceWriter")
+            .field("batch_ids", &batch_ids)
+            .field("lsn", &self.writer_lsn)
+            .finish()
```
dentiny, line 70: I strongly suggest to print out the number of data files and index blocks, they're super useful in debugging to tell whether we should trigger compaction and index merge.

src/moonlink/src/storage/mooncake_table/snapshot.rs
```diff
@@ -562,7 +563,10 @@ impl SnapshotTableState {
             && (flush_by_new_files_or_maintenance || flush_by_deletion);
 
         // TODO(hjiang): When there's only schema evolution, we should also flush even no flush.
-        if !opt.skip_iceberg_snapshot && (force_empty_iceberg_payload || flush_by_table_write) {
+        if !opt.skip_iceberg_snapshot
+            && (force_empty_iceberg_payload || flush_by_table_write)
+            && !has_pending_flushes
```
dentiny: I don't think it's good (it's correct behavior though), because it means we cannot make iceberg snapshots whenever we have **ANY** background flushes.

dentiny:
Looks like an issue:
```sh
thread 'tokio-runtime-worker' panicked at src/moonlink/src/storage/mooncake_table/snapshot.rs:394:17:
assertion failed: old_entry.puffin_deletion_blob.is_none()
stack backtrace:
   0: __rustc::rust_begin_unwind
```
https://github.com/Mooncake-Labs/moonlink/actions/runs/16791181367/job/47552933256?pr=1204

src/moonlink/src/storage/mooncake_table/snapshot.rs
```diff
@@ -701,21 +705,34 @@ impl SnapshotTableState {
...
 
-            self.committed_deletion_log[cut..]
-                .iter_mut()
-                .for_each(|d| slice.remap_deletion_if_needed(d));
+            for deletion in self.committed_deletion_log[cut..].iter_mut() {
+                if let Some(RecordLocation::DiskFile(file_id, row_idx)) =
+                    slice.remap_deletion_if_needed(deletion)
+                {
+                    // Find the disk file entry and apply the deletion
+                    for (file_ref, disk_entry) in self.current_snapshot.disk_files.iter_mut() {
```
dentiny, line 727: nit (feel free to ignore): one thing I plan to do is get the dependency graph for updating snapshot -- I made mistakes to reorder a few steps.
For example, integrating disk slice function depends on all new data files have been reflected to disk file map; would appreciate if you could mark the precondition.

src/moonlink/src/storage/mooncake_table/transaction_stream.rs
```diff
@@ -512,9 +512,21 @@ impl MooncakeTable {
             .push(TransactionStreamOutput::Commit(commit));
 
         stream_state.status = TransactionStreamStatus::Committed;
+        // We may have pending flushes that we need to stamp with this commit LSN so we store it in stream_state.
+        stream_state.commit_lsn = Some(lsn);
 
         Ok(())
     }
+
+    /// Commit a transaction stream.
```
dentiny, line 524: nit
```suggestion
    /// Commit a transaction stream.
    /// - Flushes any remaining rows from stream mem slice
    /// - Adds all in mem batches and indices to next snapshot task
    /// - Updates deletion records
    /// - Enqueues `TransactionStreamOutput::Commit` for snapshot task
```

nbiscaro:
> Looks like an issue:
> 
> ```shell
> thread 'tokio-runtime-worker' panicked at src/moonlink/src/storage/mooncake_table/snapshot.rs:394:17:
> assertion failed: old_entry.puffin_deletion_blob.is_none()
> stack backtrace:
>    0: __rustc::rust_begin_unwind
> ```
> 
> https://github.com/Mooncake-Labs/moonlink/actions/runs/16791181367/job/47552933256?pr=1204

For a minute I thought we were forgetting to remap local deletions in the case where we are applying flush post-commit, so i applied them there and pushed the change. I now realize we remap these in `integrate_disk_slices` so this was incorrect. Removed that change and seems fine.

nbiscaro:
uhhh i don't know why github closed that...

src/moonlink/src/table_handler/tests.rs
nbiscaro: dup test removed

src/moonlink/src/table_handler/tests.rs
nbiscaro: removed

src/moonlink/src/storage/mooncake_table.rs
nbiscaro: Ah, this was before disk slice was cloneable. we can just directly clone now.

src/moonlink/src/storage/mooncake_table.rs
nbiscaro: Probably. We also leave orphaned files on stream abort. Followup created https://github.com/Mooncake-Labs/moonlink/issues/1260

src/moonlink/src/storage/mooncake_table/snapshot.rs
nbiscaro: fixed to use correct condition of `flush_lsn < min_ongoing_flush_lsn`

src/moonlink/src/storage/mooncake_table/disk_slice.rs
nbiscaro: added

src/moonlink/src/storage/mooncake_table/transaction_stream.rs
nbiscaro: updated to use the `remove_pending_flush_lsn` function which does the assertion.

src/moonlink/src/storage/mooncake_table.rs
nbiscaro: changed

src/moonlink/src/storage/mooncake_table.rs
nbiscaro: `set_iceberg_snapshot_res` takes many complex parameters that are not relevant for this test. I think directly setting the lsn is clean and accomplishes only what we are trying to do. 

If you are concerned about the `#[cfg(test)]` being in with our source code, I have moved it to its own `impl` in the `table_creation_test_utils` file. 

We should actually do the same with any other test specific code we have behind this flag. Then we don't have to implement hacky workarounds when testing and can access the data we need without cluttering or complicating actual implementation. https://github.com/Mooncake-Labs/moonlink/issues/1261

src/moonlink/src/storage/mooncake_table/snapshot.rs
nbiscaro: added

dentiny:
I will review this PR as the first thing at daytime

src/moonlink/src/storage/mooncake_table.rs
```diff
@@ -283,7 +290,7 @@ impl SnapshotTask {
 
     pub fn should_create_snapshot(&self) -> bool {
         // If mooncake has new transaction commits.
-        self.new_commit_lsn > 0
+        self.commit_lsn_baseline > 0
```
dentiny: issue here: we will create new mooncake snapshot every 500ms

src/moonlink/src/storage/mooncake_table.rs
```diff
@@ -754,84 +769,166 @@ impl MooncakeTable {
...
+    // We fallback to u64::MAX if there are no pending flush LSNs so that the LSN is always greater than the flush LSN and the iceberg snapshot can proceed.
+    pub fn get_min_pending_flush_lsn(&self) -> u64 {
+        self.ongoing_flush_lsns
+            .iter()
+            .next()
+            .copied()
+            .unwrap_or(u64::MAX)
+    }
+
+    pub fn insert_pending_flush_lsn(&mut self, lsn: u64) {
```
dentiny: For naming consistency, we should call them
```
pub fn insert_ongoing_flush_lsn
pub fn remove_ongoing_flush_lsn
```

src/moonlink/src/storage/mooncake_table.rs
```diff
@@ -754,84 +769,166 @@ impl MooncakeTable {
...
+        self.flush_disk_slice(&mut disk_slice, table_notify_tx, None);
 
         Ok(())
     }
 
+    // Sets the flush LSN for the next iceberg snapshot. Note that we can only set the flush LSN if it's greater than the current min pending flush LSN. Otherwise, LSNs will be persisted to iceberg in the wrong order.
+    fn set_next_flush_lsn(&mut self, lsn: u64) {
+        let min_pending_lsn = self.get_min_pending_flush_lsn();
+        if lsn < min_pending_lsn {
+            self.next_snapshot_task.new_flush_lsn = Some(lsn);
```
dentiny, line 915: The comment said: only set the flush LSN if it's **greater than** the current min pending flush LSN
But the code reads **smaller than**?

src/moonlink/src/table_handler.rs
```diff
@@ -291,7 +291,7 @@ impl TableHandler {
                         }
                         TableEvent::FinishInitialCopy { start_lsn } => {
                             debug!("finishing initial copy");
-                            if let Err(e) = table.commit_transaction_stream(INITIAL_COPY_XACT_ID, 0).await {
+                            if let Err(e) = table.commit_transaction_stream(INITIAL_COPY_XACT_ID, 0) {
```
dentiny, line 294:
```suggestion
                            if let Err(e) = table.commit_transaction_stream(INITIAL_COPY_XACT_ID, /*lsn=*/0) {
```

src/moonlink/src/table_handler.rs
```diff
@@ -376,7 +376,8 @@ impl TableHandler {
                             table.notify_snapshot_reader(lsn);
 
                             // Process iceberg snapshot and trigger iceberg snapshot if necessary.
-                            if table_handler_state.can_initiate_iceberg_snapshot() {
+                            let min_pending_flush_lsn = table.get_min_pending_flush_lsn();
+                            if TableHandlerState::can_initiate_iceberg_snapshot(lsn, min_pending_flush_lsn, table_handler_state.iceberg_snapshot_result_consumed, table_handler_state.iceberg_snapshot_ongoing) {
```
dentiny, line 380: why not making it a member function? so you don't need to take the later two arguments

src/moonlink/src/storage/mooncake_table.rs
```diff
@@ -754,84 +769,166 @@ impl MooncakeTable {
...
-        if self.mem_slice.is_empty() {
+        let table_notify_tx = self.table_notify.as_ref().unwrap().clone();
+
+        if self.mem_slice.is_empty() || self.ongoing_flush_lsns.contains(&lsn) {
+            self.set_next_flush_lsn(lsn);
+            tokio::task::spawn(async move {
+                table_notify_tx
+                    .send(TableEvent::FlushResult {
+                        xact_id: None,
+                        flush_result: None,
```
dentiny, line 896: Every time when there's nothing to flush, we get an error message at table handler
It doesn't look ideal to me

src/moonlink/src/storage/mooncake_table/transaction_stream.rs
```diff
@@ -424,27 +423,28 @@ impl MooncakeTable {
...
-        if let Some(lsn) = lsn {
-            self.next_snapshot_task.new_flush_lsn = Some(lsn);
-        }
-        self.apply_stream_flush_result(xact_id, disk_slice);
+    pub fn flush_stream(&mut self, xact_id: u32, lsn: Option<u64>) -> Result<()> {
+        // Temporarily remove to drop reference to self
+        let mut stream_state = self
+            .transaction_stream_states
+            .remove(&xact_id)
+            .expect("Stream state not found for xact_id: {xact_id}");
```
dentiny: suggest to attach LSN as well, so ourselves could differentiate stream flush and stream commit

src/moonlink/src/table_handler.rs
```diff
@@ -679,16 +697,24 @@ impl TableHandler {
...
         // and 3. there's no snapshot creation operation ongoing
-
-        let should_force_snapshot = table_handler_state.should_force_snapshot_by_commit_lsn(lsn);
+        // and 4. there's no pending flush LSNs < lsn
+
+        let min_pending_flush_lsn = table.get_min_pending_flush_lsn();
+        let should_force_snapshot = TableHandlerState::should_force_snapshot_by_commit_lsn(
+            lsn,
+            min_pending_flush_lsn,
+            &table_handler_state.table_maintenance_process_status,
```
dentiny, line 707: not sure why not using member function here?

src/moonlink/src/storage/mooncake_table.rs
```diff
@@ -724,7 +739,7 @@ impl MooncakeTable {
     }
 
     pub fn commit(&mut self, lsn: u64) {
-        self.next_snapshot_task.new_commit_lsn = lsn;
+        self.next_snapshot_task.commit_lsn_baseline = lsn;
```
dentiny, line 756: assert LSN never goes back

src/moonlink/src/storage/mooncake_table.rs
```diff
@@ -754,84 +769,166 @@ impl MooncakeTable {
...
-        self.next_snapshot_task.new_disk_slices.push(disk_slice);
+        let mut disk_slice = self.prepare_disk_slice(lsn)?;
+
+        self.flush_disk_slice(&mut disk_slice, table_notify_tx, None);
 
         Ok(())
     }
 
+    // Sets the flush LSN for the next iceberg snapshot. Note that we can only set the flush LSN if it's greater than the current min pending flush LSN. Otherwise, LSNs will be persisted to iceberg in the wrong order.
+    fn set_next_flush_lsn(&mut self, lsn: u64) {
```
dentiny: The naming is confusing, it's an "attempt", no guarantee to set LSN

src/moonlink/src/storage/mooncake_table.rs
```diff
@@ -754,84 +769,166 @@ impl MooncakeTable {
...
-        .await?;
-        self.next_snapshot_task.new_disk_slices.push(disk_slice);
+        let mut disk_slice = self.prepare_disk_slice(lsn)?;
+
+        self.flush_disk_slice(&mut disk_slice, table_notify_tx, None);
 
         Ok(())
     }
 
+    // Sets the flush LSN for the next iceberg snapshot. Note that we can only set the flush LSN if it's greater than the current min pending flush LSN. Otherwise, LSNs will be persisted to iceberg in the wrong order.
```
dentiny:
```suggestion
    // Attempt to set the flush LSN for the next iceberg snapshot. Note that we can only set the flush LSN if it's greater than the current min pending flush LSN. Otherwise, LSNs will be persisted to iceberg in the wrong order.
```
or sth alike

dentiny: I don't really like the API design, essentially there's implicit dependency, which is error prune
- set next LSN assumes certain completed flush LSN has been removed from ongoing set
- but we don't have any assertion to guard against

Either we add assertion to check precondition
or we wrap remove + set logic into the same function

src/moonlink/src/storage/mooncake_table/transaction_stream.rs
```diff
@@ -343,36 +342,36 @@ impl MooncakeTable {
...
+            let commit_lsn = stream_state
+                .commit_lsn
+                .expect("Committed transaction must have commit_lsn");
+
+            // If there's no writer_lsn, it means this is a periodic flush before commit.
+            // In this case we assign the commit_lsn as writer_lsn.
+            if disk_slice.writer_lsn.is_none() {
+                disk_slice.writer_lsn = Some(commit_lsn);
+            }
+
```
dentiny, line 374: nit: it reads to me as long as pending (should rename to `ongoing` for consistency) count is 0, we remove from stream state, better to extract it out, maybe right after L360.

src/moonlink/src/storage/mooncake_table/transaction_stream.rs
```diff
@@ -481,7 +483,7 @@ impl MooncakeTable {
                 .iter()
                 .map(|ptr| ptr.arc_ptr()),
         );
-        self.next_snapshot_task.new_commit_lsn = lsn;
+        self.next_snapshot_task.commit_lsn_baseline = lsn;
```
dentiny, line 487: add an assertion what LSN never goes back

src/moonlink/src/storage/mooncake_table/transaction_stream.rs
```diff
@@ -512,9 +514,21 @@ impl MooncakeTable {
             .push(TransactionStreamOutput::Commit(commit));
 
         stream_state.status = TransactionStreamStatus::Committed;
+        // We may have pending flushes that we need to stamp with this commit LSN so we store it in stream_state.
```
dentiny, line 518:
```suggestion
        // We may have ongoing flushes that we need to stamp with this commit LSN so we store it in stream_state.
```

src/moonlink/src/table_handler/table_handler_state.rs
```diff
@@ -422,10 +436,17 @@ impl TableHandlerState {
     ///
     /// Used to decide whether we could create an iceberg snapshot.
     /// The completion of an iceberg snapshot is **NOT** marked as the finish of snapshot thread, but the handling of its results.
-    /// We can only create a new iceberg snapshot when (1) there's no ongoing iceberg snapshot, (2) previous snapshot results have been acknowledged.
+    /// We can only create a new iceberg snapshot when (1) there's no ongoing iceberg snapshot, (2) previous snapshot results have been acknowledged, (3) there's no pending flush LSNs < flush_lsn
```
dentiny, line 439:
```suggestion
    /// We can only create a new iceberg snapshot when (1) there's no ongoing iceberg snapshot, (2) previous snapshot results have been acknowledged, (3) there's no ongoing flush LSNs < flush_lsn
```

dentiny (commented):
Second round...

src/moonlink/src/storage/mooncake_table/disk_slice.rs
```diff
@@ -247,20 +268,26 @@ impl DiskSliceWriter {
         self.new_index.take()
     }
 
-    pub fn remap_deletion_if_needed(&self, deletion: &mut ProcessedDeletionRecord) {
+    pub fn remap_deletion_if_needed(
```
dentiny, line 271: nit: comment on return value (i.e., when to return None)

src/moonlink/src/storage/mooncake_table.rs
```diff
@@ -1112,13 +1216,14 @@ impl MooncakeTable {
...
         mut opt: SnapshotOption,
+        min_pending_flush_lsn: u64,
         table_notify: Sender<TableEvent>,
     ) {
         let uuid = std::mem::take(&mut opt.uuid);
         let snapshot_result = snapshot
             .write()
             .await
-            .update_snapshot(next_snapshot_task, opt)
+            .update_snapshot(next_snapshot_task, opt, min_pending_flush_lsn)
```
dentiny: curious why don't we place min flush LSN inside of snapshot task?

src/moonlink/src/storage/mooncake_table/snapshot.rs
```diff
@@ -525,8 +526,8 @@ impl SnapshotTableState {
             self.current_snapshot.flush_lsn = Some(new_flush_lsn);
         }
 
-        if task.new_commit_lsn != 0 {
-            self.current_snapshot.snapshot_version = task.new_commit_lsn;
+        if task.commit_lsn_baseline != 0 {
```
dentiny, line 528: nit: I feel like we could directly assign?
```rs
self.current_snapshot.snapshot_version = task.commit_lsn_baseline;
```

dentiny: Also please add assertion so we're sure LSN never goes back

src/moonlink/src/storage/mooncake_table/snapshot.rs
```diff
@@ -611,15 +615,15 @@ impl SnapshotTableState {
         let incoming = take(&mut task.new_record_batches);
         let (streaming_batches, mut non_streaming_batches): (Vec<_>, Vec<_>) = incoming
             .into_iter()
-            .partition(|(id, _, _)| *id < (1u64 << 63));
+            .partition(|batch| batch.batch_id < (1u64 << 63));
```
dentiny, line 617: making it a constant? external contributors won't understand what's the logic here

src/moonlink/src/storage/mooncake_table/snapshot.rs
```diff
@@ -701,21 +706,35 @@ impl SnapshotTableState {
...
 
             // remap deletions written *after* this slice’s LSN
+            // We set to write_lsn - 1 to maintain consistency with deletion LSNs from the streaming case.
             let cut = self.committed_deletion_log.partition_point(|d| {
                 d.lsn
-                    <= write_lsn.expect(
+                    < write_lsn.expect(
                         "Critical: LSN is None after it should have been updated by commit process",
-                    )
+                    ) - 1
```
dentiny: I don't think there's anything functionally wrong here (attempt remap is possible doesn't hurt anyway)
But could you please help me understand on the write lsn - 1? Thank you!

src/moonlink/src/storage/mooncake_table/transaction_stream.rs
```diff
@@ -343,36 +342,36 @@ impl MooncakeTable {
...
-        disk_slice.write().await?;
-        Ok(())
-    }
-
     /// Applies the result of a streaming flush to the stream state.
     /// Decrements the pending flush count for this transaction.
     /// Handles commit and abort cleanup.
     /// Removes in memory indices and record batches from the stream state.
     pub fn apply_stream_flush_result(&mut self, xact_id: u32, mut disk_slice: DiskSliceWriter) {
+        if let Some(lsn) = disk_slice.lsn() {
```
dentiny, line 350: (definitely not right now) we need to leave a TODO/issue to revisit on error handling, since we will have bad streaming state kept in memory forever -> OOM

src/moonlink/src/table_handler/failure_tests.rs
```diff
@@ -329,7 +355,8 @@ async fn test_force_data_compaction_with_failed_iceberg_persistence() {
     )
     .await;
     env.commit(/*lsn=*/ 20).await;
-    env.flush_table(/*lsn=*/ 30).await;
+    // env.flush_table(/*lsn=*/ 30).await;
```
dentiny: remove

src/moonlink/src/storage/mooncake_table/tests.rs
```diff
@@ -1142,38 +1219,1290 @@ async fn test_streaming_begin_flush_abort_end_flush_multiple() {
...
+    assert!(!table.ongoing_flush_lsns.contains(&50));
+    assert!(table.ongoing_flush_lsns.contains(&100));
+    assert_eq!(table.get_min_pending_flush_lsn(), 100);
+
+    // Now iceberg snapshots with flush_lsn < 100 should be allowed
+    let min_pending = table.get_min_pending_flush_lsn();
+    assert!(TableHandlerState::can_initiate_iceberg_snapshot(
+        75,
+        min_pending,
+        true,
```
dentiny, line 1616: please add comments beside, otherwise it's hard to see what field it is; and I still think it's better to use member function

src/moonlink/src/storage/mooncake_table/tests.rs
```diff
@@ -1142,38 +1219,1290 @@ async fn test_streaming_begin_flush_abort_end_flush_multiple() {
...
+
+    assert!(created, "Mooncake snapshot should be created");
+
+    sync_mooncake_snapshot(&mut table, &mut event_completion_rx).await;
+
+    Ok(())
+}
+
+#[tokio::test]
+async fn test_stream_commit_with_pending_flush_deletion_remapping_bug() -> Result<()> {
```
dentiny: why name the test case as `_bug`?

dentiny: If you think it's a bug fix, you need to link the issue around (duckdb does that)

src/moonlink/src/storage/mooncake_table/tests.rs
```diff
@@ -1142,38 +1219,1290 @@ async fn test_streaming_begin_flush_abort_end_flush_multiple() {
...
+    // Step 2: Delete some rows to create deletion vectors
+    table.delete(test_row(2, "Data2", 21), 11).await;
+    table.delete(test_row(3, "Data3", 22), 11).await;
+    table.commit(12);
+
+    // Step 3: Create and persist snapshot to create puffin deletion blobs
+    // This should persist the deletion vectors as puffin blobs
+    create_mooncake_and_persist_for_test(&mut table, &mut event_completion_rx).await;
+
+    // Step 4: Add more data to trigger compaction threshold
```
dentiny, line 2357: FYI: you should be able to tune the compaction threshold

src/moonlink/src/storage/mooncake_table/tests.rs
```diff
@@ -1142,38 +1219,1290 @@ async fn test_streaming_begin_flush_abort_end_flush_multiple() {
...
+    let created = table.create_snapshot(SnapshotOption {
+        uuid: uuid::Uuid::new_v4(),
+        force_create: true,
+        skip_iceberg_snapshot: true,
+        index_merge_option: MaintenanceOption::Skip,
+        data_compaction_option: MaintenanceOption::Skip,
+    });
+    assert!(created, "Mooncake snapshot should be created");
+
+    // This should succeed with the fix, but would fail without it
```
dentiny, line 2455: same suggestion here, if it's a bug fix, should link some reference
For example, duckdb puts issue fix tests under a certain directory: https://github.com/duckdb/duckdb/tree/main/test/issues/general

src/moonlink/src/storage/mooncake_table/tests.rs
```diff
@@ -1142,38 +1219,1290 @@ async fn test_streaming_begin_flush_abort_end_flush_multiple() {
...
+    sync_mooncake_snapshot(&mut table, &mut event_completion_rx).await;
+
+    Ok(())
+}
+
+#[tokio::test]
+async fn test_chaos_sequence_exact_repro_with_async_flush() -> Result<()> {
```
dentiny: what does the test name mean lol

src/moonlink/src/table_handler.rs
```diff
@@ -526,6 +527,23 @@ impl TableHandler {
...
+                        TableEvent::FlushResult { xact_id, flush_result } => {
+                            match flush_result {
+                                Some(Ok(disk_slice)) => {
+                                    if let Some(xact_id) = xact_id {
+                                        table.apply_stream_flush_result(xact_id, disk_slice);
+                                    } else {
+                                        table.apply_flush_result(disk_slice);
+                                    }
+                                }
+                                Some(Err(e)) => {
```
dentiny, line 539: let's unwrap it to keep the same behavior (prev sync flush unwraps directly)

My concern is:
- for now, for failed async flush, when an error happens we do nothing; no snapshot/mooncake table state update, no crash
- for chaos test it failed right away in the past, since event handler thread crashes; now it needs to wait for timeout on CI, and with no events printed out

src/moonlink/src/storage/mooncake_table.rs
```diff
@@ -754,84 +769,166 @@ impl MooncakeTable {
...
             "Current flush LSN is {:?}, new flush LSN is {}",
             self.next_snapshot_task.new_flush_lsn,
             lsn,
         );
-        self.next_snapshot_task.new_flush_lsn = Some(lsn);
 
-        if self.mem_slice.is_empty() {
+        let table_notify_tx = self.table_notify.as_ref().unwrap().clone();
+
+        if self.mem_slice.is_empty() || self.ongoing_flush_lsns.contains(&lsn) {
```
dentiny, line 890: (just a question, not request to change or update) do we have unit test for no mem slice to flush?

dentiny: I don't understand the second half, the only scenario I could think of where we have duplicate flush LSN is force snapshot, shouldn't we skip directly?

src/moonlink/src/storage/mooncake_table.rs
```diff
@@ -754,84 +769,166 @@ impl MooncakeTable {
...
+            self.set_next_flush_lsn(lsn);
+            tokio::task::spawn(async move {
```
dentiny, line 892: I don't understand why do we need to send an empty flush result back to eventloop, isn't updating flush LSN already enough? (I mean, eventloop treats it as no-op)

dentiny: That way, you don't need the `Option` wrapper

src/moonlink/src/storage/mooncake_table/table_operation_test_utils.rs
```diff
@@ -24,11 +26,124 @@ use crate::{ReadState, Result};
...
+                None
+            }
+        },
+        _ => {
+            panic!("Expected FlushResult as first event, but got others.");
+        }
+    }
+}
+
+/// Flush mooncake, block wait its completion and reflect result to mooncake table.
```
dentiny: The comment looks suspicious, I don't see the apply result part

src/moonlink/src/storage/mooncake_table/table_operation_test_utils.rs
```diff
@@ -24,11 +26,124 @@ use crate::{ReadState, Result};
...
+                None
+            }
+        },
+        _ => {
+            panic!("Expected FlushResult as first event, but got others.");
+        }
+    }
+}
+
+/// Commit transaction stream, block wait its completion and reflect result to mooncake table.
```
dentiny, line 118: Shouldn't we `apply_stream_flush_result` to sync result to table?

src/moonlink/src/storage/mooncake_table.rs
nbiscaro: added `previous_commit_lsn_baseline` to compare against and track changes.

src/moonlink/src/storage/mooncake_table.rs
nbiscaro: changed to `try_set_next_flush_lsn`

src/moonlink/src/storage/mooncake_table.rs
nbiscaro: It makes no such assumption. If we can set a new LSN we will, otherwise it remains None which is also valid. I changed the name to `try_set...` to make this more clear.

src/moonlink/src/storage/mooncake_table/tests.rs
nbiscaro: no just from my own dev. name changed

src/moonlink/src/storage/mooncake_table/tests.rs
nbiscaro: fixed

src/moonlink/src/storage/mooncake_table/table_operation_test_utils.rs
nbiscaro: we do 

```
            Some(Ok(disk_slice)) => {
                table.apply_stream_flush_result(xact_id, disk_slice);
            }
```

src/moonlink/src/table_handler.rs
nbiscaro: I want to directly test without initializing the entire TableHandlerState object

src/moonlink/src/storage/mooncake_table.rs
nbiscaro: oops, that should be `debug!` in the handler:

```
                                None => {
                                    debug!("flush result is none");
                                }

File: src/moonlink/src/storage/mooncake_table/transaction_stream.rs (hunk 31 above)
Comment (nbiscaro) [reply src/moonlink/src/storage/mooncake_table/transaction_stream.rs:374]: that is whats happening. If we commit && there are no ongoing flushes we remove the stream state.

File: src/moonlink/src/storage/mooncake_table.rs (hunk 36 above)
Comment (nbiscaro) [reply src/moonlink/src/storage/mooncake_table.rs]: changed

File: src/moonlink/src/storage/mooncake_table.rs (hunk 49 above)
Comment (nbiscaro) [reply src/moonlink/src/storage/mooncake_table.rs:892]: I think it's a cleaner mental model to notify the table that the flush process is complete regardless whether the flush has/does not have data or was successful/unsuccessful. It allows us to verify the flush process finished all the way through.

File: src/moonlink/src/storage/mooncake_table/snapshot.rs (hunk 39 above)
Comment (nbiscaro) [reply src/moonlink/src/storage/mooncake_table/snapshot.rs]: Added comment 

```
            // remap deletions written *after* this slice’s LSN
            // We set to write_lsn - 1 to maintain consistency with deletion LSNs from the streaming case.
            // In the case a streaming transaction commits before flush, we assign disk_slice.writer_lsn = commit_lsn.
            // The corresponding deletions have lsn = commit_lsn - 1, so we need >= write_lsn - 1
            // to ensure these deletions get remapped from memory to disk locations.

src/moonlink/src/storage/mooncake_table/snapshot.rs
nbiscaro: made constant

src/moonlink/src/storage/mooncake_table.rs
nbiscaro: > (just a question, not request to change or update) do we have unit test for no mem slice to flush?

defs covered in existing tests (not sure if there is an explicit isolated test)

> I don't understand the second half, the only scenario I could think of where we have duplicate flush LSN is force snapshot, shouldn't we skip directly?

Sure can remove

nbiscaro:
> 1. We inherit commit LSN, it seems we create mooncake snapshot every 500ms?
>    https://github.com/Mooncake-Labs/moonlink/pull/1204/files#r2261138006
> 2. Duplicate flush LSN handling read weird, I don't understand the necessity for an empty flush result; same for empty mem slice
> 3. Error handling behavior is different with before, which could block chaos test forever, not a good behavior.
>    3.1 I left a comment on failure related testing and chaos injection test infra, I **hope** (not a requirement) we could do the former in this PR, and leave chaos injection to next PR
> 4. remove completed lsn and set flush lsn needs better documentation/API, it's error prune without any assertion (actually if you take my suggestion item 2, you don't need two functions). Set flush LSN needs to be renamed, personal suggestion.
> 5. Not sure why you refactor table state handler's member function into class function, that makes code verbose (have to pass 4+ arguments); more importantly, we should keep coding style consistent (i.e., can_drop_table_now is still a member function)
> 6. naming consistency, should use "ongoing" instead of "pending" everywhere
> 7. I don't understand the deletion log partition LSN logic (https://github.com/Mooncake-Labs/moonlink/pull/1204/files#r2261249891), appreciate some explanation. Thank you!

1. Good catch, changed to use `previous_commit_baseline_lsn` to compare against and determine if the snapshot is dirty or not. 
2. removed
3. Changed to panic on error
3.1 Added test for disk write failure
4. added docs
5. So we can test that logic directly without needed to init entire table handler state
6. searched and replaced
7. added comment inline


Thank you for review :)

dentiny:
followup on https://github.com/Mooncake-Labs/moonlink/pull/1204/files#r2261670281
Don't see changes

dentiny:
> followup on https://github.com/Mooncake-Labs/moonlink/pull/1204/files#r2261670281 Don't see changes

My main point is: we don't need an empty flush result, which
- saves us from extra optional wrapper
- we could merge remove_ongoing_flush_lsn + try_set_next_flush_lsn for all callers

OK I see you reply in another thread: https://github.com/Mooncake-Labs/moonlink/pull/1204#discussion_r2261611208
Would you reconsider?

src/moonlink/src/table_notify.rs
```diff
@@ -180,6 +181,12 @@ pub enum TableEvent {
     PeriodicalWalPersistenceUpdateResult {
         result: Result<WalPersistenceUpdateResult>,
     },
+    FlushResult {
+        // Transaction ID
```
dentiny, line 185: You're putting it in the wrong place

src/moonlink/src/storage/mooncake_table/transaction_stream.rs
```diff
@@ -343,38 +342,38 @@ impl MooncakeTable {
...
+            self.remove_ongoing_flush_lsn(lsn);
+            self.try_set_next_flush_lsn(lsn);
```
dentiny, line 352: For some async flush operations, we handle them twice, which is very hard to reason.
I will do some refactor when I have some time.

PR #1223, opened 2025-08-05
[feature] Add support for struct data type in ColumnArrayBuilder - primitive type fields

### LINKED ISSUE #1108: Type: support composite type
## Summary

pg doc: https://www.postgresql.org/docs/current/rowtypes.html

success criteria:
- we could create a table with column type of composite type at pg_mooncake
- we're able to insert rows into, and query it

Code pointer: https://github.com/Mooncake-Labs/moonlink/blob/main/src/moonlink_connectors/src/pg_replicate/util.rs

peterxcli:
## Summary

Add RowType::Struct to Arrow Struct conversion(currently only support primitive type fields, excluding nested struct and list field)

## Related Issues

Closes https://github.com/Mooncake-Labs/moonlink/issues/1227

## Changes

- src/moonlink/src/row/column_array_builder.rs: add RowType::Struct to Arrow Struct conversion(currently only support primitive type fields, excluding nested struct and list field)

## Checklist

- [x] Code builds correctly

### DIFF
--- a/src/moonlink/src/row/column_array_builder.rs
+++ b/src/moonlink/src/row/column_array_builder.rs
@@ -1102,4 +1102,36 @@ mod tests {
             .unwrap();
         assert!(active_column.value(0));
     }
+
+    #[test]
+    fn test_column_array_builder_empty_struct() {
+        use arrow::array::StructArray;
+        use arrow::datatypes::Fields;
+
+        // Test struct with no fields (empty struct)
+        let struct_fields: Fields = Vec::<Arc<arrow::datatypes::Field>>::new().into();
+
+        let mut builder =
+            ColumnArrayBuilder::new(&DataType::Struct(struct_fields.clone()), 3, false);
+
+        // Add empty structs
+        builder.append_value(&RowValue::Struct(vec![])).unwrap();
+        builder.append_value(&RowValue::Struct(vec![])).unwrap();
+
+        // Add null struct
+        builder.append_value(&RowValue::Null).unwrap();
+
+        let array = builder.finish(&DataType::Struct(struct_fields));
+        assert_eq!(array.len(), 3);
+
+        let struct_array = array.as_any().downcast_ref::<StructArray>().unwrap();
+
+        // Check struct has no columns
+        assert_eq!(struct_array.num_columns(), 0);
+
+        // Check validity - first two are valid, third is null
+        assert!(!struct_array.is_null(0));
+        assert!(!struct_array.is_null(1));
+        assert!(struct_array.is_null(2));
+    }
 }
--- a/src/moonlink/src/row/column_array_builder.rs
+++ b/src/moonlink/src/row/column_array_builder.rs
@@ -486,7 +486,7 @@ impl ColumnArrayBuilder {
                 (Arc::new(builder.finish()), array_helper)
             }
             ColumnArrayBuilder::Struct(helper, array_helper) => {
-                (Arc::new(helper.finish_cloned()), array_helper)
+                (Arc::new(helper.finish()), array_helper)
             }
         };
         if let Some(helper) = array_helper.as_mut() {
--- a/src/moonlink/src/row/column_array_builder.rs
+++ b/src/moonlink/src/row/column_array_builder.rs
@@ -97,24 +97,19 @@ impl ColumnArrayBuilder {
                 array_builder,
             ),
             DataType::List(inner) => ColumnArrayBuilder::new(inner.data_type(), capacity, true),
-            DataType::Struct(fields) => {
-                // Use Arrow's built-in field builder creation for better compatibility
-                ColumnArrayBuilder::Struct(
-                    StructBuilder::from_fields(fields.clone(), capacity),
-                    array_builder,
-                )
-            }
+            DataType::Struct(fields) => ColumnArrayBuilder::Struct(
+                StructBuilder::from_fields(fields.clone(), capacity),
+                array_builder,
+            ),
             _ => panic!("data type: {data_type:?}"),
         }
     }
 
-    /// Helper function to append a value to a struct field
     fn append_value_to_struct_field(
         builder: &mut StructBuilder,
         i: usize,
         value: &RowValue,
     ) -> Result<(), Error> {
-        // Handle all primitive types supported in structs
         if let Some(bool_builder) = builder.field_builder::<BooleanBuilder>(i) {
             match value {
                 RowValue::Bool(v) => bool_builder.append_value(*v),
@@ -413,7 +408,6 @@ impl ColumnArrayBuilder {
             ColumnArrayBuilder::Struct(builder, _array_helper) => {
                 match value {
                     RowValue::Struct(fields) => {
-                        // Append values to each field
                         for i in 0..builder.num_fields() {
                             let field_value = fields.get(i).unwrap_or(&RowValue::Null);
                             Self::append_value_to_struct_field(builder, i, field_value)?;
@@ -421,7 +415,6 @@ impl ColumnArrayBuilder {
                         builder.append(true);
                     }
                     RowValue::Null => {
-                        // Append null to all fields
                         for i in 0..builder.num_fields() {
                             Self::append_value_to_struct_field(builder, i, &RowValue::Null)?;
                         }
--- a/src/moonlink/src/row/column_array_builder.rs
+++ b/src/moonlink/src/row/column_array_builder.rs
@@ -107,6 +107,39 @@ impl ColumnArrayBuilder {
             _ => panic!("data type: {data_type:?}"),
         }
     }
+    
+    /// Helper function to append null to a struct field builder at index i
+    fn append_null_to_struct_field(builder: &mut StructBuilder, i: usize) {
+        if let Some(int_builder) =
+            builder.field_builder::<PrimitiveBuilder<Int32Type>>(i)
+        {
+            int_builder.append_null();
+        } else if let Some(string_builder) =
+            builder.field_builder::<StringBuilder>(i)
+        {
+            string_builder.append_null();
+        } else if let Some(bool_builder) =
+            builder.field_builder::<BooleanBuilder>(i)
+        {
+            bool_builder.append_null();
+        } else if let Some(int64_builder) =
+            builder.field_builder::<PrimitiveBuilder<Int64Type>>(i)
+        {
+            int64_builder.append_null();
+        } else if let Some(float32_builder) =
+            builder.field_builder::<PrimitiveBuilder<Float32Type>>(i)
+        {
+            float32_builder.append_null();
+        } else if let Some(float64_builder) =
+            builder.field_builder::<PrimitiveBuilder<Float64Type>>(i)
+        {
+            float64_builder.append_null();
+        } else {
+            // TODO: handle nested struct and list
+            unreachable!("Unsupported field type in struct - only primitive types are supported")
+        }
+    }
+
     /// Append a value to this builder
     pub(crate) fn append_value(&mut self, value: &RowValue) -> Result<(), Error> {
         match self {
@@ -334,10 +367,8 @@ impl ColumnArrayBuilder {
             ColumnArrayBuilder::Struct(builder, _array_helper) => {
                 match value {
                     RowValue::Struct(fields) => {
-                        // Append each field value to the corresponding field builder
                         for i in 0..builder.num_fields() {
                             if i < fields.len() {
-                                // Try to append the field value based on the field type
                                 if let Some(int_builder) =
                                     builder.field_builder::<PrimitiveBuilder<Int32Type>>(i)
                                 {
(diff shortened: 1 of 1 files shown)


dentiny:
Interestingly formatting and linting check fails, just curious does `cargo fmt` and precommit hooks work here?
Also are you using [devcontainer](https://github.com/Mooncake-Labs/moonlink/blob/main/CONTRIBUTING.md#environment-setup) for development?

src/moonlink/src/row/column_array_builder.rs
```diff
@@ -95,9 +97,49 @@ impl ColumnArrayBuilder {
...
+                    StructBuilder::from_fields(fields.clone(), capacity),
+                    array_builder,
+                )
+            }
             _ => panic!("data type: {data_type:?}"),
         }
     }
+    
+    /// Helper function to append null to a struct field builder at index i
+    fn append_null_to_struct_field(builder: &mut StructBuilder, i: usize) {
```
dentiny: How do we make sure primitive type handling is exhaustive here?
Eg. how do we handle `Decimal128`?

src/moonlink/src/row/column_array_builder.rs
```diff
@@ -322,6 +364,84 @@ impl ColumnArrayBuilder {
...
+                                    builder.field_builder::<PrimitiveBuilder<Float64Type>>(i)
+                                {
+                                    match &fields[i] {
+                                        RowValue::Float64(v) => float64_builder.append_value(*v),
+                                        RowValue::Null => float64_builder.append_null(),
+                                        _ => unreachable!("Float64 expected from well-typed input")
+                                    }
+                                } else {
+                                    // TODO: handle nested struct and list
+                                    unreachable!("Unsupported field type in struct")
```
dentiny: Current impl is not that maintainable --- you're basically hand-crafted all primitive types inside of structs, which doesn't work for array and nested types, we should be able to get it work with recursion.
I will take a deep look at daytime.

peterxcli:
> curious does cargo fmt and precommit hooks work here?

I didn't ran `precommit install` on my local

> Also are you using [devcontainer](https://github.com/Mooncake-Labs/moonlink/blob/main/CONTRIBUTING.md?rgh-link-date=2025-08-05T11%3A29%3A30Z#environment-setup) for development?

Ah I didn't notice there is the recommend environment setup, but I think it's fine at this time as the this change only contain the unit test level change.
But yes, I will be aware of that, thanks for the heads-up!

src/moonlink/src/row/column_array_builder.rs
peterxcli: Write a unit test for this function? 
or we should find a way to unify the logic and these type handling. I'm looking at this way.

peterxcli: yeah and looks like I missed the Decimal, ByteArray and FixedLenByteArray handling. Thanks for pointing out.

peterxcli: I just refactored this patch, not sure if it's what you expected. PTAL!

src/moonlink/src/row/column_array_builder.rs
```diff
@@ -322,6 +410,27 @@ impl ColumnArrayBuilder {
...
+                    RowValue::Null => {
+                        // Append null to all fields
+                        for i in 0..builder.num_fields() {
+                            Self::append_value_to_struct_field(builder, i, &RowValue::Null)?;
+                        }
+                        builder.append(false);
+                    }
+                    _ => unreachable!("Struct expected from well-typed input"),
+                };
+                Ok(())
```
zhousun: similar to array type, 
struct == NULL is not the same as struct where all fields are null.
and arrowStruct array repesent nulls using a separate nullbuffer.

src/moonlink/src/row/column_array_builder.rs
```diff
@@ -95,9 +97,95 @@ impl ColumnArrayBuilder {
...
+                    StructBuilder::from_fields(fields.clone(), capacity),
+                    array_builder,
+                )
+            }
             _ => panic!("data type: {data_type:?}"),
         }
     }
+
+    /// Helper function to append a value to a struct field
+    fn append_value_to_struct_field(
```
zhousun, line 108: can you help brainstorm how to re-use the logic? ( i know rust makes it much harder)

src/moonlink/src/row/column_array_builder.rs
peterxcli: The ` builder.append(false);` does [insert a record into its null buffer](https://github.com/apache/arrow-rs/blob/5dd34630c742f3cf78f539245a6fbfdd92dde891/arrow-array/src/builder/struct_builder.rs#L204-L209)

And if I remove the "append null to each field" step, the [StructBuilder::validate_content](
https://github.com/apache/arrow-rs/blob/5dd34630c742f3cf78f539245a6fbfdd92dde891/arrow-array/src/builder/struct_builder.rs#L257-L279) will throw an error:
```rust
thread 'row::column_array_builder::tests::test_column_array_builder_struct_list' panicked at /Users/peter.li/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/arrow-array-55.2.0/src/builder/struct_builder.rs:269:17:
StructBuilder (Schema { fields: [Field { name: "id", data_type: Int32, nullable: true, dict_id: 0, dict_is_ordered: false, metadata: {} }, Field { name: "name", data_type: Utf8, nullable: true, dict_id: 0, dict_is_ordered: false, metadata: {} }, Field { name: "active", data_type: Boolean, nullable: true, dict_id: 0, dict_is_ordered: false, metadata: {} }], metadata: {} }) and field_builder with index 0 (Int32) are of unequal lengths: (1 != 3).
```


I'm not really understand how arrow do the encoding stuff. But I guess the null is mandatory is similar to parquet's encoding idea on null? WDYT?

peterxcli:
> can you help brainstorm how to re-use the logic? ( i know rust makes it much harder)

@zhousun I tried to refactor this, here is the work: https://github.com/peterxcli/moonlink/commit/2996213a7bc0d40f813644d8f3852d90dbd905ee

Please look through it when you have time. Thanks.
(compare with main branch: https://github.com/Mooncake-Labs/moonlink/compare/main...peterxcli:moonlink:refactor/replace-column-array-builder-with-node-builder?expand=1)

src/moonlink/src/row/column_array_builder.rs
```diff
@@ -95,9 +97,90 @@ impl ColumnArrayBuilder {
...
     }
+
+    fn append_value_to_struct_field(
+        builder: &mut StructBuilder,
+        i: usize,
+        value: &RowValue,
+    ) -> Result<(), Error> {
+        if let Some(bool_builder) = builder.field_builder::<BooleanBuilder>(i) {
+            match value {
+                RowValue::Bool(v) => bool_builder.append_value(*v),
```
dentiny, line 115: You might have better knowledge than I do, could you please tell me if we need to handle array type here?

I'm checking `append_value_to_struct_field` vs `append_value` and found large code duplication, would like to know the difference.

dentiny: NVM, I see your comment at L179

src/moonlink/src/row/column_array_builder.rs
```diff
@@ -95,9 +97,90 @@ impl ColumnArrayBuilder {
...
+                RowValue::Null => binary_builder.append_null(),
+                _ => unreachable!("Binary expected from well-typed input"),
+            }
+        } else if let Some(fixed_builder) = builder.field_builder::<FixedSizeBinaryBuilder>(i) {
+            match value {
+                RowValue::FixedLenByteArray(v) => fixed_builder.append_value(v)?,
+                RowValue::Null => fixed_builder.append_null(),
+                _ => unreachable!("FixedSizeBinary expected from well-typed input"),
+            }
+        } else {
```
dentiny, line 177: (no strong preference) 
My concern for this function is exhaustiveness, essentially with if-else branch, we have no way to tell whether we've iterated all types.

I'm even in favor of 
```rs
match value {
  RowValue::Int32(..) => ...
  RowValue::Float32(..) => ...
 ...
}
```
Though it might be tough for null types.

src/moonlink/src/row/column_array_builder.rs
```diff
@@ -322,6 +405,54 @@ impl ColumnArrayBuilder {
...
+                                _ => unreachable!("Struct expected from well-typed input"),
+                            }
+                        }
+                    }
+                    RowValue::Null => {
+                        if let Some(helper) = array_helper.as_mut() {
+                            helper.push_null();
+                        } else {
+                            for i in 0..builder.num_fields() {
+                                Self::append_value_to_struct_field(builder, i, &RowValue::Null)?;
```
dentiny, line 447: I'm not sure this is the correct implementation, my understanding for adding NULL-field wise is:
suppose you have 3 primitive fields in the struct, field-1 and field-2 are not null, field-3 is null
so you need to fill in field-3-builder with null, and mark the whole struct append as valid

But here, it's clear the struct itself is NULL, I think we should simply declare it null with
```rs
builder.append(null);
```

dentiny: In one word, I think 
(1) declare struct field as null
vs
(2) declare the whole struct as null
is different, and we should implement (2) here

src/moonlink/src/row/column_array_builder.rs
```diff
@@ -95,9 +97,90 @@ impl ColumnArrayBuilder {
...
+    fn append_value_to_struct_field(
+        builder: &mut StructBuilder,
+        i: usize,
+        value: &RowValue,
+    ) -> Result<(), Error> {
+        if let Some(bool_builder) = builder.field_builder::<BooleanBuilder>(i) {
+            match value {
+                RowValue::Bool(v) => bool_builder.append_value(*v),
+                RowValue::Null => bool_builder.append_null(),
+                _ => unreachable!("Bool expected from well-typed input"),
```
dentiny, line 117: I might be too paranoid, I would add logging for failed matching values, otherwise we have no way to debug, see https://github.com/Mooncake-Labs/moonlink/pull/1256

src/moonlink/src/row/column_array_builder.rs
```diff
@@ -354,6 +485,9 @@ impl ColumnArrayBuilder {
             ColumnArrayBuilder::Binary(builder, array_helper) => {
                 (Arc::new(builder.finish()), array_helper)
             }
+            ColumnArrayBuilder::Struct(helper, array_helper) => {
+                (Arc::new(helper.finish_cloned()), array_helper)
```
dentiny: curious why do we prefer `finish_cloned` instead of `finish` here?

src/moonlink/src/row/column_array_builder.rs
dentiny: I'm completely wrong: arrow-rs validity map implementation requires users to fill in field array as well.

This is the full sanity check at construction:
https://github.com/apache/arrow-rs/blob/a4bcd6d34f0d824aacfcbd2ec4cac88cce37c99f/arrow-array/src/array/struct_array.rs#L122-L126

src/moonlink/src/row/column_array_builder.rs
peterxcli: Good catch, I just misuse the api. fixed in [`299118b` (#1223)](https://github.com/Mooncake-Labs/moonlink/pull/1223/commits/299118b5f7c7bbb50d3caeee1fdc7c901c6d0307)

peterxcli:
> do you think it worthy to add a unit test for empty struct ([apache/arrow-rs@a4bcd6d/arrow-array/src/builder/struct_builder.rs#L226](https://github.com/apache/arrow-rs/blob/a4bcd6d34f0d824aacfcbd2ec4cac88cce37c99f/arrow-array/src/builder/struct_builder.rs#L226))?

yeah, make sense. Added in [`e0cb0ba` (#1223)](https://github.com/Mooncake-Labs/moonlink/pull/1223/commits/e0cb0ba6bb4a2e779ed21b93eadd2ccb652e0aef).

src/moonlink/src/row/column_array_builder.rs
```diff
@@ -560,4 +694,444 @@ mod tests {
...
+        );
+    }
+
+    #[test]
+    fn test_column_array_builder_struct_list() {
+        use arrow::array::{ListArray, StructArray};
+
+        // Test List<Struct> type
+        let struct_fields = vec![
+            Arc::new(arrow::datatypes::Field::new("id", DataType::Int32, true)),
```
dentiny, line 983: nit
```suggestion
            Arc::new(arrow::datatypes::Field::new("id", DataType::Int32, /*nullable=*/true)),
```

src/moonlink/src/row/column_array_builder.rs
```diff
@@ -560,4 +694,444 @@ mod tests {
...
+    #[test]
+    fn test_column_array_builder_empty_struct() {
+        use arrow::array::StructArray;
+        use arrow::datatypes::Fields;
+
+        // Test struct with no fields (empty struct)
+        let struct_fields: Fields = Vec::<Arc<arrow::datatypes::Field>>::new().into();
+
+        let mut builder =
+            ColumnArrayBuilder::new(&DataType::Struct(struct_fields.clone()), 3, false);
```
dentiny, line 1115: nit:
```suggestion
            ColumnArrayBuilder::new(&DataType::Struct(struct_fields.clone()), /*capacity=*/3, /*is_list=*/false);
```

src/moonlink/src/row/column_array_builder.rs
```diff
@@ -95,9 +97,90 @@ impl ColumnArrayBuilder {
...
+                RowValue::Null => float32_builder.append_null(),
+                _ => unreachable!("Float32 expected from well-typed input"),
+            }
+        } else if let Some(float64_builder) =
+            builder.field_builder::<PrimitiveBuilder<Float64Type>>(i)
+        {
+            match value {
+                RowValue::Float64(v) => float64_builder.append_value(*v),
+                RowValue::Null => float64_builder.append_null(),
+                _ => unreachable!("Float64 expected from well-typed input"),
```
dentiny, line 147: nit: I would suggest to include value in the logging message, so if anything bad really happens we're still able to debug. example: https://github.com/Mooncake-Labs/moonlink/pull/1256

dentiny (approved):
Thanks for the effort! There're some nit comments we could follow in the next PR, so as refactor change proposal you made.
It's a no-op change in production, I'm not too worried about unimplemented feature as well. :rocket:

src/moonlink/src/row/column_array_builder.rs
```diff
@@ -95,9 +97,90 @@ impl ColumnArrayBuilder {
...
+            }
+        } else if let Some(string_builder) = builder.field_builder::<StringBuilder>(i) {
+            match value {
+                RowValue::ByteArray(v) => {
+                    string_builder.append_value(unsafe { std::str::from_utf8_unchecked(v) })
+                }
+                RowValue::Null => string_builder.append_null(),
+                _ => unreachable!("String expected from well-typed input"),
+            }
+        } else if let Some(binary_builder) = builder.field_builder::<BinaryBuilder>(i) {
```
dentiny, line 165: another nit to mention is, we better reorder the field builder, so it's of the same order as column array builder:
https://github.com/Mooncake-Labs/moonlink/blob/23e38ada464dc60c294469af5f0ea42c2a85a4e7/src/moonlink/src/row/column_array_builder.rs#L31-L41
just easier to check if we miss anything.

Issue #852, opened 2025-07-14
Refine moonlink errors


dentiny:
## Summary

I would propose to do sth similar to what iceberg-rust does:
- Use a "retryable" property
- Add a context for metadata

Whether or not to add "source" is another topic, `RUST_BACKTRACE=1` is already useful
https://github.com/apache/iceberg-rust/blob/e47b268a6d58a58b8d79263d13cded55ff93fcff/crates/iceberg/src/error.rs#L133-L143

machichima:
Hi @dentiny ,

May I try this?

dentiny:
> Hi [@dentiny](https://github.com/dentiny) ,
> 
> May I try this?

Thanks @machichima! Assigned.
I will write a short doc for this issue, see how we could split the task, thank you!

dentiny:
Related issue: https://github.com/Mooncake-Labs/moonlink/issues/1155

dentiny:
@machichima Wrote a short doc: https://docs.google.com/document/d/1xbq8Ll2kvPmT-Wx_BPHdxIzsHw_iAjNT3PjlSBA9DWM/edit?tab=t.0, let me know if you have questions; happy to discuss any time.

machichima:
I prefer method 2 from the docs. I'll create a draft PR implementing OpenDAL's [error structure](https://github.com/apache/opendal/blob/a0cd502f8aa9d51a71d4686df0779ba15228ae65/core/src/types/error.rs#L223-L233) as our starting point.

Based on the issue description, I think the field we need to include in the struct are `status`, `message`, and `source`? So probably the `ErrorStruct` struct will be like:


```rust
  pub struct ErrorStruct {
      message: String,
      status: ErrorStatus,
      source: Option<anyhow::Error>,
  }
```

And I will directly use the `ErrorStatus` from OpenDAL like below, if the status is `Temporary`, we can deduce it as "retryable"

```rust
#[derive(Clone, Copy, Debug, PartialEq, Eq)]
enum ErrorStatus {
    Permanent,
    Temporary,
    Persistent,
}
```

Then for origin `Error` enum, change all types to `ErrorStruct`

```rust
#[derive(Clone, Debug, Error)]
pub enum Error {
    #[error("{0}")]
    Arrow(ErrorStruct),
    ...
}
```

Does above looks good to you?
Thank you!

machichima:
[...]

PR #1242, opened 2025-08-06
Refine moonlink errors with new Error struct

machichima:
## Summary

Error status should contain a few key attributes for better debuggability, including error status, message, and source

## Related Issues

Part of #852

## Changes

- Add a new `ErrorStruct` for setting fields needed for errors
- Apply `ErrorStruct` to other existing error types
- 

## Checklist

- [ ] Code builds correctly

### DIFF
--- a/src/moonlink/src/error.rs
+++ b/src/moonlink/src/error.rs
@@ -1,109 +1,144 @@
 use arrow::error::ArrowError;
 use iceberg::Error as IcebergError;
 use parquet::errors::ParquetError;
+use std::fmt;
 use std::io;
 use std::result;
-use std::sync::Arc;
 use thiserror::Error;
 use tokio::sync::watch;
+
+/// Error status indicating whether an error is retryable
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+pub enum ErrorStatus {
+    Permanent,
+    Temporary,
+    Persistent,
+}
+
+/// Custom error struct for moonlink
+#[derive(Clone, Debug)]
+pub struct ErrorStruct {
+    pub message: String,
+    pub status: ErrorStatus,
+    // TODO: take out source now to deal with other field first, should add it back
+}
+
+impl fmt::Display for ErrorStruct {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write!(f, "{}", self.message)
+    }
+}
 
 /// Custom error type for moonlink
 #[derive(Clone, Debug, Error)]
 pub enum Error {
-    #[error("Arrow error: {source}")]
-    Arrow { source: Arc<ArrowError> },
+    #[error("{0}")]
+    Arrow(ErrorStruct),
 
-    #[error("IO error: {source}")]
-    Io { source: Arc<io::Error> },
+    #[error("{0}")]
+    Io(ErrorStruct),
 
-    #[error("Parquet error: {source}")]
-    Parquet { source: Arc<ParquetError> },
+    #[error("{0}")]
+    Parquet(ErrorStruct),
 
     #[error("Transaction {0} not found")]
     TransactionNotFound(u32),
 
-    #[error("Watch channel receiver error: {source}")]
-    WatchChannelRecvError {
-        #[source]
-        source: watch::error::RecvError,
-    },
+    #[error("{0}")]
+    WatchChannelRecvError(ErrorStruct),
 
     #[error("Tokio join error: {0}. This typically occurs when a spawned task fails to complete successfully. Check the task's execution or panic status for more details.")]
     TokioJoinError(String),
 
-    #[error("Iceberg error: {source}")]
-    IcebergError { source: Arc<IcebergError> },
+    #[error("{0}")]
+    IcebergError(ErrorStruct),
 
     #[error("Iceberg error: {0}")]
     IcebergMessage(String),
 
-    #[error("OpenDAL error: {source}")]
-    OpenDal { source: Arc<opendal::Error> },
+    #[error("{0}")]
+    OpenDal(ErrorStruct),
 
     #[error("UTF-8 conversion error: {0}")]
     Utf8(#[from] std::string::FromUtf8Error),
 
-    #[error("Join error: {source}")]
-    JoinError { source: Arc<tokio::task::JoinError> },
+    #[error("{0}")]
+    JoinError(ErrorStruct),
 
-    #[error("JSON serialization/deserialization error: {source}")]
-    Json { source: Arc<serde_json::Error> },
+    #[error("{0}")]
+    Json(ErrorStruct),
 }
 
 pub type Result<T> = result::Result<T, Error>;
 
+impl From<watch::error::RecvError> for Error {
+    fn from(source: watch::error::RecvError) -> Self {
+        Error::WatchChannelRecvError(ErrorStruct {
+            message: format!("Watch channel receiver error: {source}"),
+            status: ErrorStatus::Permanent,
+        })
+    }
+}
+
 impl From<ArrowError> for Error {
     fn from(source: ArrowError) -> Self {
-        Error::Arrow {
-            source: Arc::new(source),
-        }
+        Error::Arrow(ErrorStruct {
+            message: format!("Arrow error: {source}"),
+            status: ErrorStatus::Permanent,
+        })
     }
 }
 
 impl From<IcebergError> for Error {
     fn from(source: IcebergError) -> Self {
-        Error::IcebergError {
-            source: Arc::new(source),
-        }
+        Error::IcebergError(ErrorStruct {
+            message: format!("Iceberg error: {source}"),
+            status: ErrorStatus::Permanent,
+        })
     }
 }
 
 impl From<io::Error> for Error {
     fn from(source: io::Error) -> Self {
-        Error::Io {
-            source: Arc::new(source),
-        }
+        Error::Io(ErrorStruct {
+            message: format!("IO error: {source}"),
+            status: ErrorStatus::Permanent,
+        })
     }
 }
 
 impl From<opendal::Error> for Error {
     fn from(source: opendal::Error) -> Self {
-        Error::OpenDal {
-            source: Arc::new(source),
-        }
+        Error::OpenDal(ErrorStruct {
+            message: format!("OpenDAL error: {source}"),
+            status: ErrorStatus::Permanent,
+        })
     }
 }
 
 impl From<tokio::task::JoinError> for Error {
     fn from(source: tokio::task::JoinError) -> Self {
-        Error::JoinError {
-            source: Arc::new(source),
-        }
+        Error::JoinError(ErrorStruct {
+            message: format!("Join error: {source}"),
+            status: ErrorStatus::Permanent,
+        })
     }
 }
 
 impl From<ParquetError> for Error {
     fn from(source: ParquetError) -> Self {
-        Error::Parquet {
-            source: Arc::new(source),
-        }
+        Error::Parquet(ErrorStruct {
+            message: format!("Parquet error: {source}"),
+            status: ErrorStatus::Permanent,
+        })
     }
 }
 
 impl From<serde_json::Error> for Error {
     fn from(source: serde_json::Error) -> Self {
-        Error::Json {
-            source: Arc::new(source),
-        }
+        Error::Parquet(ErrorStruct {
+            message: format!("JSON serialization/deserialization error: {source}"),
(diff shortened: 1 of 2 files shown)


src/moonlink/src/error.rs
```diff
@@ -1,109 +1,144 @@
...
+#[derive(Clone, Debug)]
+pub struct ErrorStruct {
+    pub message: String,
+    pub status: ErrorStatus,
+    // TODO: take out source now to deal with other field first, should add it back
+}
+
+impl fmt::Display for ErrorStruct {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write!(f, "{}", self.message)
```
dentiny: curious 
- `ErrorStruct` is already `Debug`, why do we need another formatting function? Apologize for my poor rust knowledge :sweat_smile:
- why don't we put `status` in the formatting message? My general principle is to put much information in the error message as possible, as long as it's not too spammy.

src/moonlink/src/error.rs
```diff
@@ -1,109 +1,218 @@
...
 use std::io;
 use std::result;
-use std::sync::Arc;
 use thiserror::Error;
 use tokio::sync::watch;
 
+/// Error status indicating whether an error is retryable
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+pub enum ErrorStatus {
+    /// Permanent means without external changes, the error never changes.
```
machichima: Copied the docstring for `ErrorStatus` from OpenDAL here

dentiny (commented):
Thanks for the PR! One thing I would like to know is your plan, are we going to following opendal and iceberg-rust's error status style to use `ErrorStatus` everywhere?

src/moonlink/src/error.rs
dentiny: If we copy it from apache 2.0 licensed project, I think we could have apache2.0 license here.

machichima:
> Thanks for the PR! One thing I would like to know is your plan, are we going to following opendal and iceberg-rust's error status style to use `ErrorStatus` everywhere?

Yes! I would like to try "method 2" mentioned in the docs (quote below) in this PR

> MVP for method (2)
> - Define an error status, suggest to do what opendal does
> - Migrate all existing errors into the updated error struct

src/moonlink/src/error.rs
machichima: Here I am defining the format function for `Display`, which is used when we do `println!("{}", err);`. When we use `println!("{:?}", err);` the `Debug` is used.

While we are using `#[error("{0}")]`, we need the `Display` trait here

machichima: I've added `status` in the formatting message! Thanks

src/moonlink/src/error.rs
machichima: Or maybe I should re-write the docstring here rather than directly copy it? So that we do not need to mention the apache2.0 license here

dentiny: I'm ok with any of the following (I have no preference)
- Copy the relevant documentation from OpenDAL directly and include proper copyright attribution, staying compliant with open-source licensing.
- Rephrase or rewrite the documentation in our own words.
- Simplify the error status into two categories: persistent (non-retryable) and temporary (retryable). My main concern is that the distinction between "permanent" and "persistent" isn't immediately clear without referring to the documentation.

src/moonlink/src/error.rs
dentiny: Thank you for the explanation!

dentiny:
Feel free to re-request review, or ping me on PR/slack when it's ready to review, thank you!

src/moonlink/src/error.rs
```diff
@@ -1,109 +1,196 @@
...
+        }
+    }
+}
+
+/// Custom error struct for moonlink
+#[derive(Clone, Debug)]
+pub struct ErrorStruct {
+    pub message: String,
+    pub status: ErrorStatus,
+    // TODO: take out source now to deal with other field first, should add it back
```
dentiny: No need to be in this PR, but worth a TODO or a separate issue.
Another suggestion that I have is to provide the ability to attach source error status: https://github.com/apache/iceberg-rust/blob/299cfedb7abc3d130268c519f8afe14c0e6728c2/crates/iceberg/src/error.rs#L256-L261
otherwise we lose the structured error information

In detail, an error could be 
```rs
struct SomeError {
  value: i64,
}
```
we shouldn't loss the ability to extract `value` out.

src/moonlink/src/error.rs
machichima: No problem, I use two categories for error status now and also rewrite the docstring.

Thanks!

src/moonlink/src/error.rs
machichima: I've added the "source" field in `ErrorStruct` and a `source()` function (following how OpenDAL did [here](https://github.com/apache/opendal/blob/e3f100d58150e2650a59635b0c16199bd1d9f960/core/src/types/error.rs#L308-L312)), so that we can use following for getting the structured error value

```rust
  if let Some(source) = error_struct.source() {
      if let Some(some_error) = source.downcast_ref::<SomeError>() {
          let value = some_error.value; 
      }
  }
```

machichima:
Added `source` to `Display`, following what OpenDAL did [here](https://github.com/apache/opendal/blob/e3f100d58150e2650a59635b0c16199bd1d9f960/core/src/types/error.rs#L235-L263)

 Also add test to ensure `ErrorStruct` can be displayed correctly and we can access structured error information if `source` is provided

src/moonlink/src/error.rs
```diff
@@ -1,109 +1,280 @@
...
+    }
+}
+
 impl From<ArrowError> for Error {
     fn from(source: ArrowError) -> Self {
-        Error::Arrow {
-            source: Arc::new(source),
-        }
+        let status = match source {
+            ArrowError::MemoryError(_) | ArrowError::IoError(_, _) => ErrorStatus::Temporary,
```
dentiny: I quickly checked, memory error is used in two places:
- memory alignment error: https://github.com/apache/arrow-rs/blob/554cafa140f0e4a007aa00a411d3f2b63bc6076c/arrow-ipc/src/writer.rs#L1250, which is not retryable
- Indicates ENOMEM: https://github.com/apache/arrow-rs/blob/554cafa140f0e4a007aa00a411d3f2b63bc6076c/arrow-array/src/ffi_stream.rs#L272, which I wouldn't retry

src/moonlink/src/error.rs
```diff
@@ -1,109 +1,280 @@
...
 }
 
 impl From<IcebergError> for Error {
     fn from(source: IcebergError) -> Self {
-        Error::IcebergError {
-            source: Arc::new(source),
-        }
+        Error::IcebergError(ErrorStruct {
+            message: format!("Iceberg error: {source}"),
+            status: ErrorStatus::Permanent,
```
dentiny: - Commit conflict is a retryable error: https://github.com/apache/iceberg-rust/blob/299cfedb7abc3d130268c519f8afe14c0e6728c2/crates/iceberg/src/error.rs#L65
- (no strong preference here) `Unexpected` error could be retryable also: https://github.com/apache/iceberg-rust/blob/299cfedb7abc3d130268c519f8afe14c0e6728c2/crates/iceberg/src/error.rs#L37C5-L37C15, it's used extensively inside of moonlink to represent IO errors; eg: https://github.com/Mooncake-Labs/moonlink/blob/c91807cedf2a2e008c919c6a4784baed776449d7/src/moonlink/src/storage/iceberg/file_catalog.rs#L246

src/moonlink/src/error.rs
```diff
@@ -1,109 +1,280 @@
...
-        Error::Io {
-            source: Arc::new(source),
-        }
+        let status = match source.kind() {
+            io::ErrorKind::TimedOut
+            | io::ErrorKind::Interrupted
+            | io::ErrorKind::WouldBlock
+            | io::ErrorKind::ConnectionRefused
+            | io::ErrorKind::ConnectionAborted
+            | io::ErrorKind::ConnectionReset => ErrorStatus::Temporary,
```
dentiny: - `BrokenPipe` needs to be retryable
- (no strong preference here) `NetworkDown`, `QuotaExceeded`, `ResourceBusy`, `Uncategorized` could be retryable

dentiny: If we declare opendal `RateLimited` to be retryable, `ResourceBusy` and `QuotaExceeded` should also be retryable for consistency

dentiny (commented):
Thank you! Just some nits

src/moonlink/src/error.rs
machichima: It is not recommended to match error with `Uncategorized` error based on docstring [here](https://github.com/rust-lang/rust/blob/9b1a30e5e69e1537ef6eb6eb829eb47075206dea/library/std/src/io/error.rs#L445), and linter will point this out as unstable feature

```sh
  clipper-check............................................................Failed
  - hook id: clipper-check
  - exit code: 101

  Checking moonlink v0.0.1 (/workspaces/moonlink/src/moonlink)
      Checking moonlink_rpc v0.0.1 (/workspaces/moonlink/src/moonlink_rpc)
  error[E0658]: use of unstable library feature `io_error_uncategorized`
     --> src/moonlink/src/error.rs:180:15
      |
  180 |             | io::ErrorKind::Uncategorized => ErrorStatus::Temporary,
      |               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^

```

machichima:
@dentiny 
I fix the error status categorize issue in commit a2e8637ec096c36c625f35e54475b88212c756b1, PTAL.

Thank you!

src/moonlink/src/error.rs
dentiny: Ah yes, it's still `unstable`. Thank you!

src/moonlink/Cargo.toml
```diff
@@ -37,6 +37,7 @@ chaos-test = ["function_name"]
 
 [dependencies]
 ahash = "0.8.11"
+anyhow = "1.0.98"
```
dentiny, line 40: Oh some nitpick:
- Using `<major>.<minor>.<patch>` require a precise match, which leads to duplicate dependency more easily
  + Most of the dependencies in the moonlink uses `<major>.<minor>`
- After moonlink error improvement done, we will like do the similar thing for other crates (i.e. moonlink_backend), so likely we need anyhow for other crates also, better to place it in root cargo toml file

src/moonlink/src/error.rs
```diff
@@ -1,109 +1,293 @@
...
+}
+
+impl ErrorStruct {
+    /// Sets the source error for this error struct.
+    ///
+    /// # Panics
+    ///
+    /// Panics if the source error has already been set.
+    pub fn with_source(mut self, src: impl Into<anyhow::Error>) -> Self {
+        debug_assert!(self.source.is_none(), "the source error has been set");
```
dentiny, line 57: My personal assertion practice (happy to discuss):
- Use `assert` for sanity check and enable in production
- Use `debug_assert` for expensive check, or use `#[cfg(test)]` to enable only in test (similar to debug)

src/moonlink/Cargo.toml
machichima: Got it, let me open a follow-up PR for fixing this!

src/moonlink/src/error.rs
machichima: Make sense! Opened a quick follow-up PR for this #1292

src/moonlink/src/union_read/read_state_manager.rs
```diff
@@ -183,15 +183,9 @@ impl ReadStateManager {
         table_snapshot_rx: &mut watch::Receiver<u64>,
     ) -> Result<()> {
         if requested_lsn_val > current_replication_lsn {
-            replication_lsn_rx
-                .changed()
-                .await
-                .map_err(|e| Error::WatchChannelRecvError { source: e })?;
+            replication_lsn_rx.changed().await.map_err(Error::from)?;
```
peterxcli, line 186:
```rust
replication_lsn_rx.changed().await?;
```

machichima: Thank you! I originally thought that we need to manually do `map_err(Error::from)` for triggering the `impl From<xxxError> for Error`. But it seems like it can be triggered automatically?

peterxcli: raise one PR: https://github.com/Mooncake-Labs/moonlink/pull/1515

PR #1251, opened 2025-08-06
Wal metadata and recovery helpers

lohpaul9:
Uses a WAL metadata file instead of iceberg metadata to indicate most up to date WAL files
Also adds helpers for WAL recovery to be used in the next PR with this metadata

### DIFF
--- a/src/moonlink/src/storage/wal.rs
+++ b/src/moonlink/src/storage/wal.rs
@@ -227,21 +227,24 @@ impl WalTransactionState {
     fn is_captured_in_iceberg_snapshot(
         &self,
         iceberg_snapshot_lsn: u64,
-        iceberg_snapshot_wal_file_num: Option<u64>,
+        lowest_file_kept: Option<u64>,
     ) -> bool {
         let completion_lsn_and_file = self.get_completion_lsn_and_file();
 
         // the xact has a known completion lsn by the iceberg snapshot lsn,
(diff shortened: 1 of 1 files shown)


src/moonlink/src/table_notify.rs
```diff
@@ -69,25 +69,35 @@ pub enum TableEvent {
         xact_id: Option<u32>,
         lsn: u64,
         is_copied: bool,
+        is_recovery: bool,
```
dentiny, line 72: do you think it's better to make it a enum?
for example
```rs
enum RowWriteState {
 Recovery,
 InitialCopy,
 Normal,
}
```
My concern is we could have more and more special states, and would to reduce the boolean states

src/moonlink/src/table_notify.rs
```diff
@@ -221,4 +231,34 @@ impl TableEvent {
...
+            | TableEvent::CommitFlush { is_recovery, .. }
+            | TableEvent::StreamFlush { is_recovery, .. } => *is_recovery,
+            _ => unimplemented!(
+                "TableEvent variant not supported for is_recovery: {:?}",
+                self
+            ),
+        }
+    }
+
+    pub fn set_is_recovery(&mut self, is_recovery_to_set: bool) {
```
dentiny, line 250: is it used anywhere?

PR #1255, opened 2025-08-06
hold deletion vectors for streaming batches in snapshot task

nbiscaro:
## Summary

In the case that a streaming transaction deletes from its own mem slice, we apply the deletion [inline](https://github.com/Mooncake-Labs/moonlink/blob/1f5f369c20e0d587d535eedca379a2385f4fb47b/src/moonlink/src/storage/mooncake_table/transaction_stream.rs#L178) and then filter the batch when preparing the disk slice/commit the transaction.

This is incorrect. We are not able to filter the batches at this point since its possible we have created deletion vectors that depend on the row_ids of these batches. However, we also currently don't persist deletion vector information when adding batches to snapshot task, so if we don't filter the batches we will lose deletion information.

The correct approach is to maintain deletion vector information when adding to snapshot task in the streaming case, and leave as none for non-streaming. Then we can avoid filtering batches and causing out-of-bound errors for deletes. 

## Related Issues

## Changes

- 
- 
- 

## Checklist

- [ ] Code builds correctly

### DIFF
--- a/src/moonlink/src/storage/mooncake_table.rs
+++ b/src/moonlink/src/storage/mooncake_table.rs
@@ -206,8 +206,8 @@ pub struct SnapshotTask {
     new_disk_slices: Vec<DiskSliceWriter>,
     disk_file_lsn_map: HashMap<FileId, u64>,
     new_deletions: Vec<RawDeletionRecord>,
-    /// Pair of <batch id, record batch>.
-    new_record_batches: Vec<(u64, Arc<RecordBatch>)>,
+    /// Pair of <batch id, record batch, optional deletion vector for streaming batches>.
+    new_record_batches: Vec<(u64, Arc<RecordBatch>, Option<BatchDeletionVector>)>,
     new_rows: Option<SharedRowBufferSnapshot>,
     new_mem_indices: Vec<Arc<MemIndex>>,
     /// Assigned (non-zero) after a commit event.
@@ -727,7 +727,9 @@ impl MooncakeTable {
         let lookup_key = self.metadata.identity.get_lookup_key(&row);
         let identity_for_key = self.metadata.identity.extract_identity_for_key(&row);
         if let Some(batch) = self.mem_slice.append(lookup_key, row, identity_for_key)? {
-            self.next_snapshot_task.new_record_batches.push(batch);
+            self.next_snapshot_task
+                .new_record_batches
+                .push((batch.0, batch.1, None));
         }
         Ok(())
     }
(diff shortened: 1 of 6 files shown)


dentiny (commented):
Any chance to add a unit test when we fix bugs field_builder

src/moonlink/src/storage/mooncake_table.rs
```diff
@@ -206,8 +206,8 @@ pub struct SnapshotTask {
     new_disk_slices: Vec<DiskSliceWriter>,
     disk_file_lsn_map: HashMap<FileId, u64>,
     new_deletions: Vec<RawDeletionRecord>,
-    /// Pair of <batch id, record batch>.
-    new_record_batches: Vec<(u64, Arc<RecordBatch>)>,
+    /// Pair of <batch id, record batch, optional deletion vector for streaming batches>.
+    new_record_batches: Vec<(u64, Arc<RecordBatch>, Option<BatchDeletionVector>)>,
```
dentiny, line 210: nit: I would prefer to use a struct if there're >= 3 fields over tuples, since it provides field naming.

src/moonlink/src/storage/mooncake_table/delete_vector.rs
```diff
@@ -53,6 +53,13 @@ impl BatchDeletionVector {
     pub(crate) fn delete_row(&mut self, row_idx: usize) -> bool {
         ma::assert_gt!(self.max_rows, 0);
 
+        assert!(
```
dentiny, line 56: Use `ma::assert_lt!` so mismatches values are printed out automatically

src/moonlink/src/table_handler/tests.rs
```diff
@@ -2150,3 +2150,16 @@ async fn test_commit_flush_streaming_transaction_with_deletes() {
 
     env.shutdown().await;
 }
+
+/// Minimal test to ensure we panic when deletion records have row_ids exceeding deletion vector capacity
+#[test]
```
dentiny, line 2155: Why not put the unit test under `delete_vector.rs`??

src/moonlink/src/storage/mooncake_table/tests.rs
```diff
@@ -1141,3 +1141,39 @@ async fn test_streaming_begin_flush_abort_end_flush_multiple() {
         assert_eq!(data_file_paths.len(), 0);
     }
 }
+
+#[tokio::test]
+async fn test_streaming_commit_with_unflushed_data() {
+    let context = TestContext::new("streaming_commit_with_unflushed_data");
```
dentiny, line 1147: FYI: https://github.com/Mooncake-Labs/moonlink/blob/28a57de011ba4572f06a8146d27e9530bae6b9df/src/moonlink/src/table_handler/chaos_test.rs#L757

Issue #1263, opened 2025-08-07
iceberg: fix iceberg error propagation with source error attached


dentiny:
## Summary

https://github.com/apache/iceberg-rust/blob/299cfedb7abc3d130268c519f8afe14c0e6728c2/crates/iceberg/src/error.rs#L256-L261

PR #1265, opened 2025-08-07
[storage] Improve iceberg error propagation with source

dentiny:
## Summary

Iceberg error allows attaching source error, which should be included if possible, otherwise we loss certain information.

## Related Issues

Closes https://github.com/Mooncake-Labs/moonlink/issues/1263

## Checklist

- [ ] Code builds correctly

### DIFF
--- a/src/moonlink/src/storage/iceberg/deletion_vector.rs
+++ b/src/moonlink/src/storage/iceberg/deletion_vector.rs
@@ -58,8 +58,10 @@ impl DeletionVector {
             .map_err(|e| {
                 IcebergError::new(
                     iceberg::ErrorKind::DataInvalid,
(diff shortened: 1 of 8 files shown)


src/moonlink/src/storage/iceberg/deletion_vector.rs
```diff
@@ -58,8 +58,10 @@ impl DeletionVector {
             .map_err(|e| {
                 IcebergError::new(
                     iceberg::ErrorKind::DataInvalid,
-                    format!("Failed to deserialize DeletionVector: {e}"),
+                    "Failed to deserialize DeletionVector puffin blob".to_string(),
```
dentiny, line 61: clippy made certain change

Issue #1268, opened 2025-08-07
BUG: not all all xacts in active txn map


dentiny:
## Summary

Error message:
```sh
thread 'tokio-runtime-worker' panicked at src/moonlink/src/storage/wal.rs:895:13:
all xacts in the metadata should be in the active transactions map
stack backtrace:
```
Link: https://github.com/Mooncake-Labs/moonlink/actions/runs/16805977757/job/47598388272
commit: 779173d6f259be5572810b77f8ab109700859576

lohpaul9:
Closed by #1272

PR #1272, opened 2025-08-07
[WAL] Fix testing wal xact metadata invariants

lohpaul9:
https://github.com/Mooncake-Labs/moonlink/actions/runs/16805977757/job/47598388272

`thread 'tokio-runtime-worker' panicked at src/moonlink/src/storage/wal.rs:895:13:
all xacts in the metadata should be in the active transactions map
stack backtrace:`

### DIFF
--- a/src/moonlink/src/storage/wal.rs
+++ b/src/moonlink/src/storage/wal.rs
@@ -892,8 +892,33 @@ impl WalManager {
                 .persistent_wal_metadata
                 .active_transactions
                 .clone();
-            for (xact_id, xact_state) in xact_map_from_metadata.iter() {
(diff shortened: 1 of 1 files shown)


lohpaul9:
https://github.com/Mooncake-Labs/moonlink/actions/runs/16813243824/job/47624404729#step:3:2879

we assume that the active xacts calculated just before the persistent update operation should be a subset of active xacts right after the operation completes, and check for it as an invariant. this is not always true because a stream abort might while the operation is in process, and the active xacts right after the operation completes will clean that up.

lohpaul9:
closes #1268

Issue #1229, opened 2025-08-05
type: support nested type for composite type


dentiny:
## Summary

This ticket focuses on nested composite types.

peterxcli:
Thanks for creating these tickets.

PR #1296, opened 2025-08-08
refactor: Introduce recursive ColumnArrayBuilder and test out nested struct

peterxcli:
## Summary

This PR refactors `ColumnArrayBuilder` to support **true recursive building** of nested Arrow types (e.g., `List<Struct<...>>`, deeply nested `Struct`, etc.) and simplifies the API.

Key points:

* **Recursive design:** `List` and `Struct` now compose other `ColumnArrayBuilder`s, removing the need to hand-code every primitive per struct field.
* **Simpler constructor:** `ColumnArrayBuilder::new(data_type, capacity)` (dropped the `is_list` flag; list-ness is inferred from `DataType`).
* **By-value finishing:** Builders are finished by value; batch builders are reset via `std::mem::replace`, enabling clean reuse per batch.
* **No boxing:** Replace `Vec<Box<ColumnArrayBuilder>>` with `Vec<ColumnArrayBuilder>` to reduce indirection.
* **Null semantics fixed:** Proper validity handling for null structs and null list elements.
* **Safer UTF-8:** Use fallible `String::from_utf8` instead of `from_utf8_unchecked`.

## Related Issues

Closes https://github.com/Mooncake-Labs/moonlink/issues/1229

## Changes

* Introduce `ListBuilderHelper` and `StructBuilderHelper` to manage offsets/validity and child builders.
* Remove `ArrayBuilderHelper` and the `is_list` parameter throughout call sites.
* Implement `ColumnArrayBuilder::{List, Struct}` variants that delegate recursively to child builders.
* Switch to **by-value** `finish(self, logical_type)`; use `std::mem::replace` where builders are retained after finishing.
* Replace boxed builders with values (`Vec<ColumnArrayBuilder>`) and use `into_iter()` when consuming.
* Keep necessary casts in `finish` only for logical aliases (`Date32`, `Timestamp`, `Time64`).
* Add new test: `test_column_array_builder_nested_struct`.

## Checklist

* [x] Code builds correctly
* [x] Tests have been added or updated (including nested struct coverage)
* [x] Documentation updated if necessary (doc comments and constructor usage)
* [x] I have reviewed my own changes

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -32,7 +32,6 @@ updates:
       interval: "weekly"
       day: "sunday"
     open-pull-requests-limit: 50
-    versioning-strategy: lockfile-only
     ignore:
        # For all packages, ignore all patch updates
       - dependency-name: "*"
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -15,18 +15,11 @@ repos:
         files: '\.toml$'
         entry: cargo sort --workspace
         pass_filenames: false
-      # Split clippy command into two parts, because `--fix` suppress the "warning to error" conversion.
-      - id: clipper-fixlint
-        name: clipper-fixlint
+      - id: clipper-check-fix
+        name: clipper-check-fix
         language: system
         files: '\.rs$'
-        entry: cargo clippy --all-targets --all-features --fix --allow-dirty --allow-staged
-        pass_filenames: false
-      - id: clipper-check
-        name: clipper-check
-        language: system
-        files: '\.rs$'
-        entry: cargo clippy --all-targets --all-features -- -D warnings
+        entry: bash -c 'cargo clippy --all-targets --all-features -- -D warnings || cargo clippy --all-targets --all-features --fix --allow-dirty --allow-staged'
         pass_filenames: false
       - id: check-dependency-license
         name: check-dependency-license
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -460,8 +460,35 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "edca88bc138befd0323b20752846e6587272d3b03b0343c8ea28a6f819e6e71f"
 dependencies = [
  "async-trait",
- "axum-core",
+ "axum-core 0.4.5",
  "bytes",
+ "futures-util",
+ "http",
+ "http-body",
+ "http-body-util",
+ "itoa",
+ "matchit 0.7.3",
+ "memchr",
+ "mime",
+ "percent-encoding",
+ "pin-project-lite",
+ "rustversion",
+ "serde",
+ "sync_wrapper",
+ "tower 0.5.2",
+ "tower-layer",
+ "tower-service",
+]
+
+[[package]]
+name = "axum"
+version = "0.8.4"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "021e862c184ae977658b36c4500f7feac3221ca5da43e3f25bd04ab6c79a29b5"
+dependencies = [
+ "axum-core 0.5.2",
+ "bytes",
+ "form_urlencoded",
  "futures-util",
  "http",
  "http-body",
@@ -469,7 +496,7 @@ dependencies = [
  "hyper",
  "hyper-util",
  "itoa",
- "matchit",
+ "matchit 0.8.4",
  "memchr",
  "mime",
  "percent-encoding",
@@ -496,6 +523,25 @@ dependencies = [
  "async-trait",
  "bytes",
  "futures-util",
+ "http",
+ "http-body",
+ "http-body-util",
+ "mime",
+ "pin-project-lite",
+ "rustversion",
+ "sync_wrapper",
+ "tower-layer",
+ "tower-service",
+]
+
+[[package]]
+name = "axum-core"
+version = "0.5.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "68464cd0412f486726fb3373129ef5d2993f90c34bc2bc1c1e9943b2f4fc7ca6"
+dependencies = [
+ "bytes",
+ "futures-core",
  "http",
  "http-body",
  "http-body-util",
@@ -2524,6 +2570,12 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "0e7465ac9959cc2b1404e8e2367b43684a6d13790fe23056cc8c6c5a6b7bcb94"
 
 [[package]]
+name = "matchit"
+version = "0.8.4"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "47e1ffaa40ddd1f3ed91f717a33c8c0ee23fff369e3aa8772b9605cc1d22f4c3"
+
+[[package]]
 name = "md-5"
 version = "0.10.6"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -2771,7 +2823,7 @@ version = "0.0.1"
 dependencies = [
  "arrow-ipc",
  "arrow-schema",
- "axum",
+ "axum 0.8.4",
  "clap",
  "moonlink",
  "moonlink_backend",
@@ -2782,7 +2834,7 @@ dependencies = [
  "serde_json",
  "thiserror 2.0.12",
  "tokio",
- "tower-http 0.5.2",
+ "tower-http",
  "tracing",
 ]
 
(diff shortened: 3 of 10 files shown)


src/moonlink/src/row/column_array_builder.rs
```diff
@@ -2,579 +2,363 @@ use crate::error::Error;
...
+use arrow::compute::cast;
+use arrow::datatypes::{DataType, FieldRef};
+use arrow::error::ArrowError;
 use std::sync::Arc;
 
-pub(crate) struct ArrayBuilderHelper {
-    offset_builder: Vec<i32>,
-    null_builder: NullBufferBuilder,
+/// Helper for building list arrays recursively
+pub(crate) struct ListBuilderHelper {
```
peterxcli, line 15: Not sure if there is any better naming, maybe `RecursiveListBuilder` and `RecursiveStructBuilder`?

dentiny:
Thanks for the PR! I will review later today.

src/moonlink/src/row/column_array_builder.rs
```diff
@@ -2,579 +2,363 @@ use crate::error::Error;
...
             }
-        } else if let Some(int32_builder) = builder.field_builder::<PrimitiveBuilder<Int32Type>>(i)
-        {
-            match value {
-                RowValue::Int32(v) => int32_builder.append_value(*v),
-                RowValue::Null => int32_builder.append_null(),
-                _ => unreachable!("Int32 expected from well-typed input, but get {:?}", value),
+            DataType::Decimal128(p, s) => {
+                let b = PrimitiveBuilder::<Decimal128Type>::with_capacity(capacity)
+                    .with_precision_and_scale(*p, *s)
```
dentiny: not sure why we change the error message?

dentiny: nit: `precision` and `scale` also seems proper name

src/moonlink/src/row/column_array_builder.rs
```diff
@@ -2,579 +2,363 @@ use crate::error::Error;
...
-                _ => unreachable!(
-                    "Decimal128 expected from well-typed input, but get {:?}",
-                    value
-                ),
+            other => panic!("unsupported type in builder: {other:?}"),
+        }
+    }
+
+    /// Append a value to this builder
+    pub(crate) fn append_value(&mut self, v: &RowValue) -> Result<(), Error> {
```
dentiny: curious does this work?
```suggestion
    pub(crate) fn append_value(&mut self, v: &RowValue) -> Result<()> {
```

src/moonlink/src/row/column_array_builder.rs
```diff
@@ -2,579 +2,363 @@ use crate::error::Error;
...
+            (b @ Self::Boolean(_), RowValue::Array(_))
+            | (b @ Self::Int32(_), RowValue::Array(_))
+            | (b @ Self::Int64(_), RowValue::Array(_))
+            | (b @ Self::Float32(_), RowValue::Array(_))
+            | (b @ Self::Float64(_), RowValue::Array(_))
+            | (b @ Self::Decimal128(_), RowValue::Array(_))
+            | (b @ Self::Utf8(_), RowValue::Array(_))
+            | (b @ Self::FixedSizeBinary(_), RowValue::Array(_))
+            | (b @ Self::Binary(_), RowValue::Array(_)) => {
+                Err(ArrowError::InvalidArgumentError(format!(
```
dentiny, line 313: Let's add a unit test for error cases

src/moonlink/src/row/column_array_builder.rs
```diff
@@ -2,579 +2,363 @@ use crate::error::Error;
...
+            (b, other) => Err(ArrowError::InvalidArgumentError(format!(
+                "Type mismatch: builder={:?}, value={:?}",
+                std::mem::discriminant(b),
+                other
+            ))
+            .into()),
+        }
+    }
+
+    fn len(&self) -> usize {
```
dentiny, line 331: Let's add some documentation

src/moonlink/src/row/column_array_builder.rs
```diff
@@ -2,579 +2,363 @@ use crate::error::Error;
...
+                // push same offset -> empty list; mark null element
+                b.append_null();
+                Ok(())
             }
-            ColumnArrayBuilder::Decimal128(builder, array_helper) => {
-                (Arc::new(builder.finish()), array_helper)
+
+            // ===== structs (recursive)
+            (Self::Struct(b), RowValue::Struct(vals)) => {
+                // fill missing fields as Null if vals shorter than fields
```
dentiny: curious is there reference for the behavior? I was thinking about assertion failure if number mismatch.
Actually for your current implementation, mismatches length should fail

src/moonlink/src/row/column_array_builder.rs
```diff
@@ -2,579 +2,363 @@ use crate::error::Error;
...
 
-pub(crate) struct ArrayBuilderHelper {
-    offset_builder: Vec<i32>,
-    null_builder: NullBufferBuilder,
+/// Helper for building list arrays recursively
+pub(crate) struct ListBuilderHelper {
+    field: FieldRef,
+    offsets: Vec<i32>,
+    nulls: NullBufferBuilder,
+    inner: Box<ColumnArrayBuilder>,
```
dentiny, line 21: Nice!

src/moonlink/src/row/column_array_builder.rs
```diff
@@ -2,579 +2,363 @@ use crate::error::Error;
...
+use arrow::error::ArrowError;
 use std::sync::Arc;
 
-pub(crate) struct ArrayBuilderHelper {
-    offset_builder: Vec<i32>,
-    null_builder: NullBufferBuilder,
+/// Helper for building list arrays recursively
+pub(crate) struct ListBuilderHelper {
+    field: FieldRef,
+    offsets: Vec<i32>,
```
dentiny, line 19: nit: add a comment for `offsets`

src/moonlink/src/row/column_array_builder.rs
dentiny: Let's add a unit test for this, whatever the behavior is

src/moonlink/src/row/column_array_builder.rs
```diff
@@ -2,579 +2,363 @@ use crate::error::Error;
...
+        Self {
+            fields: children,
+            nulls: NullBufferBuilder::new(capacity),
+            len: 0,
+        }
+    }
+
+    pub fn append_values(&mut self, vals: &[RowValue]) -> Result<(), Error> {
+        for (i, (_f, child)) in self.fields.iter_mut().enumerate() {
+            let rv = vals.get(i).unwrap_or(&RowValue::Null);
```
dentiny: Please add documentation on append behavior; for example, default NULL is definitely sth worth calling out.

src/moonlink/src/row/column_array_builder.rs
```diff
@@ -2,579 +2,363 @@ use crate::error::Error;
...
-        self.null_builder.append_null();
+
+    pub fn len(&self) -> usize {
+        self.len
     }
 }
+
+/// Helper for building struct recursively
+pub(crate) struct StructBuilderHelper {
+    fields: Vec<(FieldRef, ColumnArrayBuilder)>,
```
dentiny, line 71: Nice!

src/moonlink/src/row/column_array_builder.rs
peterxcli: will revert the change. thanks for catching this

src/moonlink/src/row/column_array_builder.rs
peterxcli: yeah, thanks for the suggestion

src/moonlink/src/row/column_array_builder.rs
peterxcli: I look into arrow's codebase, it panic when the field length mismatch, so I decided to make the error to be explicit.

peterxcli:
@dentiny Thanks for the thorough review. Addressed all of you comments. Please take another look.

src/moonlink/src/row/column_array_builder.rs
```diff
@@ -1,580 +1,377 @@
...
+        self.nulls.append_null();
+        self.len += 1;
+        Ok(())
+    }
+
+    pub fn finish(mut self) -> ArrayRef {
+        let mut arrays = Vec::with_capacity(self.fields.len());
+        let mut schema_fields = Vec::with_capacity(self.fields.len());
+        for (f, b) in self.fields.drain(..) {
+            schema_fields.push(f.clone());
```
dentiny, line 121: nit: we could copy data type out first

Issue #1253, opened 2025-08-06
Type: record batch equality check for array type


dentiny:
## Summary

https://github.com/Mooncake-Labs/moonlink/blob/1f5f369c20e0d587d535eedca379a2385f4fb47b/src/moonlink/src/row/moonlink_row.rs#L117-L119

fweilun:
I'm interested in this issue. Could you assign this to me?

dentiny:
> I'm interested in this issue. Could you assign this to me?

Sure thanks!

PR #1297, opened 2025-08-08
[Type] Add record batch equality check for array type

fweilun:
## Summary

Refactor the original closure into a regular function to allow recursive calls for checking array items.

## Related Issues

Closes https://github.com/Mooncake-Labs/moonlink/issues/1253

## Changes

 - Move closure to a function
 - Implement recursive array check
 - Add test case

## Checklist

- [✓] Code builds correctly
- [✓] Documentation updated if necessary


src/moonlink/src/row/moonlink_row.rs
```diff
@@ -29,105 +29,138 @@ impl MoonlinkRow {
...
+            }
+            RowValue::Null => column.is_null(idx),
+        }
+    }
+
+    /// Check whether the `offset`-th record batch matches the current moonlink row.
+    /// The `batch` here has been projected.
+    fn equals_record_batch_at_offset_impl(&self, batch: &RecordBatch, offset: usize) -> bool {
+        if offset >= batch.num_rows() {
+            panic!("Offset is out of bounds");
```
dentiny: nit:
```suggestion
            ma::assert_ge!(offset, batch.num_rows());
```
the benefit of which is it prints out values if assertion fails

dentiny:
Thanks for the PR! I will review later today.

src/moonlink/src/row/moonlink_row.rs
```diff
@@ -29,105 +29,138 @@ impl MoonlinkRow {
...
+                    }
+                    for (i, ev) in v.iter().enumerate() {
+                        if !Self::value_matches_column(ev, &child, i) {
+                            return false;
+                        }
+                    }
+                    true
+                } else {
+                    false
                 }
```
fweilun: These are the main changes to support array comparison.
Most of the diff is just moving the existing logic into a standalone function.
Hopefully this saves you some time when reviewing.

src/moonlink/src/row/moonlink_row.rs
```diff
@@ -29,105 +29,138 @@ impl MoonlinkRow {
...
+                    .as_any()
+                    .downcast_ref::<arrow::array::FixedSizeBinaryArray>()
+                {
+                    array.value(idx) == v.as_slice()
+                } else {
+                    false
+                }
+            }
+            RowValue::Array(v) => {
+                let child_opt: Option<arrow::array::ArrayRef> = if let Some(array) =
```
dentiny, line 103: It reads to me, here we're trying to downcast `Array` into all possible detailed types.
Several questions:
- It looks like there're other array types, for example: https://github.com/apache/arrow-rs/blob/25114c5caaa9e767050b366e57c20eca54aca987/arrow-array/src/array/list_array.rs#L33-L34
- My biggest concern is, how do we make sure we exhaustively iterate all possible array types?
- What if there're new array types in the future?

fweilun: After some consideration, I believe using three explicit downcasts is the better approach.
`RowValue::Array` implies that the column is of an array-of-arrays type. Currently, there are only three corresponding types, and it is unlikely this will change in the near future.

For types such as StringArray, I don’t think they should be handled within this Array branch.

fweilun: I tried using `ArrayRef::slice` to locate the element and then comparing the resulting ArrayRefs. However, slice returns a `ListArray<xxArray>` view, while the element I need to compare is in the inner Array. That mismatch makes the approach awkward, so I think it’s not the right direction here.

dentiny: Slice won't work, it's retrieving a sub-array

dentiny: Sure, thanks for the explanation and investigation!
If that's the case, we should explicitly fail on type downcast, instead of silently returning none.

dentiny (commented):
Comment left inline, also a small nit on assertion.

dentiny:
Hi @fweilun could you please take a look at the failed unit test?
> thread 'row::moonlink_row::tests::test_equals_record_batch_at_offset_array_type' panicked at src/moonlink/src/row/moonlink_row.rs:160:9:
    assertion failed: `(left >= right)`
      left: `0`,
     right: `2`

fweilun:
> Hi @fweilun could you please take a look at the failed unit test?
> 
> > thread 'row::moonlink_row::tests::test_equals_record_batch_at_offset_array_type' panicked at src/moonlink/src/row/moonlink_row.rs:160:9:
> > assertion failed: `(left >= right)`
> > left: `0`,
> > right: `2`
> > note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace

Resolved. Thanks for your review and guidance.

dentiny:
> Resolved.

Ahhh it looks like my comment typo, sorry about that :cry:

PR #1299, opened 2025-08-08
[deployment] Implement moonlink TCP server

dentiny:
## Summary

This PR implements an optional TCP server, which allows pg_mooncake to connect to moonlink standalone deployment via TCP request and response, not only unix domain socket.
This PR is a no-op to production env, TCP server is by default disabled.

Initial test setup:
https://docs.google.com/document/d/1h1UUKSJDpsUk6BUbVP9sEAo1x8vwKIgWz9T_glY7M8M/edit?usp=sharing

- I confirmed table creation and snapshot creation works
- Haven't figured out on scan table

## Checklist

- [ ] Code builds correctly

### DIFF
--- a/src/moonlink_service/src/main.rs
+++ b/src/moonlink_service/src/main.rs
@@ -11,10 +11,16 @@ struct Cli {
     /// Port for REST API server (optional, defaults to 3030)
     #[arg(long, short = 'p')]
     rest_api_port: Option<u16>,
-
     /// Disable REST API server
     #[arg(long)]
     no_rest_api: bool,
+
+    /// Port for moonlink standalone server (optional, defaults to 3031).
+    #[arg(long, short = 'p')]
+    tcp_port: Option<u16>,
+    /// Disable standalone deployment.
+    #[arg(long)]
+    no_tcp_api: bool,
 }
 
 #[tokio::main]
@@ -28,6 +34,11 @@ pub async fn main() -> Result<()> {
         } else {
             Some(cli.rest_api_port.unwrap_or(3030))
         },
+        tcp_port: if cli.no_tcp_api {
+            None
+        } else {
+            Some(cli.tcp_port.unwrap_or(3031))
+        }
     };
 
     start_with_config(config).await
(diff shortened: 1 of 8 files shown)


src/moonlink_service/src/rpc_server.rs
```diff
@@ -37,13 +39,38 @@ pub async fn start_unix_server(
...
+        let backend = Arc::clone(&backend);
+        tokio::spawn(async move {
+            match handle_stream(backend, stream).await {
+                Err(Error::Rpc(moonlink_rpc::Error::Io(e)))
+                    if matches!(e.kind(), BrokenPipe | ConnectionReset | UnexpectedEof) => {}
+                Err(e) => panic!("{e}"),
+                Ok(()) => {}
+            }
+        });
+    }
```
dpxcc, line 63: nit: ideally we should also de-dup this code block. tokio has wrappers `UnixListenerStream` and `TcpListenerStream ` for `UnixListener` and `TcpListener` that implement `Stream` trait

src/moonlink_service/src/main.rs
```diff
@@ -11,10 +11,16 @@ struct Cli {
...
     /// Disable REST API server
     #[arg(long)]
     no_rest_api: bool,
+
+    /// Port for moonlink standalone server (optional, defaults to 3031).
+    #[arg(long)]
+    tcp_port: Option<u16>,
+    /// Disable standalone deployment.
+    #[arg(long)]
+    no_tcp_api: bool,
```
dpxcc, line 23: why do we need `no_rest_api` and `no_tcp_api`? if user doesn't specify `rest_api_port` and `tcp_port`, they will be None and means no such service

dentiny: Its temporary code, I want to make this pr noop, and will clean up after integration and testing done

src/moonlink_service/src/rpc_server.rs
dentiny: Will followup pr

Issue #1252, opened 2025-08-06
type: support composite type conversion from pg to moonlink row


dentiny:
## Summary

rough code pointer: 
- https://github.com/Mooncake-Labs/moonlink/blob/main/src/moonlink_connectors/src/pg_replicate/conversions.rs
- https://github.com/Mooncake-Labs/moonlink/blob/main/src/moonlink_connectors/src/pg_replicate/util.rs

peterxcli:
@dentiny Thanks for creating these tickets.

PR #1301, opened 2025-08-08
Support composite type conversion from pg to moonlink row

peterxcli:
## Summary

Adds support for PostgreSQL composite types (user-defined types) in the CDC pipeline, enabling conversion from PostgreSQL text format to Arrow/Parquet structures.

## Related Issues

Closes https://github.com/Mooncake-Labs/moonlink/issues/1252

## Changes

- Added Composite variant to Cell and ArrayCell enums for composite type support
- Implemented parsing for PostgreSQL composite text format (e.g., `(field1,"field2")` and
`{"(1,text)","(2,more)"})`
- Added conversion logic from `Cell::Composite` to `RowValue::Struct` for Arrow compatibility
- Support for nested composites, arrays within composites, and arrays of composites

## Checklist

- [x] Code builds correctly


src/moonlink_connectors/src/pg_replicate/conversions/text.rs
```diff
@@ -150,7 +170,17 @@ impl TextFormatConverter {
             Type::JSON_ARRAY | Type::JSONB_ARRAY => Cell::Array(ArrayCell::Json(Vec::default())),
             Type::OID => Cell::U32(u32::default()),
             Type::OID_ARRAY => Cell::Array(ArrayCell::U32(Vec::default())),
-            _ => Cell::Null,
+            _ => match typ.kind() {
```
peterxcli, line 173: IIUC, this is for filling value for unchanged toast. 
I'm not sure how our downstream will handle the unchanged toast value, so I just mock them as other types.

dentiny:
> Also I wonder if we have some e2e test framework, I think it would be better to test the real pg 2 moonlink path in this composite type scenario.

Yes, I use [pg_mooncake](https://github.com/Mooncake-Labs/pg_mooncake) devcontainer for e2e test.

dentiny:
One general preference, I would like to keep changeset small.
[SW at google](https://abseil.io/resources/swe-book/html/toc.html) recommends PR less than 200 LOC

> “Small” changes should generally be limited to about 200 lines of code.

Reference: https://abseil.io/resources/swe-book/html/ch09.html#write_small_changes

src/moonlink_connectors/src/pg_replicate/conversions/text.rs
```diff
@@ -150,7 +170,17 @@ impl TextFormatConverter {
...
+                Kind::Composite(_) => Cell::Composite(Vec::default()),
+                Kind::Array(inner_type) => {
+                    // Handle arrays of composite types
+                    match inner_type.kind() {
+                        Kind::Composite(_) => Cell::Array(ArrayCell::Composite(Vec::default())),
```
dentiny, line 181: Can internal type also by array type?

src/moonlink_connectors/src/pg_replicate/conversions/text.rs
```diff
@@ -150,7 +170,17 @@ impl TextFormatConverter {
             Type::JSON_ARRAY | Type::JSONB_ARRAY => Cell::Array(ArrayCell::Json(Vec::default())),
             Type::OID => Cell::U32(u32::default()),
             Type::OID_ARRAY => Cell::Array(ArrayCell::U32(Vec::default())),
-            _ => Cell::Null,
+            _ => match typ.kind() {
+                Kind::Composite(_) => Cell::Composite(Vec::default()),
```
dentiny, line 174: just curious (not request to change), why don't we match type kind here?

src/moonlink_connectors/src/pg_replicate/conversions/text.rs
```diff
@@ -305,7 +335,19 @@ impl TextFormatConverter {
...
-            _ => Err(FromTextError::InvalidConversion()),
+            _ => match typ.kind() {
+                Kind::Composite(fields) => TextFormatConverter::parse_composite(str, fields),
+                Kind::Array(inner_type) => {
+                    // Check if the array contains composite types
+                    match inner_type.kind() {
+                        Kind::Composite(fields) => {
+                            TextFormatConverter::parse_composite_array(str, fields)
+                        }
+                        _ => Err(FromTextError::InvalidConversion()),
```
dentiny, line 357: Can we have nested array type?

src/moonlink_connectors/src/pg_replicate/conversions/text.rs
```diff
@@ -371,6 +413,151 @@ impl TextFormatConverter {
 
         Ok(Cell::Array(m(res)))
     }
+
+    fn parse_composite(
+        str: &str,
```
dentiny: nit: I usually avoid name variable type name (i.e., I prefer `s` / `text` / etc)

src/moonlink_connectors/src/pg_replicate/conversions/text.rs
```diff
@@ -371,6 +413,151 @@ impl TextFormatConverter {
 
         Ok(Cell::Array(m(res)))
     }
+
+    fn parse_composite(
```
dentiny, line 437: Let's add a comment on the behavior.

src/moonlink_connectors/src/pg_replicate/conversions/text.rs
```diff
@@ -371,6 +413,151 @@ impl TextFormatConverter {
...
+        str: &str,
+        fields: &[tokio_postgres::types::Field],
+    ) -> Result<Cell, FromTextError> {
+        if str.len() < 2 {
```
dentiny: Do you mind referencing a doc besides?

dentiny (commented):
Thanks for the thorough testing! Did a quick first round review and left some questions.

peterxcli:
> One general preference, I would like to keep changeset small. [SW at google](https://abseil.io/resources/swe-book/html/toc.html) recommends PR less than 200 LOC
> 
> > “Small” changes should generally be limited to about 200 lines of code.
> 
> Reference: [abseil.io/resources/swe-book/html/ch09.html#write_small_changes](https://abseil.io/resources/swe-book/html/ch09.html#write_small_changes)

yeah. Thanks for the heads-up. I will aware of it.

src/moonlink_connectors/src/pg_replicate/conversions/text.rs
peterxcli: We should add support for this

src/moonlink_connectors/src/pg_replicate/conversions/text.rs
peterxcli: I'm tryng add some test for https://github.com/peterxcli/moonlink/blob/d4536a5d240f36bc42a182f5027dab53038cc1ea/src/moonlink_connectors/src/pg_replicate/clients/postgres.rs#L183-L322

to check what exactly the type it will return...

src/moonlink_connectors/src/pg_replicate/conversions/text.rs
peterxcli: https://github.com/Mooncake-Labs/moonlink/issues/1338

src/moonlink_connectors/src/pg_replicate/conversions/text.rs
peterxcli: ditto

src/moonlink_connectors/src/pg_replicate/conversions/text.rs
peterxcli:
```sql
CREATE TYPE test_point AS (x INTEGER, y INTEGER);
```

dentiny:
Feel free to ping me when ready for review. :)

peterxcli:
@dentiny I took a deep look into the codebase and found the pg client hasnt supported those complex type - https://github.com/Mooncake-Labs/moonlink/issues/1359.

Which acted as a gatekeeper, so this patch is a no-op of prod code.

and the only thing that we should make sure is to make the implementation of https://github.com/Mooncake-Labs/moonlink/issues/1359 match the type matching rule introduced in this patch, or figuring out another matching way.

peterxcli:
So if you dont have any new comments, suggestions or questions, I think this is good to go.

peterxcli:
@dentiny Thanks so much for the prompt and thorough review. Addressed all of you comments. Please take another look.

src/moonlink_connectors/src/pg_replicate/conversions/text.rs
```diff
@@ -150,7 +170,21 @@ impl TextFormatConverter {
...
+            _ => match typ.kind() {
+                Kind::Composite(_) => Cell::Composite(Vec::default()),
+                Kind::Array(inner_type) => {
+                    // Handle arrays of composite types.
+                    // Note: inner_type here refers to the element type of the array.
+                    // PostgreSQL supports multi-dimensional arrays (e.g., text[][]),
+                    // but we currently only handle arrays of composite types here.
+                    match inner_type.kind() {
+                        Kind::Composite(_) => Cell::Array(ArrayCell::Composite(Vec::default())),
+                        Kind::Array(_) => Cell::Null, // TODO: Multi-dimensional arrays not yet handled
```
dentiny, line 182: For unsupported types, you should fail instead of letting it pass silently.

src/moonlink_connectors/src/pg_replicate/conversions/text.rs
```diff
@@ -459,4 +681,354 @@ mod tests {
...
+        let composite_str = "(42,\"hello world\")";
+        let cell = TextFormatConverter::parse_composite(composite_str, &fields).unwrap();
+
+        match cell {
+            Cell::Composite(values) => {
+                assert_eq!(values.len(), 2);
+                assert!(matches!(values[0], Cell::I32(42)));
+                assert!(matches!(values[1], Cell::String(ref s) if s == "hello world"));
+            }
+            _ => panic!("expected composite cell"),
```
dentiny: nit: we could print out the cell here for debugging

src/moonlink_connectors/src/pg_replicate/util.rs
```diff
@@ -362,6 +378,16 @@ impl From<PostgresTableRow> for MoonlinkRow {
                 Cell::Array(value) => {
                     values.push(RowValue::Array(convert_array_cell(value)));
                 }
+                Cell::Composite(composite_values) => {
```
dentiny: nit:
```suggestion
                Cell::Composite(value) => {
```

src/moonlink_connectors/src/pg_replicate/conversions/text.rs
```diff
@@ -459,4 +681,354 @@ mod tests {
             _ => panic!("expected string array cell"),
         }
     }
+
```
dentiny, line 685: We miss two unit tests:
- Empty struct
- A few error cases

src/moonlink_connectors/src/pg_replicate/util.rs
```diff
@@ -304,6 +304,22 @@ fn convert_array_cell(cell: ArrayCell) -> Vec<RowValue> {
...
             .collect(),
+        ArrayCell::Composite(values) => values
+            .into_iter()
+            .map(|v| {
+                v.map(|cells| {
+                    let struct_values: Vec<RowValue> = cells
+                        .into_iter()
+                        .map(|cell| PostgresTableRow(TableRow { values: vec![cell] }).into())
+                        .map(|row: MoonlinkRow| {
+                            row.values.into_iter().next().unwrap_or(RowValue::Null)
```
dentiny, line 321: I'm not sure I understand this, could you please explain to me?
It reads to me there's either one or None values in the row

peterxcli: The basic idea of this is just want to convert the `Cell` type to `RowValue` type. And it require this complex transformation. But if you get any better idea on transforming, then I will be really appreciate!

also, added some assertion when doing the transform: https://github.com/Mooncake-Labs/moonlink/pull/1301/commits/a793418d223f2bf869688876eb35731f96d3bce3

src/moonlink_connectors/src/pg_replicate/conversions/text.rs
dentiny: Thanks!

src/moonlink_connectors/src/pg_replicate/conversions.rs
```diff
@@ -52,4 +53,5 @@ pub enum ArrayCell {
     Uuid(Vec<Option<Uuid>>),
     Json(Vec<Option<serde_json::Value>>),
     Bytes(Vec<Option<Vec<u8>>>),
+    Composite(Vec<Option<Vec<Cell>>>),
```
dentiny, line 63: nit: suggest to add some comments, the type is very complicated and not easy to understand

src/moonlink_connectors/src/pg_replicate/conversions/text.rs
```diff
@@ -305,7 +339,25 @@ impl TextFormatConverter {
...
+                    // Check if the array contains composite types.
+                    // PostgreSQL supports multi-dimensional arrays (e.g., int[][], text[][]),
+                    // but here we currently only handle arrays of composite types.
+                    match inner_type.kind() {
+                        Kind::Composite(fields) => {
+                            TextFormatConverter::parse_composite_array(str, fields)
+                        }
+                        Kind::Array(_) => {
+                            // TODO: Multi-dimensional arrays not yet implemented
+                            Err(FromTextError::InvalidConversion())
```
dentiny, line 355: Need to print out the unsupported type (included in error type), otherwise hard to debug

src/moonlink_connectors/src/pg_replicate/util.rs
```diff
@@ -304,6 +304,23 @@ fn convert_array_cell(cell: ArrayCell) -> Vec<RowValue> {
...
+                        .map(|row: MoonlinkRow| {
+                            assert_eq!(row.values.len(), 1, "Composite cell should have exactly one value, but got {} values. This indicates an unexpected conversion from PostgresTableRow to MoonlinkRow", row.values.len());
+                            row.values.into_iter().next().unwrap_or(RowValue::Null)
+                        })
+                        .collect();
+                    RowValue::Struct(struct_values)
+                })
+                .unwrap_or(RowValue::Null)
+            })
+            .collect(),
```
dentiny, line 328: Need to add unit test here: https://github.com/Mooncake-Labs/moonlink/blob/586f4ce775be03caa7341b8077d3bb9c30a06adf/src/moonlink_connectors/src/pg_replicate/util.rs#L545

src/moonlink_connectors/src/pg_replicate/util.rs
```diff
@@ -768,6 +796,130 @@ mod tests {
...
+                        assert_eq!(fields[0], RowValue::Int32(200));
+                        assert_eq!(fields[1], RowValue::ByteArray("bob".as_bytes().to_vec()));
+                        assert_eq!(fields[2], RowValue::Bool(false));
+                    }
+                    _ => panic!("Expected struct in array"),
+                }
+            }
+            _ => panic!("Expected array"),
+        }
+    }
```
dentiny, line 988: Add a unit test for nested composite type please

src/moonlink_connectors/src/pg_replicate/util.rs
dentiny: It's possible, you need to have a separate from Cell type to row value type, which handles recursion.
I won't be too picky here, let's leave a TODO and an issue so we could revisit later.

src/moonlink_connectors/src/pg_replicate/conversions/text.rs
dentiny: Not requested for the current PR (since it's not a regression), feel free to leave a TODO and an issue so we don't forget.

dentiny (commented):
Some last comments, thanks!

src/moonlink_connectors/src/pg_replicate/util.rs
peterxcli: https://github.com/Mooncake-Labs/moonlink/pull/1301/files#diff-72fb3d5bb3bcfb1881a67cbd56f6f5ee37169e5d930a4964b065a9cbd375e6abR811-R814 already has it. but I will add some comment or doc on it to make it clear

src/moonlink_connectors/src/pg_replicate/util.rs
peterxcli: There are some prod code change in https://github.com/Mooncake-Labs/moonlink/pull/1301/commits/be755ee4be5ce2d33331683cfd4dc12a08ab5088

But that's just completing the missed step.

dentiny:
chaos tests failure has nothing to do with the PR.

peterxcli:
> Thanks! For the followup items feel free to create issues and I will assign to you.

Created. Thanks for the heads-up!
- https://github.com/Mooncake-Labs/moonlink/issues/1462
- https://github.com/Mooncake-Labs/moonlink/issues/1461

PR #1306, opened 2025-08-08
[WIP] apply deletes to batch dv on remap

nbiscaro:
## Summary

## Related Issues

## Changes

- 
- 
- 

## Checklist

- [ ] Code builds correctly

### DIFF
--- a/src/moonlink/src/storage/mooncake_table/transaction_stream.rs
+++ b/src/moonlink/src/storage/mooncake_table/transaction_stream.rs
@@ -419,7 +419,16 @@ impl MooncakeTable {
 
         // Remap local in mem deletions to disk deletions
         for deletion in stream_state.local_deletions.iter_mut() {
-            disk_slice.remap_deletion_if_needed(deletion);
+            if let Some(RecordLocation::DiskFile(file_id, row_idx)) =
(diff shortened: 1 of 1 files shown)


src/moonlink/src/storage/mooncake_table/transaction_stream.rs
```diff
@@ -372,8 +372,12 @@ impl MooncakeTable {
                 disk_slice.writer_lsn = Some(commit_lsn);
             }
 
+            let should_remove = stream_state.ongoing_flush_count == 0;
```
dentiny, line 374: qq: why do we need these two lines?

nbiscaro: I need to drop the mutable reference to self before calling `self.try_set_next_flush_lsn`.

dentiny: why do we need to drop it?

nbiscaro: https://doc.rust-lang.org/book/ch04-02-references-and-borrowing.html#mutable-references

PR #1313, opened 2025-08-09
Simplified event loop

fweilun:
## Summary

Replace the loop { tokio::select! { … } } event loop with a single-source receive loop:

Before:
 ```
 loop {
       tokio::select! {
       // Process events from the queue
       Some(event) = event_receiver.recv() => {
```

After: 
```
while let Some(event) = event_receiver.recv().await {
    ...
}
```

Move the shutdown error handling to the end of the loop:
```
if let Err(e) = table.shutdown().await {
    error!(error = %e, "failed to shutdown table");
}
debug!("all event senders dropped, shutting down table handler");
```
## Checklist

- [✓] Code builds correctly

### DIFF
--- a/src/moonlink/src/table_handler.rs
+++ b/src/moonlink/src/table_handler.rs
@@ -181,389 +181,514 @@ impl TableHandler {
         };
 
         // Process events until the receiver is closed or a Shutdown event is received
-        loop {
-            tokio::select! {
-                // Process events from the queue
-                Some(event) = event_receiver.recv() => {
-                    // Record event if requested.
-                    if let Some(replay_tx) = &event_replay_tx {
-                        replay_tx.send(event.clone()).unwrap();
+        while let Some(event) = event_receiver.recv().await {
+            // Process events from the queue
+
+            // Record event if requested.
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by fweilun** [comment]
> This PR contains two commits, one of the is not unintended?

Thank you for reminding me. I'll deal with it.

File: src/moonlink/src/table_handler.rs [hunk 1]
```diff
@@ -181,389 +181,514 @@ impl TableHandler {
...
-        loop {
-            tokio::select! {
-                // Process events from the queue
-                Some(event) = event_receiver.recv() => {
-                    // Record event if requested.
-                    if let Some(replay_tx) = &event_replay_tx {
-                        replay_tx.send(event.clone()).unwrap();
-                    }
+        while let Some(event) = event_receiver.recv().await {
+            // Process events from the queue
```
dentiny: nit: no need for comment (at least it shouldn't be placed here)

dentiny:
> Thank you for reminding me. I'll deal with it.

Feel free to re-request review after comment addressed next time.

dentiny (approved):
Thanks! Just one super nit

Issue #1304, opened 2025-08-08
precommit: clippy check is taking long


dentiny:
## Summary

clippy precommit check is taking minutes long

fweilun:
Could you assign this issue to me?

dentiny:
> Could you assign this issue to me?

Thanks for taking this!

PR #1332, opened 2025-08-10
Remove clippy fix

fweilun:
## Summary

I searched for repositories such as **datafusion-ray** and **arrow-rs** and found that most of them don’t use the `--fix` option. Removing it can save us around 12 seconds during Clippy checks.

## Other Considerations

The total time for the Clippy check is about 120 seconds, with the bottleneck being dependency build time (~100 seconds).  
By removing `--all-targets --all-features`, the total time is reduced to about 43 seconds. (This is the approach used by **arrow-rs**.)  I’ll leave the decision on this change to the repository owner.

## Related Issues

Related: https://github.com/Mooncake-Labs/moonlink/issues/1304

## Changes

- Remove Clippy `--fix` from pre-commit hook

## Checklist

- [✓] Code builds correctly

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -15,13 +15,6 @@ repos:
         files: '\.toml$'
         entry: cargo sort --workspace
         pass_filenames: false
-      # Split clippy command into two parts, because `--fix` suppress the "warning to error" conversion.
-      - id: clipper-fixlint
-        name: clipper-fixlint
-        language: system
-        files: '\.rs$'
-        entry: cargo clippy --all-targets --all-features --fix --allow-dirty --allow-staged
-        pass_filenames: false
       - id: clipper-check
         name: clipper-check
         language: system
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -15,11 +15,11 @@ repos:
         files: '\.toml$'
         entry: cargo sort --workspace
         pass_filenames: false
-      - id: clipper-check
-        name: clipper-check
+      - id: clipper-check-fix
+        name: clipper-check-fix
         language: system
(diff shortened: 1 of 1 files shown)


dentiny:
> The total time for the Clippy check is about 120 seconds, with the bottleneck being dependency build time (~100 seconds).

If most of the time is spent on building dependency, I don't see how removing fix option saves runtime (by a lot) ?

fweilun:
> > The total time for the Clippy check is about 120 seconds, with the bottleneck being dependency build time (~100 seconds).
> 
> If most of the time is spent on building dependency, I don't see how removing fix option saves runtime (by a lot) ?

I removed it mainly because most projects skip --fix. It only saves a little time, so it's optional.

From my observation, the only way to significantly reduce runtime is to drop --all-targets --all-features.

dentiny:
> I removed it mainly because most projects skip --fix. It only saves a little time, so it's optional.
> 
> From my observation, the only way to significantly reduce runtime is to drop --all-targets --all-features.

I personally rely on clippy fix a lot in my workflow, one way I could think of to optimize runtime (on happy path) is to only run fix when clippy check fails, do you think it's viable?

dentiny:
> By removing --all-targets --all-features, the total time is reduced to about 43 seconds. 

We used to have issue with not linting non-default crate, and leads to compilation error, so prefer to keep it.

dentiny (approved):
Thanks!

dentiny:
Ah the PR title is incorrect -- we merge two clippy commands into one instead of removing fix.

fweilun:
> Ah the PR title is incorrect -- we merge two clippy commands into one instead of removing fix.

Ahh, my bad, let me fix it.

PR #1363, opened 2025-08-11
wal recovery

lohpaul9:
## Summary

## Related Issues

## Changes

- 
- 
- 

## Checklist

- [ ] Code builds correctly


src/moonlink/src/storage/mooncake_table/transaction_stream.rs
```diff
@@ -306,10 +306,16 @@ impl MooncakeTable {
...
-            .get_mut(&xact_id)
-            .expect("Stream state not found for xact_id: {xact_id}");
+        let stream_state = self.transaction_stream_states.get_mut(&xact_id);
+
+        let stream_state = match stream_state {
+            Some(stream_state) => stream_state,
+            None => {
+                // this is the case where a stream abort is sent by moonlink during recovery
+                // to close any unclosed transactions in the WAL that existed before crash.
+                return;
```
nbiscaro: We should have some way to differentiate between the invalid state where we attempt to abort a transaction that doesn't exist, and this new edge case. I don't think we should completely remove the assertion as this will now fail silently.

src/moonlink/src/storage/wal.rs
```diff
@@ -280,7 +280,7 @@ impl WalTransactionState {
 #[derive(Clone, PartialEq, Serialize, Deserialize)]
 pub struct PersistentWalMetadata {
     curr_file_number: u64,
-    highest_seen_lsn: u64,
+    highest_completion_lsn: u64,
```
nbiscaro, line 293: Document all of this state with comments and how its used/relevant to the WAL manager.

src/moonlink/src/storage/wal.rs
```diff
@@ -291,7 +291,7 @@ impl std::fmt::Debug for PersistentWalMetadata {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
         f.debug_struct("PersistentWalMetadata")
             .field("curr_file_number", &self.curr_file_number)
-            .field("highest_seen_lsn", &self.highest_seen_lsn)
+            .field("highest_completion_lsn", &self.highest_completion_lsn)
```
nbiscaro: nit: move this debug formatter to the bottom. Keep you public APIs and constructors at the top of the file.

src/moonlink/src/storage/wal.rs
```diff
@@ -1167,41 +1209,140 @@ impl WalManager {
...
+                            // if the xact is still open, it means postgres will replay this event because we have not yet flushed it in the WAL
+                            false
+                        }
+                        // if the xact is not in the xact map, it means it is closed before the iceberg snapshot (ie we are already not tracking it)
+                        None => {
+                            if let TableEvent::Commit { lsn, .. } = event {
+                                assert!(WalManager::event_already_captured_in_iceberg_snapshot(
+                                    *lsn,
+                                    last_iceberg_snapshot_lsn,
+                                ), "an untracked streaming xact should be captured in the iceberg snapshot, but it was not");
```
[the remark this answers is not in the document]
nbiscaro: is there a reason we are guarding `assert` with a compilation flag vs use `debug_assert!()`. Should we guard with debug flag at all?

src/moonlink/src/storage/wal.rs
```diff
@@ -1167,41 +1209,141 @@ impl WalManager {
...
-                            last_iceberg_snapshot_lsn,
-                        )
+                    // Main xact - if it is > than the iceberg snapshot lsn, it is not yet in the iceberg snapshot
+                    let already_captured_in_iceberg_snapshot =
+                        if let Some(last_iceberg_snapshot_lsn) = last_iceberg_snapshot_lsn {
+                            *lsn <= last_iceberg_snapshot_lsn
+                        } else {
+                            // if no last iceberg snapshot lsn, means there is no iceberg snapshot
+                            false
+                        };
```
nbiscaro: nit: 
```suggestion
                    let already_captured_in_iceberg_snapshot = last_iceberg_snapshot.is_some() && *lsn <= last_iceberg_snapshot.unwrap();
```

src/moonlink_backend/src/lib.rs
```diff
@@ -319,10 +319,26 @@ where
         Ok(read_state.clone())
     }
 
+    pub async fn wait_for_wal_flush(&self, database_id: D, table_id: T, lsn: u64) -> Result<()> {
```
nbiscaro, line 323: Document what this public API does.

src/moonlink_backend/tests/common.rs
```diff
@@ -168,6 +169,160 @@ pub fn ids_from_state(read_state: &ReadState) -> HashSet<i64> {
         .collect()
 }
 
+#[allow(dead_code)]
```
nbiscaro: why do we this macro?

src/moonlink_connectors/src/pg_replicate.rs
```diff
@@ -440,9 +461,11 @@ impl PostgresConnection {
     }
 
     /// Shutdown PostgreSQL replication
-    pub async fn shutdown(&mut self) -> Result<()> {
-        self.drop_publication().await?;
-        self.drop_replication_slot().await?;
+    pub async fn shutdown(&mut self, drop_publication_and_replication: bool) -> Result<()> {
```
nbiscaro:
```suggestion
    pub async fn shutdown(&mut self, drop_all: bool) -> Result<()> {
```

nbiscaro: + add docs about what the flag does/when its used.

src/moonlink_connectors/src/pg_replicate.rs
```diff
@@ -405,6 +408,24 @@ impl PostgresConnection {
             is_recovery,
         )
         .await?;
+        if is_recovery {
```
nbiscaro, line 415: I think its the case that recovery and initial copy are mutually exclusive, although that's not as clear in the code. Can we be more explicit about this/add some assertions/comments.

nbiscaro (commented):
Please add a much more detailed PR description, especially for larger changes. This makes reviewing the code much easier and important for maintaining good documentation of our changes. 

I didn't look at the testing very closely. One thing I would also suggest is that in addition to your integration/unit tests, actually try testing this from `pg_mooncake` and ensure everything behaves as expected. Try killing the service mid transaction etc. and see if you can recover. 

One other note, I notice many of these changes are around renaming. In the future, we should separate these noop changes from actual logic changes.

src/moonlink/src/storage/wal.rs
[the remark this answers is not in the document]
lohpaul9: Keeping consistent with other debug assertions in the codebase for now, although am open to change in the future
https://github.com/Mooncake-Labs/moonlink/blob/cefc41343a25e205ca05cdaefddb4f1c57c45de4/src/moonlink/src/storage/mooncake_table/snapshot.rs#L595

https://github.com/Mooncake-Labs/moonlink/blob/cefc41343a25e205ca05cdaefddb4f1c57c45de4/src/moonlink/src/storage/iceberg/iceberg_table_syncer.rs#L10

src/moonlink_backend/tests/common.rs
lohpaul9: to test de-duplication - previously we only have a set of primary-key IDs, but this is not informative for testing if there were duplicate events that were applied, which is needed to test for during recovery

nbiscaro: If these are for testing only they should be guarded by a #[cfg(test)] macro.

src/moonlink_connectors/src/pg_replicate.rs
lohpaul9: added an assert with message

src/moonlink_connectors/src/pg_replicate.rs
lohpaul9: added docs, changed flag name

src/moonlink/src/storage/mooncake_table/transaction_stream.rs
lohpaul9: Introduced a field in StreamAbort indicating when this special case happens. Put the assertion back.

src/moonlink_backend/src/lib.rs
lohpaul9: added comment

src/moonlink/src/storage/wal.rs
lohpaul9: changed

src/moonlink/src/storage/wal.rs
lohpaul9: moved

src/moonlink/src/storage/wal.rs
lohpaul9: added documentation for persistent wal metadata, and the persistent wal update struct

src/moonlink_backend/tests/common.rs
nbiscaro: discussed offline. need to guard uses of `mod common` with cfg test flag, then we can remove the dead code flags.

lohpaul9:
Tested this end-to-end with pg_mooncake as well. 
Sequence to recreate:

`cd moonlink
RUST_LOG=debug RUST_BACKTRACE="1" cargo run -p moonlink_service ~/.pgrx/data-17/pg_mooncake --no-rest-api`

`RUST_LOG=debug make dev`


```
ALTER SYSTEM SET logical_decoding_work_mem = '64kB';

SELECT pg_reload_conf();

exit;
```


`RUST_LOG=debug make dev`

```
INSERT INTO trades VALUES
  (1,  'AMD', '2024-06-05 10:00:00', 119),
  (2, 'AMZN', '2024-06-05 10:05:00', 207),
  (3, 'AAPL', '2024-06-05 10:10:00', 203),
  (4, 'AMZN', '2024-06-05 10:15:00', 210);

BEGIN;

DO $$
DECLARE
  i INT;
  symbols TEXT[] := ARRAY['AAPL', 'MSFT', 'AMZN', 'GOOG', 'META', 'TSLA', 'NVDA', 'ORCL', 'INTC', 'NFLX', 'CRM'];
  base_time TIMESTAMP := '2024-06-05 10:00:00';
BEGIN
  FOR i IN 1000..10000 LOOP
    INSERT INTO trades(id, symbol, time, price)
    VALUES (
      i + 1000,
      symbols[(random() * (array_length(symbols, 1) - 1) + 1)::INT],
      base_time + (i || ' minutes')::INTERVAL,
      round((100 + random() * 1000)::numeric, 2)::float
    );
  END LOOP;
END $$;
```

We should observe streaming transaction starting because of the unfinished xact.
Kill moonlink, and restart using the same command. 

```
COMMIT;
SELECT COUNT(*) FROM trades_iceberg;
9005
```

src/moonlink_backend/src/lib.rs
dentiny: It's only used in test
```sh
src/moonlink_backend/src/lib.rs:    pub async fn wait_for_wal_flush(&self, schema: String, table: String, lsn: u64) -> Result<()> {
src/moonlink_backend/tests/test_backend.rs:            .wait_for_wal_flush(database_id, TABLE, lsn)
src/moonlink_backend/tests/test_backend.rs:            .wait_for_wal_flush(database_id, TABLE, lsn)
src/moonlink_backend/tests/test_backend.rs:            .wait_for_wal_flush(database_id, TABLE, completed_lsn)
src/moonlink_backend/tests/test_backend.rs:            .wait_for_wal_flush(database_id, TABLE, wal_flush_lsn)
src/moonlink_backend/tests/test_backend.rs:            .wait_for_wal_flush(database_id, TABLE, lsn)
src/moonlink_backend/tests/test_backend.rs:            .wait_for_wal_flush(database_id, TABLE, lsn)
```
any reason we expose it?

dentiny: so backend APIs are more or less public APIs, which I don't think we should expose WAL level info?

src/moonlink_backend/Cargo.toml
```diff
@@ -23,6 +23,7 @@ moonlink_metadata_store = { path = "../moonlink_metadata_store" }
 more-asserts = { workspace = true }
 nix = { workspace = true }
 parquet = { workspace = true, features = ["arrow"] }
+rstest = "0.26.1"
```
dentiny, line 26: -> rstest = "0.26"

src/moonlink_backend/src/lib.rs
lohpaul9: https://github.com/Mooncake-Labs/moonlink/pull/1506

have since gated the backend API behind test-utils

PR #1364, opened 2025-08-12
feat: add location to ErrorStruct

machichima:
## Summary

Add `location` (file, row, column) to `ErrorStruct` for easier debugging

## Related Issues

Related to https://github.com/Mooncake-Labs/moonlink/issues/852#issuecomment-3167857257

## Changes

- Add `location` to `ErrorStruct`
- Enable auto capture the error location with `Location::caller()`
- Add test to ensure the location is shown in the correct format

## Example

We created two files, `main.rs` and `helper.rs`, where the error is raised in `helper.rs` and printed in `main.rs`. The location in the printed error will points to the line in `helper.rs`.

Output:

```sh
❯ cargo run
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.20s
     Running `target/debug/moonlink_examples`
=== Error Location Tracking Demo ===

Caught error:
Display: Mock error for demonstration (permanent) at src/helper.rs:7:19
Debug: Io(ErrorStruct { message: "Mock error for demonstration", status: Permanent, source: None, location: Some(Location { file: "src/helper.rs", line: 7, column: 19 }) })

```

Scripts:

- `helper.rs`

```rust
use moonlink::{Error, ErrorStatus, ErrorStruct};

/// This function creates a mock error to demonstrate location tracking
/// The error location should point to this function, not the caller
pub fn produce_error() -> Result<String, Error> {
    // This line should be captured as the error location
    Err(Error::Io(ErrorStruct::new(   // <- line 7, where the error is produced
        "Mock error for demonstration".to_string(),
        ErrorStatus::Permanent,
    )))
}
```

- `main.rs`

```rust
mod helper;

fn main() {
    println!("=== Error Location Tracking Demo ===\n");

    // Call function that will generate a mock error
    match helper::produce_error() {
        Err(err) => {
            println!("Caught error:");
            println!("Display: {}", err);
            println!("Debug: {:?}", err);
        }
        Ok(_) => println!("Unexpected success"),
    }
}
```

## Checklist

- [x] Code builds correctly

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -62,7 +62,7 @@ jobs:
         include:
           - type: feature
             name: GCS feature test
-            timeout-minutes: 5
+            timeout-minutes: 10
             cmd: |
               cargo test --lib --features=storage-gcs
           - type: chaos
@@ -117,7 +117,7 @@ jobs:
         include:
           - type: feature
             name: S3 feature test
-            timeout-minutes: 5
+            timeout-minutes: 10
             cmd: |
               cargo test --lib --features=storage-s3
           - type: chaos
@@ -187,7 +187,7 @@ jobs:
         with: { tool: nextest }
 
       - name: Run tests via llvm-cov/nextest
-        timeout-minutes: 5
+        timeout-minutes: 10
         run: |
           cargo llvm-cov \
             --locked \


machichima:
@dentiny PTAL, thank you!

Cargo.toml
```diff
@@ -53,6 +53,7 @@ paste = "1"
 postgres-replication = { git = "https://github.com/Mooncake-labs/rust-postgres.git", rev = "e6bd7d5cacc4eb7a03930b5ca3db1ef9caf0a3d5" }
 postgres-types = { git = "https://github.com/Mooncake-Labs/rust-postgres.git", rev = "e6bd7d5cacc4eb7a03930b5ca3db1ef9caf0a3d5", features = ["with-serde_json-1"] }
 rand = "0.9"
+regex = "1.11"
```
dentiny, line 60: Test dependencies should go to dev dependency section

dentiny (approved):
Thanks! Just one dependency issue

dentiny:
Also would be nice to put an example in pr description (your test error status comes from the same file, which is not very clear to tell whether source error location is correctly populated )

Cargo.toml
machichima: Just to confirm, I think we should keep the root `Cargo.toml` as is and move `regex` in `src/moonlink/Cargo.toml` to dev dependency?

dentiny: Yes

PR #1366, opened 2025-08-12
[storage] Define serializable replay table event

dentiny:
## Summary

This PR is the first part to reproducible mooncake table operations.

Context:
- We decide to record events at mooncake table layer, instead of table handler layer due to higher ROI
  + Table handler layer has much more concurrency issues to handle, it's much easier for mooncake table
- To persist and replay table events, this PR defines corresponding structs for serde

Disclaimer: this PR is a no-op change, I will followup with another one to integrate with mooncake table.
Send out a PR simply to avoid giant PR, which is hard for (self) review.

## Checklist

- [ ] Code builds correctly

### DIFF
--- a/src/moonlink/src/storage/mooncake_table.rs
+++ b/src/moonlink/src/storage/mooncake_table.rs
@@ -5,6 +5,7 @@ mod disk_slice;
 mod iceberg_persisted_records;
 mod mem_slice;
 mod persistence_buffer;
+pub(crate) mod replay;
 mod shared_array;
 mod snapshot;
 mod snapshot_cache_utils;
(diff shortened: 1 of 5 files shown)


src/moonlink/src/storage/mooncake_table.rs
```diff
@@ -975,22 +877,198 @@ impl MooncakeTable {
...
+        };
+
+        for cur_file in evicted_files_to_delete.into_iter() {
+            tokio::fs::remove_file(cur_file).await?;
+        }
+
+        Ok(())
+    }
+
+    /// =======================
```
dentiny, line 981: reorder the table events, so all below functions should have its own event serialized and recorded for replay

src/moonlink/src/storage/mooncake_table.rs
```diff
@@ -975,22 +877,198 @@ impl MooncakeTable {
...
-        true
-    }
-
     /// Notify mooncake snapshot as completed.
     pub fn mark_mooncake_snapshot_completed(&mut self) {
         assert!(self.mooncake_snapshot_ongoing);
         self.mooncake_snapshot_ongoing = false;
     }
 
+    /// Update table schema to the provided [`updated_table_metadata`].
```
dentiny, line 886: no op change, just reorder

Issue #1365, opened 2025-08-12
chaos: add options to dump snapshot state


dentiny:
## Summary

two purposes:
- we could store / log it in the chaos test
- we could do extra check in chaos test or replay system

PR #1378, opened 2025-08-12
[storage] Provide option to dump table snapshot

dentiny:
## Summary

As titled, provide an option to pass back snapshot, so I could 
- validate in chaos test / replay system
- persist in a separate file for debugging purpose

## Related Issues

Closes https://github.com/Mooncake-Labs/moonlink/issues/1365

## Checklist

- [ ] Code builds correctly


src/moonlink/src/storage/mooncake_table/snapshot.rs
```diff
@@ -601,6 +603,11 @@ impl SnapshotTableState {
             data_compaction_payload,
             file_indices_merge_payload,
             evicted_data_files_to_delete,
+            current_snapshot: if opt.dump_snapshot {
+                self.current_snapshot.clone()
+            } else {
+                None
+            },
```
machichima: I think we can use following instead for simplification?

```rust
current_snapshot: opt.dump_snapshot.then(|| self.current_snapshot.clone()),
```

dentiny: Thank you for the hint! I think it makes sense, updated.

PR #1383, opened 2025-08-12
[storage] Fix timestamp with timezone parse

dentiny:
## Summary

This PR does two things:
- Fix timestamp with timezone parsing, we need to consider DST change, which leads to multiple possibilities and non-existent timestamp
- Add comments for timezone handling override

## Checklist

- [x] Code builds correctly

### DIFF
--- a/src/moonlink_connectors/src/rest_ingest/json_converter.rs
+++ b/src/moonlink_connectors/src/rest_ingest/json_converter.rs
@@ -41,6 +41,13 @@ impl JsonToMoonlinkRowConverter {
 
     fn convert_value(field: &Field, value: &Value) -> Result<RowValue, JsonToMoonlinkRowError> {
         match field.data_type() {
+            DataType::Boolean => {
+                if let Some(b) = value.as_bool() {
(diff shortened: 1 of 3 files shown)


src/moonlink_connectors/src/rest_ingest/json_converter.rs
```diff
@@ -41,23 +41,23 @@ impl JsonToMoonlinkRowConverter {
 
     fn convert_value(field: &Field, value: &Value) -> Result<RowValue, JsonToMoonlinkRowError> {
         match field.data_type() {
-            DataType::Int32 => {
-                if let Some(i) = value.as_i64() {
-                    Ok(RowValue::Int32(i as i32))
+            DataType::Boolean => {
```
dentiny, line 44: noop change here, simply group similar types together

peterxcli:
@dentiny Thanks for catching this and also the prompt fix!

peterxcli (commented):
@dentiny Thanks for catching this and also the prompt fix!

Issue #1314, opened 2025-08-09
BUG: snapshot contains extra row


dentiny:
## Summary

https://github.com/Mooncake-Labs/moonlink/actions/runs/16846364064/job/47726461879?pr=1313
https://github.com/Mooncake-Labs/moonlink/actions/runs/16844113466/job/47720640076

PR #1406, opened 2025-08-13
Fix remap partition

nbiscaro:
## Summary

We are incorrectly assuming that `committed_deletion_log` is LSN ordered. We now have multiple producers appending to this log (streaming and non-streaming transactions) therefore we cannot assume it is by default ordered by increasing LSN.

When computing the cut point we should iterate the entire log and filter rows that don't meet the condition rather than scanning for some partition point.

## Changes

- 
- 
- 

## Checklist

- [ ] Code builds correctly


dentiny:
Could you pleas also tag on https://github.com/Mooncake-Labs/moonlink/issues/1314

nbiscaro:
Still failing so not closing the issues, but this logic is certainly incorrect as is.

PR #1408, opened 2025-08-13
[minor] Provide alias for storage config

dentiny:
## Summary

Keep json consistency with other keys.

## Checklist

- [x] Code builds correctly

### DIFF
--- a/src/moonlink_backend/src/table_config.rs
+++ b/src/moonlink_backend/src/table_config.rs
@@ -124,7 +124,7 @@ mod tests {
                 },
                 "iceberg": {
                     "storage_config": {
-                        "FileSystem": {
+                        "fs": {
                             "root_directory": "/tmp"
                         }
                     }
@@ -163,7 +163,7 @@ mod tests {
                 },
                 "iceberg": {
                     "storage_config": {
-                        "Gcs": {
+                        "gcs": {
                             "project": "gcs-proj",
                             "region": "us-west1",
                             "bucket": "moonlink",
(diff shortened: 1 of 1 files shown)


src/moonlink/src/storage/filesystem/storage_config.rs
```diff
@@ -7,6 +7,7 @@ use serde::{Deserialize, Serialize};
 #[derive(Clone, Deserialize, PartialEq, Serialize)]
 pub enum StorageConfig {
     #[cfg(feature = "storage-fs")]
+    #[serde(rename = "fs", alias = "file_system", alias = "FileSystem")]
```
dpxcc: I don't think we should allow aliases here. Having more options is more confusing to the users and harder to maintain. Same for below

dentiny: Users don't need to know aliases.

dpxcc: Then why keep it in codebase..

dentiny: Take GCS as an example, `GCS`, `Gcs`, `gcs` are all well-known and commonly used terms for google cloud storage, I think it's user-friendly to accept all of them.

dentiny: > Users don't need to know aliases.

I mean, users don't need to know how many alias we provide, they just use whatever they're familiar with.

dpxcc: Not really, the users don't know which configs there are
They will just read our doc and use the one in the doc

dentiny: I will disagree but commit.

dentiny:
Unit tests have all passed.

dentiny:
Everything apart from chaos test passes.

PR #1412, opened 2025-08-13
check lsn for flusing batch

zhousun:
## Summary

For the following sequence,
Append 1
Flush 
Flush Done
Snapshot
Delete 1
Append 1
Flush
Snapshot 

The snapshot will see two candidates when applying Delete 1.
but only the first is valid, since the second is just created, which should not be visible to the deletion yet.

To solve it,
we introduce a structure 'flushing_batch_lsn_map' to snapshot_task.
For batch that is just flushed, we record the lsn of flush,  and when applying deletion, we ignore any match if 'any deletion's LSN is before the batch LSN'

## Related Issues

## Changes

- 
- 
- 

## Checklist

- [ ] Code builds correctly


src/moonlink/src/storage/mooncake_table/transaction_stream.rs
```diff
@@ -524,6 +531,9 @@ impl MooncakeTable {
                     deletions: batch.batch.deletions.clone(),
                 },
             );
+            self.next_snapshot_task
```
dentiny: check return value please

src/moonlink/src/storage/mooncake_table.rs
```diff
@@ -717,6 +722,11 @@ impl MooncakeTable {
                     deletion_vector: None,
                 });
         }
+        for batch in batches.iter() {
+            self.next_snapshot_task
+                .flushing_batch_lsn_map
```
dentiny: same here, it helped me to spot a few issues

PR #1422, opened 2025-08-13
Refine moonlink_rpc errors with new Error struct

machichima:
## Summary

Error status should contain a few key attributes for better debuggability, including error status, message, source, and location

## Related Issues

Follow up #1242

## Changes

- Move `ErrorStruct` from `moonlink` to lightweight `moonlink_error` crate to share among other crates
- Apply `ErrorStruct` to error types in `moonlink_rpc`

## Checklist

- [x] Code builds correctly


src/moonlink_rpc/Cargo.toml
```diff
@@ -6,6 +6,7 @@ license = "MIT"
 
 [dependencies]
 bincode = { workspace = true }
+moonlink = { path = "../moonlink" }
```
dentiny: This is definitely not ideal, rpc is designed to be lightweight and not depend on moonlink.
I think we should define own error struct within each crate.

dentiny: If you want to reuse the same code, we should define error in its own crate and shared among all moonlink_* crates.

src/moonlink_service/src/rpc_server.rs
```diff
@@ -31,7 +32,12 @@ pub async fn start_unix_server(
         tokio::spawn(async move {
             match handle_stream(backend, stream).await {
                 Err(Error::Rpc(moonlink_rpc::Error::Io(e)))
-                    if matches!(e.kind(), BrokenPipe | ConnectionReset | UnexpectedEof) => {}
+                    if error::Error::source(&e)
```
machichima, line 35: We need to get the `source` of the `ErrorStruct` and downcast to the original error type in order to use `.kind()`

src/moonlink_rpc/src/error.rs
```diff
@@ -1,13 +1,86 @@
...
+            io::ErrorKind::TimedOut
+            | io::ErrorKind::Interrupted
+            | io::ErrorKind::WouldBlock
+            | io::ErrorKind::ConnectionRefused
+            | io::ErrorKind::ConnectionAborted
+            | io::ErrorKind::ConnectionReset
+            | io::ErrorKind::BrokenPipe
+            | io::ErrorKind::NetworkDown
+            | io::ErrorKind::ResourceBusy
+            | io::ErrorKind::QuotaExceeded => ErrorStatus::Temporary,
```
machichima, line 62: The error status here follows our category in `moonlink/src/error.rs`

https://github.com/Mooncake-Labs/moonlink/blob/f9070772809316ba1df7e027d5af72e2eb15e8b0/src/moonlink/src/error.rs#L200-L209

src/moonlink/src/error.rs
```diff
@@ -31,10 +31,10 @@ impl fmt::Display for ErrorStatus {
...
 #[derive(Clone, Debug)]
 pub struct ErrorStruct {
-    message: String,
-    status: ErrorStatus,
-    source: Option<Arc<anyhow::Error>>,
-    location: Option<&'static Location<'static>>,
+    pub message: String,
+    pub status: ErrorStatus,
+    pub source: Option<Arc<anyhow::Error>>,
+    pub location: Option<&'static Location<'static>>,
```
machichima: Make those public so that we can use them in other crates

src/moonlink_rpc/Cargo.toml
machichima: Got it! I prefer having error in its own crate. I'll name it as `moonlink_error` and import from other moonlink_* crates.

dentiny: Hmm a second thought, I think my concern for that is also extra dependency.
Suppose we have certain centralized error struct, it would be a giant enum containing all possible errors including iceberg/opendal/json/postgres/sqlite/etc.

machichima: Oh, my thought is to only share the `ErrorStruct`, and keep the `Error` enum separately for each crate instead of sharing. Does this sound good to you?

machichima: Otherwise, I can surely put `ErrorStruct` in each crate separately, just wondering about the duplicate code for `ErrorStruct` (we need to modify multiple places if we want to modify `ErrorStruct`, which might not be ideal?)

dentiny: > Oh, my thought is to only share the ErrorStruct, and keep the Error enum separately for each crate instead of sharing.

Sure, sounds good to me, I cannot think of downsides.

src/moonlink_error/src/error.rs
machichima, line 1: Directly move the `ErrorStruct` defined in `moonlink` here

src/moonlink_rpc/Cargo.toml
machichima: Moved to `moonlink_error`. PTAL, thank you!

dpxcc:
Thanks for working on this! 👍
A good followup task is to propagate moonlink ErrorStruct to client https://github.com/Mooncake-Labs/moonlink/issues/1431

src/moonlink_error/Cargo.toml
```diff
@@ -0,0 +1,12 @@
+[package]
+name = "moonlink_error"
+version = { workspace = true }
+edition = { workspace = true }
+license = { workspace = true }
```
dpxcc: I think for moonlink_error, we can use MIT license similar to moonlink_rpc

dentiny:
Codebase is currently broken, you don't need to care about the failed chaos test, passing unit tests are pretty much enough.

src/moonlink_error/Cargo.toml
machichima: Fixed in https://github.com/Mooncake-Labs/moonlink/pull/1422/commits/f883d593c51a18dfb6dff5b702812d24d1857952, thank you!

dentiny:
I confirm the failed chaos tests have nothing to do with this PR.

Issue #1421, opened 2025-08-13
chaos: persist test config


dentiny:
## Summary

so we could create the same mooncake table at replay

PR #1423, opened 2025-08-13
[chaos] [1/N] Initiation event replay for append, delete, commit and flush

dentiny:
## Summary

This PR does a few things:
- persist mooncake table config, which is used to replay (i.e., whether iceberg is stored at local filesystem or object store)
- fix an issue where replay program shares the same event id assigner, it leads to problem that two mooncake tables (old table and table for replay) not getting the same id
- implement initial replay logic for several table events: append, delete, commit and flush

## Related Issues

Closes https://github.com/Mooncake-Labs/moonlink/issues/1421

## Checklist

- [x] Code builds correctly

### DIFF
--- a/src/moonlink/src/table_handler/chaos_replay.rs
+++ b/src/moonlink/src/table_handler/chaos_replay.rs
@@ -104,6 +104,7 @@ pub(crate) async fn replay() {
     // Start a background thread which continuously read from event receiver.
     tokio::spawn(async move {
         while let Some(table_event) = table_event_receiver.recv().await {
+            #[allow(clippy::single_match)]
             match table_event {
                 TableEvent::FlushResult {
                     id,


src/moonlink/src/table_handler/chaos_replay.rs
```diff
@@ -0,0 +1,200 @@
...
+                    id,
+                    xact_id,
+                    flush_result,
+                } => {
+                    let completed_flush = CompletedFlush {
+                        xact_id,
+                        flush_result,
+                    };
+                    let mut guard = completed_flush_events_clone.lock().await;
+                    assert!(guard.insert(id, completed_flush).is_none());
```
dentiny, line 119: CC @zhousun after replay program completes, it should contain all snapshot states, so you don't need to record flushed file ids in completed replay events.

dentiny:
Chaos test failures are all known issues.

Issue #1098, opened 2025-07-29
object storage: add cache for object metadata


dentiny:
## Summary

On iceberg table load, multiple stats are issued.

hrh47:
Hi @dentiny,
Can I try this issue? thanks

dentiny:
> Hi [@dentiny](https://github.com/dentiny), Can I try this issue? thanks

Sure, thanks for your interest. Assigned!

PR #1428, opened 2025-08-13
Feature/1098 add metadata cache

hrh47:
## Summary

Implement metadata cache

## Related Issues

https://github.com/Mooncake-Labs/moonlink/issues/1098

## Checklist

- [x] Code builds correctly


peterxcli (commented):
Curious that how this patch integrate with current read path? or this is a cache infra for obj storage?

src/moonlink/Cargo.toml
```diff
@@ -58,6 +58,7 @@ iceberg = { workspace = true }
 itertools = { workspace = true }
 lru = { workspace = true }
 memmap2 = "0.9"
+moka = { version = "0.12.10", features = ["future"] }
```
dentiny: let's use <major>.<minor> version for dependencies, otherwise it's easy to have multi-version for one single deps

src/moonlink/src/storage/cache/metadata/base_cache.rs
```diff
@@ -0,0 +1,19 @@
+use async_trait::async_trait;
+
+#[async_trait]
+#[allow(dead_code)]
+pub trait MetadataCacheTrait<K, V>: Send + Sync
+where
+    K: Send + Sync + 'static,
```
dentiny, line 8: curious why do we need `'static`?

src/moonlink/src/storage/cache/metadata/base_cache.rs
```diff
@@ -0,0 +1,19 @@
...
+    V: Send + Sync + 'static,
+{
+    async fn get(&self, key: &K) -> Option<V>
```
dentiny, line 15: comment on return value that it's making a copy

src/moonlink/src/storage/cache/metadata/cache_config.rs
```diff
@@ -0,0 +1,15 @@
+use std::time::Duration;
+
+#[derive(Debug, Clone)]
+#[allow(dead_code)]
+pub struct MetadataCacheConfig {
+    pub max_size: u64,
```
dentiny, line 13: Doc on max size
- it's for key? value? or total?
- it's for total memory usage? entry number?

src/moonlink/src/storage/cache/metadata/cache_config.rs
```diff
@@ -0,0 +1,15 @@
...
+    pub ttl: Duration,
```
dentiny, line 21: doc on ttl
- is it for ttl after last usage?
- is it for ttl since insertion?

src/moonlink/src/storage/cache/metadata/cache_config.rs
```diff
@@ -0,0 +1,15 @@
...
+}
+
+#[allow(dead_code)]
+impl MetadataCacheConfig {
```
dentiny, line 25: might be good to provide a default config
```rs
impl Default for Config {

}
```

src/moonlink/src/storage/cache/metadata/base_cache.rs
dentiny: should key also implements hash and value implements clone?

src/moonlink/src/storage/cache/metadata/moka_cache.rs
```diff
@@ -0,0 +1,71 @@
+use std::fmt::Debug;
+
+use crate::storage::cache::metadata::{
+    base_cache::MetadataCacheTrait, cache_config::MetadataCacheConfig,
+};
+use async_trait::async_trait;
+use moka::future::Cache;
+
+pub struct MokaCache<K, V> {
```
dentiny, line 34: let's put some doc on the cache class:
- what's the eviction policy
- consistency guarantee
- (optional) feature supported, i.e. ttl, max size

src/moonlink/src/storage/cache/metadata/moka_cache.rs
```diff
@@ -0,0 +1,71 @@
...
+            .max_capacity(config.max_size)
+            .time_to_live(config.ttl)
+            .eviction_policy(moka::policy::EvictionPolicy::lru())
+            .build();
+
+        Self { cache }
+    }
+
+    #[cfg(test)]
+    pub async fn initialize_for_test(&self, entries: Vec<(K, V)>) {
```
dentiny: that's sth you could put a test util function, I would like to reduce test util functions in production code, rust allows you to split the implementation for a class into multiple files

src/moonlink/src/storage/cache/metadata/moka_cache.rs
```diff
@@ -0,0 +1,71 @@
...
+where
+    K: std::hash::Hash + Eq + Clone + Send + Sync + 'static,
+    V: Clone + Send + Sync + 'static,
+{
+    async fn get(&self, key: &K) -> Option<V> {
+        self.cache.get(key).await
+    }
+
+    async fn put(&self, key: K, value: V) {
+        self.cache.insert(key, value).await;
```
dentiny, line 86: need documentation:
- if key already exists, does ttl gets refreshes?
- if already exists, does old content gets overwritten?

src/moonlink/src/storage/cache/metadata/base_cache.rs
```diff
@@ -0,0 +1,19 @@
...
+    where
+        V: Clone;
+
+    async fn put(&self, key: K, value: V);
+
+    async fn clear(&self);
+
+    async fn evict(&self, key: &K);
```
dentiny: need doc, if the key doesn't exist, what's the behavior?
- does it crash / panic?
- does it log error?
- does it silently pass?

I would suggest to return the old value if possible

src/moonlink/src/storage/cache/metadata/moka_cache.rs
```diff
@@ -0,0 +1,71 @@
...
+    pub async fn dump_all_for_test(&self) -> Vec<(K, V)> {
+        self.cache
+            .iter()
+            .map(|(k_arc, v)| ((*k_arc).clone(), v.clone()))
+            .collect()
+    }
+
+    #[cfg(test)]
+    pub async fn force_cleanup_for_test(&self) {
+        self.cache.run_pending_tasks().await;
```
dentiny: curious why do we need it for tests? isn't `clear` already enough?

src/moonlink/src/storage/cache/metadata/moka_cache_test.rs
```diff
@@ -0,0 +1,118 @@
...
+        .initialize_for_test(vec![
+            ("key1".to_string(), "value1".to_string()),
+            ("key2".to_string(), "value2".to_string()),
+        ])
+        .await;
+
+    cache.clear().await;
+    cache.force_cleanup_for_test().await;
+
+    let all_entries = cache.dump_all_for_test().await;
```
dentiny: you can provide a `len` interface for cache

src/moonlink/src/storage/cache/metadata/test_utils.rs
```diff
@@ -0,0 +1,29 @@
...
+pub struct MokaCacheTestBuilder {
+    max_size: u64,
+    ttl: Duration,
+}
+
+impl MokaCacheTestBuilder {
+    pub fn new() -> Self {
+        Self {
+            max_size: 2,
+            ttl: Duration::from_secs(3),
```
dentiny: let's do longer TTL (i.e., 1h); I'm afraid short TTL leads to flakiness

src/moonlink/src/storage/cache/metadata/moka_cache_test.rs
```diff
@@ -0,0 +1,118 @@
...
+        ])
+        .await;
+
+    let all_entries_before_ttl = cache.dump_all_for_test().await;
+
+    assert_eq!(all_entries_before_ttl.len(), 2);
+    assert!(all_entries_before_ttl.contains(&("key1".to_string(), "value1".to_string())));
+    assert!(all_entries_before_ttl.contains(&("key2".to_string(), "value2".to_string())));
+
+    tokio::time::sleep(Duration::from_secs(4)).await;
```
dentiny: This is anti-pattern for unit test, which is supposed to be short (runtime-wise), deterministic, self-contained.
If you want to test timeout, you could either make a cache with 0 ttl, or `advance_time` API (redis supports that).

src/moonlink/src/storage/cache/metadata/moka_cache_test.rs
```diff
@@ -0,0 +1,118 @@
+use std::time::Duration;
+
+use crate::storage::cache::metadata::base_cache::MetadataCacheTrait;
+use crate::storage::cache::metadata::test_utils::MokaCacheTestBuilder;
+
+#[tokio::test]
+async fn test_it_can_get_values() {
```
dentiny:
```suggestion
async fn test_get_values() {
```
remove `it_can` for all test cases

dentiny:
> Curious that how this patch integrate with current read path? or this is a cache infra for obj storage?

Hi @peterxcli thanks for asking! I could add some context here:
- iceberg read and write involves multiple stat calls on the same object, so I propose to add metadata caching somewhere
  + example: https://github.com/Mooncake-Labs/moonlink/blob/d0d6fb02ba07c15b9a38eb0195375f633e6df52b/src/moonlink/src/storage/iceberg/file_catalog.rs#L649
- a few options available is: (1) iceberg file catalog exclusive (owned by catalog); (2) shared by all file catalog; (3) filesystem level cache, owned by [filesystem accessor](https://github.com/Mooncake-Labs/moonlink/blob/main/src/moonlink/src/storage/filesystem/accessor/filesystem_accessor.rs); (4) shared cache at filesystem layer

Some pros and cons:
- it's hard to cap resource usage for exclusive cache
- filesystem accessor could provide read-after-write consistency at best
- cache at filesystem layer could benefit all storage users, and we could provide option to skip and evict caching

So I prefer (4). I know you're a storage expert, happy to see your insights :)

src/moonlink/src/storage/cache/metadata/moka_cache_test.rs
dentiny: To be specific, I think it's ok to set ttl to 0, and check keys're not there any longer.
Another way is to register a listener (only for test): https://docs.rs/moka/latest/moka/future/struct.Cache.html#per-entry-expiration-policy

src/moonlink/src/storage/cache/metadata/moka_cache.rs
hrh47: `self.cache.run_pending_tasks().await;` is needed in tests because Moka’s cache performs some operations asynchronously and lazily.

For example:
- Expiration (TTL, idle time)
- Eviction due to size limits

These tasks are not always executed immediately when you insert or remove. Instead, Moka schedules them to be processed in the background.

src/moonlink/src/storage/cache/metadata/base_cache.rs
hrh47: added

src/moonlink/src/storage/cache/metadata/cache_config.rs
hrh47: added

src/moonlink/src/storage/cache/metadata/cache_config.rs
hrh47: added

src/moonlink/src/storage/cache/metadata/moka_cache.rs
hrh47: added

src/moonlink/src/storage/cache/metadata/moka_cache.rs
hrh47: added

src/moonlink/src/storage/cache/metadata/moka_cache_test.rs
hrh47: added

src/moonlink/src/storage/cache/metadata/test_utils.rs
hrh47: fixed

src/moonlink/src/storage/cache/metadata/moka_cache_test.rs
hrh47: fixed

src/moonlink/src/storage/cache/metadata/moka_cache_test.rs
hrh47: fixed

src/moonlink/src/storage/cache/metadata/cache_config.rs
hrh47: added

src/moonlink/src/storage/cache/metadata/base_cache.rs
hrh47: added

src/moonlink/Cargo.toml
hrh47: fixed

dentiny: Oh i actually mean `<major>.<minor>`..

Cargo.toml
```diff
@@ -42,6 +42,7 @@ iceberg = { git = "https://github.com/apache/iceberg-rust.git", rev = "7cde26a7d
 ] }
 itertools = { version = "0.14" }
 lru = { version = "0.14" }
+moka = { version = "0.12.10", features = ["future"] }
```
dentiny:
```suggestion
moka = { version = "0.12", features = ["future"] }
```

src/moonlink/src/storage/cache/metadata/base_cache.rs
dentiny: let's leave a TODO for `'static`, so we could revisit later

src/moonlink/src/storage/cache/metadata/base_cache.rs
```diff
@@ -0,0 +1,27 @@
...
+{
+    /// Retrieves a value for the given key.
+    ///
+    /// **Note:** This returns a cloned copy of the value stored in the cache.
+    /// Modifying the returned value does not affect the cached value.
+    async fn get(&self, key: &K) -> Option<V>
+    where
+        V: Clone;
+
+    async fn put(&self, key: K, value: V);
```
dentiny, line 26: Need documentation on put behavior; for example, what if 
- if key already exists, does it get overwritten? or keep there?
- should we return the old value if exists?

src/moonlink/src/storage/cache/metadata/cache_config.rs
```diff
@@ -0,0 +1,49 @@
...
+    /// from the cache, even if it has not been accessed.
+    ///
+    /// Note: This is **not** an idle expiration (i.e., it does not reset on access).
+    pub ttl: Duration,
+}
+
+#[allow(dead_code)]
+impl MetadataCacheConfig {
+    pub(crate) const DEFAULT_MAX_SIZE: u64 = 1000;
+    pub(crate) const DEFAULT_TTL_SECS: u64 = 3600; // 1 hour
```
dentiny, line 27: what about we directly using `Duration` here?

src/moonlink/src/storage/cache/metadata/moka_cache.rs
```diff
@@ -0,0 +1,123 @@
...
+use async_trait::async_trait;
+use moka::future::Cache;
+
+#[cfg(test)]
+mod test_utils;
+
+/// A wrapper around [`moka::future::Cache`] providing async cache operations.
+///
+/// # Eviction Policy
+/// Uses a **size-based eviction** policy: when the cache reaches its `max_size`,
```
dentiny, line 15: Curious I see you use `**` everywhere, is it because it renders nice somewhere?

dentiny: Do you mind we naming the "size" to "capacity"?
- It's consistent with moka's terminology
- It's not straightforward to tell whether "size" means "number" or "memory usage"

src/moonlink/src/storage/cache/metadata/moka_cache.rs
```diff
@@ -0,0 +1,123 @@
...
+{
+    /// Retrieves a value for the given key.
+    ///
+    /// **Note:** This returns a cloned copy of the value stored in the cache.
+    /// Modifying the returned value does not affect the cached value.
+    async fn get(&self, key: &K) -> Option<V> {
+        self.cache.get(key).await
+    }
+
+    /// Inserts a key-value pair into the cache.
```
dentiny, line 69: We should move the documentation to trait level

src/moonlink/src/storage/cache/metadata/moka_cache.rs
```diff
@@ -0,0 +1,123 @@
...
+    }
+}
+
+#[async_trait]
+impl<K, V> MetadataCacheTrait<K, V> for MokaCache<K, V>
+where
+    K: std::hash::Hash + Eq + Clone + Send + Sync + 'static,
+    V: Clone + Send + Sync + 'static,
+{
+    /// Retrieves a value for the given key.
```
dentiny, line 61: same here, let's move it to the interface level

src/moonlink/src/storage/cache/metadata/moka_cache.rs
```diff
@@ -0,0 +1,123 @@
...
+    ///     assert_eq!(cache.evict(&"a").await, Some(1)); // removed value
+    ///     assert_eq!(cache.evict(&"a").await, None);    // already gone, silently ignored
+    /// # }
+    /// ``
+    async fn evict(&self, key: &K) -> Option<V> {
+        self.cache.remove(key).await
+    }
+
+    async fn len(&self) -> u64 {
+        self.cache.run_pending_tasks().await;
```
dentiny, line 120: It should be O(1) cheap sync call

I understand it's providing an estimated count, but that's expected due to the timestamp-based eviction nature

src/moonlink/src/storage/cache/metadata/base_cache.rs
```diff
@@ -0,0 +1,49 @@
...
+    /// Modifying the returned value does not affect the cached value.
+    async fn get(&self, key: &K) -> Option<V>
+    where
+        V: Clone;
+
+    /// Inserts a key-value pair into the cache.
+    ///
+    /// **Behavior:**
+    /// - If the key does not exist, the key-value pair is inserted.
+    /// - If the key already exists, the old value is overwritten with the new value.
```
dentiny, line 23: Also need to document the expiration behavior: if the key-value pair gets overriden, do we recalculate the timeout or it heherits.

src/moonlink/src/storage/cache/metadata/base_cache.rs
```diff
@@ -0,0 +1,49 @@
...
+    async fn clear(&self);
+
+    /// Removes the entry for the given key.
+    ///
+    /// **Behavior:**
+    /// - If the key exists, the entry is removed and its value is returned.
+    /// - If the key does not exist, returns `None`.
+    /// - This operation does not panic or log errors for missing keys.
+    ///
+    /// **Note:** The returned value is a cloned copy of the cached value.
```
dentiny, line 42: Sounds weird to me, the value has been extracted out from the cache, so we should take the ownership of it, instead of making a copy?

src/moonlink/src/storage/cache/metadata/moka_cache.rs
```diff
@@ -0,0 +1,123 @@
+use std::fmt::Debug;
+
+use crate::storage::cache::metadata::{
+    base_cache::MetadataCacheTrait, cache_config::MetadataCacheConfig,
+};
+use async_trait::async_trait;
+use moka::future::Cache;
+
+#[cfg(test)]
```
dentiny, line 9: The general practice is to place it at the end of the file.

src/moonlink/src/storage/cache/metadata/moka_cache.rs
dentiny: This should be handled

dentiny:
This PR is overall ok, but still have some improvement space. Given it's doesn't affect production usage, I will merge as is and followup with some updates myself.

Issue #1225, opened 2025-08-05
Error propagation: use expect instead of unwrap


dentiny:
## Summary

In rust,  compared with `unwrap`, `expect` provides additional string context, which is usually better for debugging and logging purpose.

mitanshu-23:
Hi @dentiny,
I'm new to moonlink. 
Can I try this one?

dentiny:
> Hi [@dentiny](https://github.com/dentiny), I'm new to moonlink. Can I try this one?

sure, go ahead.

You don't need to do everything in one single giant PR, feel free to break it down to smaller piece.

mitanshu-23:
Okay. Thank you.

Issue #1135, opened 2025-07-30
[EPIC] Improve moonlink error handling


dpxcc:
## Description

Moonlink should always propagate errors up rather than crashing

panic/expect/etc should only be used when an error indicates a bug within Moonlink itself. In certain cases, using assert's instead is likely more appropriate

We should grep moonlink codebase and make sure all panic/expect/etc are used when violating it indicates moonlink bugs in all non-testing code

## Version

- moonlink version (if applicable):
- pg_mooncake version (if applicable):

## Expected Behavior

What did you expect to happen?

## Actual Behavior

What actually happened?

## Additional Context

Any other information (logs, screenshots, configs) that might help.

dentiny:
> panic/expect/etc should only be used when an error indicates a bug within Moonlink itself

Do you have a concrete example?

zhousun:
@dpxcc why getting rid of expect?
IMO we should get rid of unwrap for sure, but usually they would be replaced with expect

dpxcc:
The main thing is that we should grep all panic/expect/etc and make sure they are only used when violating it indicates moonlink bugs...

PR #1434, opened 2025-08-13
Fix/error handling 1225

mitanshu-23:
## Summary

Replaced **_unwrap()_** with **_expect()_** in the **_src/moonlink_** which provides additional string context, which is usually better for debugging and logging purpose.

## Related Issues

Refs #1225.

## Changes

- Replaced unwrap() calls in production code with expect() calls containing descriptive error messages .

## Checklist

- [ ] Code builds correctly

### DIFF
--- a/src/moonlink/src/row/moonlink_row.rs
+++ b/src/moonlink/src/row/moonlink_row.rs
@@ -519,16 +519,16 @@ mod tests {
         )
         .unwrap();
 
-        let tmp_dir = tempdir().expect("Failed to create temporary directory for parquet test");
+        let tmp_dir = tempdir().unwrap();
         let file_path = tmp_dir.path().join("output.parquet");
-        let file = tokio::fs::File::create(&file_path).await.expect("Failed to create file for parquet test");
+        let file = tokio::fs::File::create(&file_path).await.unwrap();
         let props = WriterProperties::builder()
             .set_compression(parquet::basic::Compression::UNCOMPRESSED)
             .build();
 
-        let mut writer = AsyncArrowWriter::try_new(file, schema, Some(props)).expect("Failed to create parquet writer");
-        writer.write(&record_batch).await.expect("Failed to write record batch to parquet");
-        writer.close().await.expect("Failed to close parquet writer");
+        let mut writer = AsyncArrowWriter::try_new(file, schema, Some(props)).unwrap();
+        writer.write(&record_batch).await.unwrap();
+        writer.close().await.unwrap();
 
         // Create moonlink row to match against, which matches the second row in the parquet file.
         let row = MoonlinkRow::new(vec![RowValue::Int32(2), RowValue::Int64(20)]);
@@ -536,7 +536,7 @@ mod tests {
         // Check cases with full row as identity property.
         assert!(
             row.equals_parquet_at_offset(
-                file_path.to_str().expect("Failed to convert file path to str"),
+                file_path.to_str().unwrap(),
                 /*offset=*/ 1,
                 &IdentityProp::FullRow
             )
@@ -544,7 +544,7 @@ mod tests {
         );
         assert!(
             !row.equals_parquet_at_offset(
-                file_path.to_str().expect("Failed to convert file path to str"),
+                file_path.to_str().unwrap(),
                 /*offset=*/ 0,
                 &IdentityProp::FullRow
             )
@@ -553,12 +553,12 @@ mod tests {
 
         let identity = IdentityProp::Keys(vec![1])
             .extract_identity_columns(row.clone())
-            .expect("Failed to extract identity columns for parquet test");
+            .unwrap();
         // Check cases with specified keys.
         assert!(
             identity
                 .equals_parquet_at_offset(
-                    file_path.to_str().expect("Failed to convert file path to str"),
+                    file_path.to_str().unwrap(),
                     /*offset=*/ 1,
                     &IdentityProp::Keys(vec![1])
                 )
@@ -567,7 +567,7 @@ mod tests {
         assert!(
             !identity
                 .equals_parquet_at_offset(
-                    file_path.to_str().expect("Failed to convert file path to str"),
+                    file_path.to_str().unwrap(),
                     /*offset=*/ 0,
                     &IdentityProp::Keys(vec![1])
                 )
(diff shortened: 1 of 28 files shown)


src/moonlink/src/storage/cache/object_storage/cache_config.rs
```diff
@@ -49,7 +49,7 @@ impl ObjectStorageCacheConfig {
                 }
             }
         }
-        std::fs::create_dir_all(DEFAULT_CACHE_DIRECTORY).unwrap();
+        std::fs::create_dir_all(DEFAULT_CACHE_DIRECTORY).expect("Failed to create cache directory");
```
dentiny, line 52: You need to include path, which is the purpose for better error propagation

src/moonlink/src/row/moonlink_row.rs
```diff
@@ -574,4 +574,4 @@ mod tests {
                 .await
         );
     }
-}
+}
```
dentiny, line 577: newliner

src/moonlink/src/union_read/table_metadata.rs
```diff
@@ -151,7 +151,7 @@ impl TableMetadata {
         for i in 0..data_files_len {
             let start = data_file_offsets[i];
             let end = data_file_offsets[i + 1];
-            let s = String::from_utf8(data[cursor + start..cursor + end].to_vec()).unwrap();
+            let s = String::from_utf8(data[cursor + start..cursor + end].to_vec()).expect("Invalid UTF-8 in data filepath data blob during decode operation");
```
dentiny, line 154: please include the string content in the error message

src/moonlink/src/union_read/table_metadata.rs
```diff
@@ -164,7 +164,7 @@ impl TableMetadata {
             let start = puffin_file_offsets[i];
             let end = puffin_file_offsets[i + 1];
             let cur_puffin_filepath =
-                String::from_utf8(data[cursor + start..cursor + end].to_vec()).unwrap();
+                String::from_utf8(data[cursor + start..cursor + end].to_vec()).expect("Invalid UTF-8 in puffin filepath data blob during decode");
```
dentiny, line 167: same here

src/moonlink/src/storage/iceberg/deletion_vector_manifest_manager.rs
```diff
@@ -151,7 +151,7 @@ fn get_data_file_for_deletion_vector(
             .get(DELETION_VECTOR_CADINALITY)
             .unwrap()
             .parse()
-            .unwrap(),
+            .expect("Failed to parse deletion vector cardinality from properties during deserialization"),
```
dentiny, line 154: please include the string in the error message

dentiny: Also for L152, let's include the map value maybe

src/moonlink/src/storage/iceberg/deletion_vector.rs
```diff
@@ -222,7 +222,7 @@ impl DeletionVector {
             .get(MOONCAKE_DELETION_VECTOR_NUM_ROWS)
             .unwrap()
             .parse()
-            .unwrap();
+            .expect("Failed to parse max number of rows from deletion vector properties during deserialization");
```
dentiny, line 225: Same here, you should print out the string format of numerical value

dentiny: Suggest to improve at L223 as well

dentiny (commented):
Thanks for the effort!
As we chatted offline,
- I don't think we care too much about testing, so no need to update testing part; but it's fine to keep it
- the benefit for `expect` over `unwrap` is to provide more context, i.e., filepath; if we don't include extra information I don't think we need to do the modification

dentiny:
Current codebase is broken, so don't panic if you see chaos test failure.

dpxcc:
Thanks for working on this!

Here is our guidelines for error handling: https://github.com/Mooncake-Labs/moonlink/issues/1135

We should only be using `expect()` when Error indicates a bug within Moonlink itself, otherwise the bug should always be propagated up. For example, errors in `equals_parquet_at_offset()` in `src/moonlink/src/row/moonlink_row.rs` should not use `expect()`

And for message in `expect()`, I would suggest to follow Rust doc: https://doc.rust-lang.org/std/result/enum.Result.html#method.expect. Basically, the message needs to explain why the error case should never happen

src/moonlink/src/row/moonlink_row.rs
mitanshu-23: Noted.

src/moonlink/src/union_read/table_metadata.rs
mitanshu-23: I’ll update the error message to include the string content.

mitanshu-23:
Hello @dentiny and @dpxcc , thank you for the guidance.

I discussed this with @dentiny, and @dentiny suggested categorizing the unwrap/expect cases into two parts:

1. Real I/O cases → where we should properly propagate errors (already handled by @owner).
2. Error message improvement cases → where unwraps can be replaced with expect to provide clearer context.

Based on guideline #1135 mentioned by @dpxcc , we should propagate errors only if they’re not Moonlink bugs. For the second category, most cases don’t require propagation, so my current focus is improving expect messages to include meaningful context (e.g., showing the string content when decoding fails).

Does this approach make sense? Please let me know if I’m missing something or if you’d like me to adjust direction.

mitanshu-23:
Hello @dentiny and @dpxcc , thank you for the guidance.

@dentiny suggested categorizing the unwrap/expect cases into two parts:

1. Real I/O cases → where we should properly propagate errors (is being handled by @dentiny).
2. Error message improvement cases → where unwraps can be replaced with expect to provide clearer context.

Based on guideline #1135 mentioned by @dpxcc , we should propagate errors only if they’re not Moonlink bugs. For the second category, most cases don’t require propagation, so my current focus is improving expect messages to include meaningful context (e.g., showing the string content when decoding fails).

Does this approach make sense? Please let me know if I’m missing something or if you’d like me to adjust direction.

dpxcc:
> @dentiny suggested categorizing the unwrap/expect cases into two parts:
>
> Real I/O cases → where we should properly propagate errors (is being handled by @dentiny).
> Error message improvement cases → where unwraps can be replaced with expect to provide clearer context.

I don’t think that categorization covers all cases. There are many errors that are not read I/O but still need to be propagated up, e.g. encoding/decoding failures, invalid user input (like malformed table configs or non-existent tables etc)

The guideline should be simple:
1. Use expect only if the Err case is truly unexpected and indicates a bug in moonlink itself
2. Propagate everything else up

Also, splitting the work by error categorization doesn’t seem like a good idea. Both contributors would need to scan the same code to decide which category each error falls into, which just duplicates effort

PR #1464, opened 2025-08-14
fix another edge case in deletion index lookup

zhousun:
## Summary

Deletion of the same value need to be process in the LSN order.
Otherwise, the deletion with larger LSN will see multiple candidates, and the one with smaller LSN will see less.

This normally never happens, because in general we add records to new_deletions in LSN order.

But streamingXact that delete's row in memslice, 
and the memslice got flushed before streamingXact finishes, 
may cause them to be added in different order.

## Related Issues

## Changes

- 
- 
- 

## Checklist

- [ ] Code builds correctly

### DIFF
--- a/src/moonlink/src/storage/mooncake_table/snapshot.rs
+++ b/src/moonlink/src/storage/mooncake_table/snapshot.rs
@@ -820,7 +820,7 @@ impl SnapshotTableState {
                 .map(|(loc, deletion)| Self::build_processed_deletion(deletion, loc))
                 .collect(),
             Ordering::Less => {
-                panic!("find less than expected candidates to deletions {deletions:?}")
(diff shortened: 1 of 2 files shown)


dentiny (commented):
Same here any chance to add a regression test? Or leave a todo / issue?

zhousun:
I would do it after replay system work, then we can just save replay log as tests

dentiny:
> I would do it after replay system work, then we can just save replay log as tests

A second thought, this is a non trivial workload. Current replay system relies on a monotonically increasing id to identify each event, but if we change our event type in the future (ie add a new event), event ID assignment is also changed.

Essentially it's hard to bind a replay event with interface.

dpxcc:
Just use event_type_name instead of event_type_id?
In case event_type_name also changes, it will immediately break the test and we can just update test

dentiny:
> Just use event_type_name instead of event_type_id?

You can have multiple events concurrently for the same event type.

PR #1472, opened 2025-08-14
[3/N] Update all interfaces to from database-id/table-id to schema/table

dentiny:
## Summary

This PR updates all public-facing interfaces from <database-id / table-id> to <schema-name / table-name>.

## Checklist

- [x] Code builds correctly

### DIFF
--- a/src/moonlink_backend/src/lib.rs
+++ b/src/moonlink_backend/src/lib.rs
@@ -189,21 +189,18 @@ impl MoonlinkBackend {
     }
 
     pub async fn drop_table(&self, schema: String, table: String) {
-        let mooncake_table_id = MooncakeTableId {
-            schema: schema.clone(),
-            table: table.clone(),
-        };
+        let mooncake_table_id = MooncakeTableId { schema, table };
 
         let table_exists = {
             let mut manager = self.replication_manager.write().await;
-            manager.drop_table(mooncake_table_id).await.unwrap()
+            manager.drop_table(&mooncake_table_id).await.unwrap()
         };
         if !table_exists {
             return;
         }
 
         self.metadata_store_accessor
-            .delete_table_metadata(&schema, &table)
+            .delete_table_metadata(&mooncake_table_id.schema, &mooncake_table_id.table)
             .await
             .unwrap()
     }
--- a/src/moonlink_backend/src/lib.rs
+++ b/src/moonlink_backend/src/lib.rs
@@ -17,7 +17,6 @@ pub use moonlink_connectors::{
     REST_API_URI,
 };
 use moonlink_metadata_store::base_metadata_store::MetadataStoreTrait;
-use std::hash::Hash;
 use std::sync::Arc;
 use tokio::sync::RwLock;
 
@@ -25,10 +24,7 @@ use crate::recovery_utils::BackendAttributes;
 use crate::table_config::TableConfig;
 use crate::table_status::TableStatus;
 
-pub struct MoonlinkBackend<
-    D: std::convert::From<u32> + Eq + Hash + Clone + std::fmt::Display,
-    T: std::convert::From<u32> + Eq + Hash + Clone + std::fmt::Display,
-> {
+pub struct MoonlinkBackend {
     // Base directory for all tables.
     base_path: String,
     // Functor used to remap local filepath within [`ReadState`] to data server URI if specified and if possible, so table access is routed to data server.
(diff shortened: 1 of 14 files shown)


src/moonlink_backend/src/lib.rs
```diff
@@ -140,19 +129,17 @@ where
...
         src_table_name: String,
         src_uri: String,
         table_config: String,
         input_schema: Option<Schema>,
     ) -> Result<()> {
         let mooncake_table_id = MooncakeTableId {
-            database_id: database_id.clone(),
-            table_id,
+            schema: schema.clone(),
+            table: table.clone(),
```
dpxcc, line 141: Can we avoid clone here? It seems that later code can just use `&mooncake_table_id.schema`

src/moonlink_backend/src/lib.rs
```diff
@@ -204,13 +188,11 @@ where
         Ok(())
     }
 
-    pub async fn drop_table(&self, database_id: D, table_id: T) {
+    pub async fn drop_table(&self, schema: String, table: String) {
         let mooncake_table_id = MooncakeTableId {
-            database_id: database_id.clone(),
-            table_id,
+            schema: schema.clone(),
+            table: table.clone(),
```
dpxcc: same

src/moonlink_datafusion/src/table_provider.rs
```diff
@@ -37,11 +37,17 @@ pub struct MooncakeTableProvider {
...
+    pub async fn try_new(
+        uri: &str,
+        schema_name: String,
+        table_name: String,
+        lsn: u64,
+    ) -> Result<Self> {
         let mut stream = UnixStream::connect(uri).await?;
-        let schema = get_table_schema(&mut stream, database_id, table_id).await?;
+        let schema = get_table_schema(&mut stream, schema_name.clone(), table_name.clone()).await?;
         let schema = StreamReader::try_new(schema.as_slice(), None)?.schema();
```
dpxcc: nit: I would suggest to change this from `schema` to `table_schema`, and rename your `schema_name` and `table_name` to `schema` and `table` to be consistent

src/moonlink_rpc/src/lib.rs
```diff
@@ -25,14 +25,14 @@ macro_rules! rpcs {
 }
 
 rpcs! {
-    create_snapshot(database_id: u32, table_id: u32, lsn: u64) -> ();
-    create_table(database_id: u32, table_id: u32, src: String, src_uri: String, table_config: String) -> ();
-    drop_table(database_id: u32, table_id: u32) -> ();
-    get_table_schema(database_id: u32, table_id: u32) -> Vec<u8>;
+    create_snapshot(schema: String, table: String, lsn: u64) -> ();
```
dpxcc, line 28: You are cloning schema and table for all rpc calls in moonlink_datafusion, and prob elsewhere
Instead of that, it would be better to prob change `$($name: $type)` to `$($name: &$type)` to only take a reference

dpxcc: nvm, I don't think this is possible - we need to construct the Request which has to own each arg

src/moonlink_backend/src/lib.rs
dentiny: https://github.com/Mooncake-Labs/moonlink/blob/e5f45af4421a4c02ade3453648f677c693cb46de/src/moonlink_backend/src/lib.rs#L167

https://github.com/Mooncake-Labs/moonlink/blob/e5f45af4421a4c02ade3453648f677c693cb46de/src/moonlink_backend/src/lib.rs#L179

we pass `mooncake_table_id` by value

src/moonlink_backend/src/lib.rs
dentiny: updated

src/moonlink_datafusion/src/table_provider.rs
dentiny: updated

src/moonlink_backend/src/lib.rs
dentiny: In `ReplicationManager::add_table`, mooncake table id's ownership gets passed to hash map: https://github.com/Mooncake-Labs/moonlink/blob/e5f45af4421a4c02ade3453648f677c693cb46de/src/moonlink_connectors/src/replication_manager.rs#L88-L89

dpxcc (approved):

Issue #1461, opened 2025-08-14
Add a separate from Cell type to row value type, which handles recursion.


peterxcli:
## Summary

https://github.com/peterxcli/moonlink/blob/7df09d87e3a65301ae0dda0085693daece1d836c/src/moonlink_connectors/src/pg_replicate/util.rs#L315

PR #1492, opened 2025-08-15
refactor: introduce recursive `from cell to RowValue`

peterxcli:
## Summary

Introduce from cell to RowValue which handle recursive for cleaner conversion

## Related Issues

Closes https://github.com/Mooncake-Labs/moonlink/issues/1461

## Changes

- Add `impl From<Cell> for RowValue`

## Checklist

- [x] Code builds correctly

### DIFF
--- a/src/moonlink_connectors/src/pg_replicate/util.rs
+++ b/src/moonlink_connectors/src/pg_replicate/util.rs
@@ -312,15 +312,8 @@ fn convert_array_cell(cell: ArrayCell) -> Vec<RowValue> {
             .into_iter()
             .map(|v| {
                 v.map(|cells| {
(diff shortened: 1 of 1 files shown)


peterxcli:
cc @dentiny PTAL. codecov seems like a false postive.

dentiny:
> codecov seems like a false postive.

I create another issue for UT coverage: https://github.com/Mooncake-Labs/moonlink/issues/1496; I will check and handle it.

peterxcli:
@dentiny Thanks for the review and for checking the codecov!

peterxcli:
BTW, I'm thinking merging the `convert_array_cell` and `impl From<Cell> for RowValue` together so we can have only one type matching path. But it seems not that easy to make that change😥

Maybe we could create a issue first and seeking other's opinion?

Issue #1457, opened 2025-08-14
[Improvement] Refine moonlink_service errors with new Error struct


machichima:
## Summary

Apply ErrorStruct to error types in moonlink_service.

Can simply follow what we have done for the `src/moonlink_rpc/src/error.rs` in PR https://github.com/Mooncake-Labs/moonlink/pull/1422, and apply changes to [src/moonlink_service/src/error.rs](https://github.com/Mooncake-Labs/moonlink/blob/main/src/moonlink_service/src/error.rs)

## Context

Error status should contain a few key attributes for better debuggability, including error status, message, source, and location. The ErrorStruct that includes all info needed for good error message is added in https://github.com/Mooncake-Labs/moonlink/pull/1422 to moonlink_error crate. We should use it in all other crates

## Tasks

- [ ] 

## Notes

Any extra notes or details that might be useful.

peterxcli:
Thanks @machichima for raising this. I would like to help on this.
cc @dentiny

dentiny:
> Thanks [@machichima](https://github.com/machichima) for raising this. I would like to help on this. cc [@dentiny](https://github.com/dentiny)

Thank you both!

PR #1493, opened 2025-08-15
Refine moonlink_service errors with new Error struct

peterxcli:
## Summary

Apply ErrorStruct to error types in moonlink_service.

Can simply follow what we have done for the src/moonlink_rpc/src/error.rs in PR https://github.com/Mooncake-Labs/moonlink/pull/1422, and apply changes to [src/moonlink_service/src/error.rs](https://github.com/Mooncake-Labs/moonlink/blob/main/src/moonlink_service/src/error.rs?rgh-link-date=2025-08-14T11%3A32%3A12.000Z)

## Related Issues

Closes https://github.com/Mooncake-Labs/moonlink/issues/1457

## Changes

- Apply ErrorStruct to error types in moonlink_service and implement proper error conversion 
- Adapt this new error interface in the rpc_server

## Checklist

- [x] Code builds correctly

### DIFF
--- a/src/moonlink_service/src/error.rs
+++ b/src/moonlink_service/src/error.rs
@@ -1,9 +1,7 @@
 use arrow_schema::ArrowError;
 use moonlink_error::{ErrorStatus, ErrorStruct};
 use std::io;
-use std::panic::Location;
 use std::result;
-use std::sync::Arc;
 use thiserror::Error;
 
 #[derive(Clone, Debug, Error)]
@@ -36,24 +34,17 @@ impl From<ArrowError> for Error {
             _ => ErrorStatus::Permanent,
         };
 
-        Error::Arrow(ErrorStruct {
-            message: format!("Arrow error: {source}"),
-            status,
-            source: Some(Arc::new(source.into())),
-            location: Some(Location::caller()),
-        })
+        Error::Arrow(ErrorStruct::new(format!("Arrow error: {source}"), status).with_source(source))
     }
 }
 
 impl From<moonlink_backend::Error> for Error {
     #[track_caller]
     fn from(source: moonlink_backend::Error) -> Self {
-        Error::Backend(ErrorStruct {
-            message: format!("Backend error: {source}"),
-            status: ErrorStatus::Permanent,
-            source: Some(Arc::new(source.into())),
-            location: Some(Location::caller()),
-        })
+        Error::Backend(
+            ErrorStruct::new(format!("Backend error: {source}"), ErrorStatus::Permanent)
+                .with_source(source),
+        )
     }
 }
 
@@ -76,40 +67,34 @@ impl From<io::Error> for Error {
             _ => ErrorStatus::Permanent,
         };
 
-        Error::Io(ErrorStruct {
-            message: format!("IO error: {source}"),
-            status,
-            source: Some(Arc::new(source.into())),
-            location: Some(Location::caller()),
-        })
+        Error::Io(ErrorStruct::new(format!("IO error: {source}"), status).with_source(source))
     }
 }
 
 impl From<moonlink_rpc::Error> for Error {
     #[track_caller]
     fn from(source: moonlink_rpc::Error) -> Self {
-        Error::Rpc(ErrorStruct {
-            message: format!("RPC error: {source}"),
-            status: match &source {
-                moonlink_rpc::Error::Io(err) => err.status,
-                moonlink_rpc::Error::Decode(err) => err.status,
-                moonlink_rpc::Error::Encode(err) => err.status,
-                moonlink_rpc::Error::PacketTooLong(err) => err.status,
-            },
-            source: Some(Arc::new(source.into())),
-            location: Some(Location::caller()),
-        })
+        Error::Rpc(
+            ErrorStruct::new(
+                format!("RPC error: {source}"),
+                match &source {
+                    moonlink_rpc::Error::Io(err) => err.status,
+                    moonlink_rpc::Error::Decode(err) => err.status,
+                    moonlink_rpc::Error::Encode(err) => err.status,
+                    moonlink_rpc::Error::PacketTooLong(err) => err.status,
+                },
+            )
+            .with_source(source),
+        )
     }
 }
 
 impl From<tokio::task::JoinError> for Error {
     #[track_caller]
     fn from(source: tokio::task::JoinError) -> Self {
-        Error::TaskJoin(ErrorStruct {
-            message: format!("Join error: {source}"),
-            status: ErrorStatus::Permanent,
-            source: Some(Arc::new(source.into())),
-            location: Some(Location::caller()),
-        })
+        Error::TaskJoin(
+            ErrorStruct::new(format!("Join error: {source}"), ErrorStatus::Permanent)
+                .with_source(source),
+        )
     }
 }
--- a/src/moonlink_service/src/error.rs
+++ b/src/moonlink_service/src/error.rs
@@ -1,4 +1,5 @@
 use arrow_schema::ArrowError;
+use moonlink_error::io_error_utils::get_io_error_status;
 use moonlink_error::{ErrorStatus, ErrorStruct};
 use std::io;
 use std::result;
@@ -51,22 +52,7 @@ impl From<moonlink_backend::Error> for Error {
 impl From<io::Error> for Error {
     #[track_caller]
     fn from(source: io::Error) -> Self {
-        let status = match source.kind() {
-            io::ErrorKind::TimedOut
-            | io::ErrorKind::Interrupted
-            | io::ErrorKind::WouldBlock
-            | io::ErrorKind::ConnectionRefused
-            | io::ErrorKind::ConnectionAborted
-            | io::ErrorKind::ConnectionReset
-            | io::ErrorKind::BrokenPipe
-            | io::ErrorKind::NetworkDown
-            | io::ErrorKind::ResourceBusy
-            | io::ErrorKind::QuotaExceeded => ErrorStatus::Temporary,
-
-            // All other errors are permanent
-            _ => ErrorStatus::Permanent,
-        };
-
+        let status = get_io_error_status(&source);
         Error::Io(ErrorStruct::new("IO error".to_string(), status).with_source(source))
     }
 }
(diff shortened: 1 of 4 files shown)


peterxcli:
cc @dentiny, @machichima PTAL. Thanks!

src/moonlink_service/src/rpc_server.rs
```diff
@@ -31,13 +30,17 @@ pub async fn start_unix_server(
...
+                Err(Error::Rpc(error_struct)) => {
+                    if let Some(source) = error_struct.source.as_ref() {
+                        if let Some(io_err) = source.downcast_ref::<std::io::Error>() {
+                            if matches!(io_err.kind(), BrokenPipe | ConnectionReset | UnexpectedEof)
+                            {
+                                return;
+                            }
+                        }
+                    }
+                    panic!("{error_struct}");
```
dentiny: nit: let's add some prompt words

src/moonlink_service/src/rpc_server.rs
```diff
@@ -55,13 +58,17 @@ pub async fn start_tcp_server(backend: Arc<MoonlinkBackend>, addr: SocketAddr) -
...
```
dentiny: ditto

src/moonlink_service/src/rpc_server.rs
```diff
@@ -31,14 +30,18 @@ pub async fn start_unix_server(
...
+                Err(Error::Rpc(error_struct)) => {
+                    if let Some(source) = error_struct.source.as_ref() {
+                        if let Some(io_err) = source.downcast_ref::<std::io::Error>() {
+                            if matches!(io_err.kind(), BrokenPipe | ConnectionReset | UnexpectedEof)
+                            {
+                                return;
+                            }
+                        }
+                    }
+                    panic!("Unix RPC server error: {error_struct}");
```
machichima: Curious, why are we making those changes?

peterxcli: to adapt the newly introduced `Error` type. we have to take the upstream error out to align the existing logic for `matches!(io_err.kind(), BrokenPipe | ConnectionReset | UnexpectedEof)` matching

src/moonlink_service/src/error.rs
```diff
@@ -1,15 +1,118 @@
...
+        })
+    }
+}
+
+impl From<moonlink_backend::Error> for Error {
+    #[track_caller]
+    fn from(source: moonlink_backend::Error) -> Self {
+        Error::Backend(ErrorStruct {
+            message: format!("Backend error: {source}"),
+            status: ErrorStatus::Permanent,
```
dentiny: Backend errors should have `ErrorStruct` inside, so whether permanent/transient should inherit from internal error

src/moonlink_service/src/error.rs
```diff
@@ -1,15 +1,118 @@
...
+        })
+    }
+}
+
+impl From<moonlink_rpc::Error> for Error {
+    #[track_caller]
+    fn from(source: moonlink_rpc::Error) -> Self {
+        let status = match &source {
+            // Use the status that was already determined in the moonlink_rpc crate
+            moonlink_rpc::Error::Io(err) => err.status,
```
dentiny: curious why don't we simply inherit status from internal moonlink rpc error?

src/moonlink_service/src/rpc_server.rs
```diff
@@ -55,14 +58,18 @@ pub async fn start_tcp_server(backend: Arc<MoonlinkBackend>, addr: SocketAddr) -
...
-                            matches!(io_err.kind(), BrokenPipe | ConnectionReset | UnexpectedEof)
-                        })
-                        .unwrap_or(false) => {}
-                Err(e) => panic!("{e}"),
+                Err(Error::Rpc(error_struct)) => {
+                    if let Some(source) = error_struct.source.as_ref() {
+                        if let Some(io_err) = source.downcast_ref::<std::io::Error>() {
+                            if matches!(io_err.kind(), BrokenPipe | ConnectionReset | UnexpectedEof)
+                            {
+                                return;
```
dentiny: why return? prev logic is do nothing

src/moonlink_service/src/rpc_server.rs
```diff
@@ -31,14 +30,18 @@ pub async fn start_unix_server(
...
```
dentiny: looks like an issue, we should do nothing here if not matched?

src/moonlink_service/src/rpc_server.rs
```diff
@@ -31,14 +30,18 @@ pub async fn start_unix_server(
...
-                        .and_then(|src| src.downcast_ref::<std::io::Error>())
-                        .map(|io_err| {
-                            matches!(io_err.kind(), BrokenPipe | ConnectionReset | UnexpectedEof)
-                        })
-                        .unwrap_or(false) => {}
-                Err(e) => panic!("{e}"),
+                Err(Error::Rpc(error_struct)) => {
+                    if let Some(source) = error_struct.source.as_ref() {
+                        if let Some(io_err) = source.downcast_ref::<std::io::Error>() {
+                            if matches!(io_err.kind(), BrokenPipe | ConnectionReset | UnexpectedEof)
```
dentiny: (no requested in this PR) we should make a util for acceptable IO error kinds

src/moonlink_service/src/rpc_server.rs
machichima: I think then the following `panic!("Unix RPC server error: {error_struct}");` will be called if not matched

peterxcli: Correct, thanks @machichima for explanation.

src/moonlink_service/src/error.rs
peterxcli: you are right, mybrain isn't functioning...

src/moonlink_service/src/error.rs
peterxcli: not yet: https://github.com/Mooncake-Labs/moonlink/blob/a00360262f63abc41c6247f8b3b616ee2d405892/src/moonlink_backend/src/error.rs#L10-L33

src/moonlink_service/src/rpc_server.rs
peterxcli: @dentiny 
How about we change the logic to: "if the error status is `ErrorStatus::Permanent`, then panic, otherwise"

src/moonlink_service/src/rpc_server.rs
peterxcli: https://github.com/Mooncake-Labs/moonlink/pull/1493#discussion_r2280310387

src/moonlink_service/src/error.rs
peterxcli: https://github.com/Mooncake-Labs/moonlink/pull/1493/commits/8170bc05f363457e56a886e2441e851b0cbfd1f1
done, thanks for catching this 👍

src/moonlink_service/src/error.rs
dentiny: gotcha!

src/moonlink_service/src/rpc_server.rs
dentiny: sorry for the typo, I mean 

we should do nothing here if **matched**?

dentiny: essentially for these errors we should keep reading from stream

src/moonlink_service/src/rpc_server.rs
dentiny: > How about we change the logic to: "if the error status is ErrorStatus::Permanent, then panic, otherwise"

I think the logic here is to only retry on unix-domain specific transient errors, but just general-purpose transient errors (i.e., generally we treat resource exhausted as retryable, but don't think it apply here).

machichima:
Hi @peterxcli ,

[This PR](https://github.com/Mooncake-Labs/moonlink/pull/1519) makes `location` into `String` type to make ErrorStruct (de)serializable, could you please use following format in `impl From<XXXError> for Error` instead? It can construct the `ErrorStruct` instance without needed to specify fields explicitly.

Thank you!

```rust
        Error::Encode(
            ErrorStruct::new("xxxxx".to_string(), status)
                .with_source(source),
        )
```

src/moonlink_service/src/rpc_server.rs
```diff
@@ -31,14 +31,15 @@ pub async fn start_unix_server(
         let backend = Arc::clone(&backend);
         tokio::spawn(async move {
             match handle_stream(backend, stream).await {
-                Err(Error::Rpc(moonlink_rpc::Error::Io(e)))
-                    if error::Error::source(&e)
+                Err(Error::Rpc(error_struct))
+                    if error_struct
+                        .source()
```
peterxcli, line 36: if let guards in match arms are experimental and not stable, and the .source() impl require to import `std::error::Error`

peterxcli:
> Hi @peterxcli ,
> 
> [This PR](https://github.com/Mooncake-Labs/moonlink/pull/1519) makes `location` into `String`...

done. https://github.com/Mooncake-Labs/moonlink/pull/1493/commits/9a1801e7e64f4f5e618e204e6f4752dc34e33bec.
thanks!

peterxcli:
hi @dentiny, @machichima I think I misunderstood the syntax, I've revert the match pattern change in the rpc server, please take another look. Thanks!

src/moonlink_service/src/error.rs
```diff
@@ -1,15 +1,100 @@
...
+    #[track_caller]
+    fn from(source: ArrowError) -> Self {
+        let status = match source {
+            ArrowError::IoError(_, _) => ErrorStatus::Temporary,
+
+            // All other errors are regard as permanent
+            _ => ErrorStatus::Permanent,
+        };
+
+        Error::Arrow(ErrorStruct::new(format!("Arrow error: {source}"), status).with_source(source))
```
dentiny: `source` here is duplicate at both string message, also as a separate source field, don't think we need it twice.

src/moonlink_service/src/error.rs
```diff
@@ -1,15 +1,100 @@
...
+                .with_source(source),
+        )
+    }
+}
+
+impl From<io::Error> for Error {
+    #[track_caller]
+    fn from(source: io::Error) -> Self {
+        let status = match source.kind() {
+            io::ErrorKind::TimedOut
```
dentiny: maybe use this util function to avoid code duplication?
https://github.com/Mooncake-Labs/moonlink/blob/main/src/moonlink_error/src/io_error_utils.rs

dentiny (commented):
Thanks for working on this! only two minor comments

Issue #1487, opened 2025-08-15
BUG: Source location information lost on propagation


dentiny:
## Summary

https://github.com/Mooncake-Labs/moonlink/actions/runs/16979985977/job/48137770236?pr=1473

```sh
called `Result::unwrap()` on an `Err` value: IcebergError(ErrorStruct { message: "Iceberg error: DataInvalid => Cannot add files that are already referenced by table, files: s3://test-minio-warehouse-z63sbnplo38h/namespace/table/data/data-0198ab2a-ae65-76d0-a710-8d27e7403320.parquet, s3://test-minio-warehouse-z63sbnplo38h/namespace/table/data/data-0198ab2a-ae6a-7a50-97c1-21440261af6f.parquet, s3://test-minio-warehouse-z63sbnplo38h/namespace/table/data/data-0198ab2a-ae6b-70b0-bd10-eefc321e33e0.parquet", status: Permanent, source: Some(DataInvalid => Cannot add files that are already referenced by table, files: s3://test-minio-warehouse-z63sbnplo38h/namespace/table/data/data-0198ab2a-ae65-76d0-a710-8d27e7403320.parquet, s3://test-minio-warehouse-z63sbnplo38h/namespace/table/data/data-0198ab2a-ae6a-7a50-97c1-21440261af6f.parquet, s3://test-minio-warehouse-z63sbnplo38h/namespace/table/data/data-0198ab2a-ae6b-70b0-bd10-eefc321e33e0.parquet), location: Some(Location { file: "src/moonlink/src/storage/mooncake_table.rs", line: 1425, col: 87 }) })
```

PR #1495, opened 2025-08-15
Fix iceberg error propagation

dentiny:
## Summary

To keep source location information during error propagation, we should avoid wrapping ourselves.

## Related Issues

Closes https://github.com/Mooncake-Labs/moonlink/issues/1487

## Checklist

- [ ] Code builds correctly

### DIFF
--- a/src/moonlink/src/table_handler.rs
+++ b/src/moonlink/src/table_handler.rs
@@ -17,7 +17,6 @@ use crate::storage::snapshot_options::SnapshotOption;
 use crate::storage::{io_utils, MooncakeTable};
 use crate::table_handler_timer::TableHandlerTimer;
 use crate::table_notify::TableEvent;
-use crate::Error;
 use tokio::sync::mpsc::{self, Receiver, Sender};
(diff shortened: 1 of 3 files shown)


src/moonlink/src/table_handler.rs
```diff
@@ -580,9 +579,7 @@ impl TableHandler {
                                 .update_iceberg_persisted_lsn(iceberg_flush_lsn, replication_lsn);
                         }
                         Err(e) => {
-                            let err = Err(Error::IcebergMessage(format!(
-                                "Failed to create iceberg snapshot: {e:?}"
-                            )));
+                            let err = Err(e.clone());
```
dentiny, line 582: `e` here contains `ErrorStruct` and source information

dentiny:
Source information is automatically generated only when XXXError converts to `Error` type, for most of the cases it's fine;
In iceberg I'm using iceberg error here and there; two ways to resolve:
- Rewrite the whole iceberg error propagation
- Attach source location information at source error

Issue #974, opened 2025-07-22
wal / object store


lohpaul9:
## Summary
All WAL features should work on all the different storage backends

lohpaul9:
closed by #1680

PR #1497, opened 2025-08-15
refactor before obj store changes

lohpaul9:
## Summary

Refactoring to use WalConfig and to make any reads / writes in testing go through the wal file system accessor instead of directly using the file system.  

## Related Issues

#974 

## Checklist

- [X ] Code builds correctly
- [ X] Tests have been added or updated

### DIFF
--- a/src/moonlink/src/storage/wal.rs
+++ b/src/moonlink/src/storage/wal.rs
@@ -15,11 +15,12 @@ use std::pin::Pin;
 use std::sync::Arc;
 use tokio::sync::mpsc::Sender;
 
-pub const DEFAULT_WAL_FOLDER: &str = "wal";
+pub const DEFAULT_WAL_FOLDER: &str = "_wal";
 
-#[derive(Debug, Clone)]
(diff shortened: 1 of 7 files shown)


src/moonlink/src/storage/wal.rs
```diff
@@ -343,6 +357,10 @@ impl PersistentWalMetadata {
     pub fn get_iceberg_snapshot_lsn(&self) -> Option<u64> {
         self.iceberg_snapshot_lsn
     }
+
+    pub fn get_mooncake_table_id(&self) -> &str {
```
dentiny, line 361: mooncake table id is a transient concept, which means after recovery it changes
so it should NOT be used to indicate path

lohpaul9: What I am using here is mooncake_table_id, not table_id, which looks to be the combination of a given schema and table. This should be unique, unlike table_id which is transient.

dentiny: Thanks for the clarification!

PR #1506, opened 2025-08-15
gated wal flush pub function

lohpaul9:
## Summary

Gates moonlink backend's `wait_for_wal_flush` behind a test util flag, so that it is not a public facing API.

### DIFF
--- a/src/moonlink_backend/Cargo.toml
+++ b/src/moonlink_backend/Cargo.toml
@@ -8,6 +8,7 @@ license = { workspace = true }
 default = ["storage-fs"]
 
 profiling = ["console-subscriber"]
+test-utils = []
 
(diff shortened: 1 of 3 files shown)


src/moonlink_backend/Cargo.toml
```diff
@@ -8,6 +8,7 @@ license = { workspace = true }
 default = ["storage-fs"]
 
 profiling = ["console-subscriber"]
+test-utils = []
```
dentiny, line 11: please use moonlink one: https://github.com/Mooncake-Labs/moonlink/blob/2d8fc2b349e55e0c10072d05adf002aace8f5753/src/moonlink/Cargo.toml#L10
otherwise each crate has its own separate feature

Issue #1488, opened 2025-08-15
snapshot read: avoid unwrap potential IO error


dentiny:
## Summary

https://github.com/Mooncake-Labs/moonlink/blob/6437db10c8b191d2b4553fbc5216e1e944c3307f/src/moonlink/src/storage/mooncake_table/snapshot_read_output.rs#L74-L85

peterxcli:
I would like to help on this~

dentiny:
> I would like to help on this~

thanks! Assigned

PR #1513, opened 2025-08-16
snapshot read: avoid unwrap potential IO error

peterxcli:
## Summary

https://github.com/Mooncake-Labs/moonlink/blob/6437db10c8b191d2b4553fbc5216e1e944c3307f/src/moonlink/src/storage/mooncake_table/snapshot_read_output.rs#L74-L85

## Related Issues

Closes https://github.com/Mooncake-Labs/moonlink/issues/1488

## Changes

- return Result instead of unwrap and panic if IO error happened.

## Checklist

- [x] Code builds correctly

### DIFF
--- a/src/moonlink/src/storage/mooncake_table/data_file_state_tests.rs
+++ b/src/moonlink/src/storage/mooncake_table/data_file_state_tests.rs
@@ -268,7 +268,8 @@ async fn prepare_state_3(
         let snapshot_read_output = perform_read_request_for_test(&mut table).await;
         let read_state = snapshot_read_output
             .take_as_read_state(get_read_state_filepath_remap())
-            .await;
+            .await
+            .unwrap();
 
         // Persist.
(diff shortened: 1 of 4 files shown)


peterxcli:
cc @dentiny PTAL. Thanks!

dentiny:
Hi @peterxcli could you please help resolve the conflict? Thank you!

peterxcli:
done. thanks

dentiny (approved):
This PR worth a followup item, some context here:
- `take_as_read_state` tries to read object from remote and pin in the data file cache, that's why we have error status;
- for the current implementation, unwrap directly crashes moonlink, which is definitely bad but we could recover;
- after switching to error propagation, the cache handle is still pinned in cache, so we actually need to unpin before leave.

I don't think it's a merge blocker (since we haven't released officially), but please do followup.

peterxcli:
Thanks for the review, created one for above suggestion/comments - https://github.com/Mooncake-Labs/moonlink/issues/1518

Issue #1435, opened 2025-08-13
Enable bloom filter for compacted files


dentiny:
## Summary

Due to severe size boom on small files, I didn't enable bloom filter by default.
But worth enabling for compacted ones.

Issue #525, opened 2025-06-20
storage: configure parquet files compatible with pg_mooncake v0.1


dentiny:
## Summary

moonlink config:
https://github.com/Mooncake-Labs/moonlink/blob/main/src/moonlink/src/storage/parquet_utils.rs

pg_mooncake v0.1:
https://github.com/Mooncake-Labs/pg_mooncake/blob/d68f1e5f023ba36e0810dbcf7d8deecc275eccc8/src/columnstore/columnstore_table.cpp#L66-L71

fweilun:
Im interested in this issue. Could you assign this to me?

dentiny:
> Im interested in this issue. Could you assign this to me?

Sure, thanks!

PR #1523, opened 2025-08-16
[storage] Set storage bloom filter for compacted files

dentiny:
## Summary

This PR does two things:
- I found bloom filter affect file size A LOT; for example, for file row count <100, file size could increase from 100 byte to 6MB
- Fix estimated size into memory size, I found prev calculation doesn't account for bloom filter size

## Related Issues

Closes https://github.com/Mooncake-Labs/moonlink/issues/1435
Closes https://github.com/Mooncake-Labs/moonlink/issues/525

## Checklist

- [x] Code builds correctly

### DIFF
--- a/src/moonlink/src/storage/parquet_utils.rs
+++ b/src/moonlink/src/storage/parquet_utils.rs
@@ -24,7 +24,5 @@ pub(crate) fn get_default_parquet_properties() -> WriterProperties {
 /// Parquet option for already compacted files.
 pub(crate) fn get_parquet_properties_for_compaction() -> WriterProperties {
     let builder = get_default_parquet_properties_builder();
-    builder
-        .set_bloom_filter_enabled(true)
-        .build()
+    builder.set_bloom_filter_enabled(true).build()
 }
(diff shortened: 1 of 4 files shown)


src/moonlink/src/storage/parquet_utils.rs
```diff
@@ -1,17 +1,28 @@
...
+
+pub(crate) fn get_default_parquet_properties() -> WriterProperties {
+    let builder = get_default_parquet_properties_builder();
+    builder.build()
+}
+
+/// Parquet option for already compacted files.
+pub(crate) fn get_parquet_properties_for_compaction() -> WriterProperties {
+    let builder = get_default_parquet_properties_builder();
+    builder.set_bloom_filter_enabled(true).build()
```
dpxcc, line 27: To be fully compatible with the default settings in DuckDB, you also need to `set_bloom_filter_fpp` and `set_bloom_filter_ndv`

DuckDB aims for fpp=0.01 (https://github.com/duckdb/duckdb/blob/v1.3.2/extension/parquet/parquet_extension.cpp#L232) and hit_ratio=0.01% (https://github.com/duckdb/duckdb/blob/v1.3.2/extension/parquet/parquet_statistics.cpp#L566)

dentiny: Sure, PR raised: https://github.com/Mooncake-Labs/moonlink/pull/1526

Issue #1524, opened 2025-08-16
Add pg cdc event integration test


peterxcli:
## Summary

Use this to test the change stream from pgoutput can be parse as correct CdcEvent

## Context

Why is this issue important? Link any relevant discussions, designs, or documents.

## Tasks

- [ ] type conversion
  - primitive types
  - composite, array of composite, nested array(1-dim) in composite
  - multi-dim array
- [ ] More CdcEvent enum coverage, eg. Relation, StreamStop...

## Notes

ref:
- [pkg/source/postgres_test.go](https://github.com/replicase/pgcapture/blob/ae7ef5f35660495f45a7a5f0891acbd68e0dd1b4/pkg/source/postgres_test.go): End-to-end capture from Postgres logical replication (DDL/DML, tx boundaries, checkpoints).
- [pkg/source/pulsar_test.go](https://github.com/replicase/pgcapture/blob/ae7ef5f35660495f45a7a5f0891acbd68e0dd1b4/pkg/source/pulsar_test.go): Capture from Pulsar topic as a CDC source (message ordering, checkpoints).

- [pkg/decode/pglogical_test.go](https://github.com/replicase/pgcapture/blob/ae7ef5f35660495f45a7a5f0891acbd68e0dd1b4/pkg/decode/pglogical_test.go): Decoding CDC messages from pglogical (INSERT/UPDATE/DELETE, various types).
- [pkg/decode/pgoutput_test.go](https://github.com/replicase/pgcapture/blob/ae7ef5f35660495f45a7a5f0891acbd68e0dd1b4/pkg/decode/pgoutput_test.go): Decoding CDC messages from pgoutput (Postgres 14+).

- [pkg/sink/postgres_test.go](https://github.com/replicase/pgcapture/blob/ae7ef5f35660495f45a7a5f0891acbd68e0dd1b4/pkg/sink/postgres_test.go): Applying CDC changes to Postgres sink (batching, data types, tx handling).
- [pkg/sink/pulsar_test.go](https://github.com/replicase/pgcapture/blob/ae7ef5f35660495f45a7a5f0891acbd68e0dd1b4/pkg/sink/pulsar_test.go): Writing CDC messages to Pulsar sink (checkpoint tracking and commit).

[...]

PR #1525, opened 2025-08-16
Add integration tests for CDC event conversion from logical replication stream with composite types

peterxcli:
## Summary

Add integration tests for CDC event conversion from logical replication stream with composite types

## Related Issues

Closes https://github.com/Mooncake-Labs/moonlink/issues/1524

## Changes

- Add pg cdc event IT skeleton and one `test_composite_types_in_cdc_stream` test case to test   **composite**, **array of composite**, **nested composite** and **nested array in composite**
- Add a `TestResources` struct with drop implemented to release pg resource
- Correct pg wal level config correctly in github ci file

## Checklist

- [x] Code builds correctly

## Notes

```
Event 14: XLogData(XLogDataBody { wal_start: 54529448, wal_end: 54529448, timestamp: 808677494744232, data: Update(UpdateBody { xid: None, rel_id: 28017, old_tuple: None, key_tuple: None, new_tuple: Tuple([Text(b"1"), Text(b"(\"Updated St\",\"Updated City\",54321)"), Text(b"(Home,\"(40.7,-74)\")"), Text(b"(Alice,\"{\"\"(\\\\\"\"123 Main St\\\\\"\",NYC,10001)\"\",\"\"(\\\\\"\"456 Oak Ave\\\\\"\",LA,90210)\"\"}\",\"(Work,\"\"(34,-118)\"\")\")"), Text(b"{\"(\\\"789 Pine St\\\",Chicago,60601)\",\"(\\\"321 Elm St\\\",Boston,2101)\"}")]) }) })
```

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -425,7 +425,6 @@ jobs:
           --health-retries 5
         ports:
           - 5432:5432
-        # no host volumes to avoid root-owned directories pre-checkout
 
     steps:
       - uses: actions/checkout@v5
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -456,8 +456,6 @@ jobs:
       - name: Configure PostgreSQL for logical replication
         run: |
           PGPASSWORD=postgres psql -h localhost -p 5432 -U postgres -d postgres -c "ALTER SYSTEM SET wal_level = 'logical';"
-          PGPASSWORD=postgres psql -h localhost -p 5432 -U postgres -d postgres -c "ALTER SYSTEM SET max_wal_senders = 16;"
-          PGPASSWORD=postgres psql -h localhost -p 5432 -U postgres -d postgres -c "ALTER SYSTEM SET max_replication_slots = 16;"
           PGPASSWORD=postgres psql -h localhost -p 5432 -U postgres -d postgres -c "SELECT pg_reload_conf();"
 
       - name: Restart PostgreSQL to apply wal_level
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -418,12 +418,14 @@ jobs:
           POSTGRES_PASSWORD: postgres
           POSTGRES_DB: postgres
         options: >-
+          --name postgres
           --health-cmd pg_isready
           --health-interval 10s
           --health-timeout 5s
           --health-retries 5
         ports:
           - 5432:5432
+        # no host volumes to avoid root-owned directories pre-checkout
 
     steps:
       - uses: actions/checkout@v5
@@ -454,7 +456,21 @@ jobs:
       - name: Configure PostgreSQL for logical replication
         run: |
           PGPASSWORD=postgres psql -h localhost -p 5432 -U postgres -d postgres -c "ALTER SYSTEM SET wal_level = 'logical';"
+          PGPASSWORD=postgres psql -h localhost -p 5432 -U postgres -d postgres -c "ALTER SYSTEM SET max_wal_senders = 16;"
+          PGPASSWORD=postgres psql -h localhost -p 5432 -U postgres -d postgres -c "ALTER SYSTEM SET max_replication_slots = 16;"
           PGPASSWORD=postgres psql -h localhost -p 5432 -U postgres -d postgres -c "SELECT pg_reload_conf();"
+
+      - name: Restart PostgreSQL to apply wal_level
+        run: |
+          docker restart postgres
+          for i in {1..10}; do
+            if pg_isready -h localhost -p 5432 -U postgres; then
+              echo "PostgreSQL restarted and ready!"
+              break
+            fi
+            echo "Waiting for PostgreSQL after restart..."
+            sleep 2
+          done
 
       - name: Run PostgreSQL integration tests
         timeout-minutes: 10
--- a/src/moonlink_connectors/src/pg_replicate/tests/cdc_event_test.rs
+++ b/src/moonlink_connectors/src/pg_replicate/tests/cdc_event_test.rs
@@ -15,6 +15,7 @@ use tokio_postgres::{connect, NoTls};
 
 const STREAM_NEXT_TIMEOUT_MS: u64 = 100;
 const EVENT_COLLECTION_SECS: u64 = 5;
+const PUBLIC_SCHEMA: &str = "public";
 
 async fn fetch_table_schema(publication: &str, table_name_str: &str) -> TableSchema {
     let url = database_url();
@@ -26,7 +27,7 @@ async fn fetch_table_schema(publication: &str, table_name_str: &str) -> TableSch
     });
     let mut schema_client = ReplicationClient::from_client(schema_pg_client);
     let table_name = TableName {
-        schema: "public".to_string(),
+        schema: PUBLIC_SCHEMA.to_string(),
         name: table_name_str.to_string(),
     };
     let src_table_id = schema_client
@@ -46,15 +47,11 @@ fn spawn_sql_executor(database_url: String) -> mpsc::UnboundedSender<String> {
     tokio::spawn(async move {
         let (bg_client, bg_connection) = connect(&database_url, NoTls).await.unwrap();
         tokio::spawn(async move {
-            if let Err(e) = bg_connection.await {
-                eprintln!("Background connection error: {e}");
-            }
+            bg_connection.await.unwrap();
         });
 
         while let Some(sql) = rx.recv().await {
-            if let Err(e) = bg_client.simple_query(&sql).await {
-                eprintln!("background SQL failed: {e}; sql: {sql}");
-            }
+            bg_client.simple_query(&sql).await.unwrap();
         }
     });
     tx
@@ -141,6 +138,7 @@ async fn test_composite_types() {
     let mut replication_client = create_replication_client().await;
 
     // Create replication slot using the replication client
+    // https://github.com/supabase/etl/blob/4da956c6b9be8476a1dbe87a4d88689e0671b7c1/etl/docs/Replication%20in%20Postgres.md?plain=1#L70
     replication_client
         .begin_readonly_transaction()
         .await
@@ -429,6 +427,7 @@ async fn test_null() {
         .unwrap();
 
     let mut replication_client = create_replication_client().await;
+    // https://github.com/supabase/etl/blob/4da956c6b9be8476a1dbe87a4d88689e0671b7c1/etl/docs/Replication%20in%20Postgres.md?plain=1#L70
     replication_client
         .begin_readonly_transaction()
         .await
--- a/src/moonlink_connectors/src/pg_replicate/tests/cdc_event_test.rs
+++ b/src/moonlink_connectors/src/pg_replicate/tests/cdc_event_test.rs
@@ -1,61 +1,21 @@
 #![cfg(feature = "connector-pg")]
 
-use super::test_utils::{create_replication_client, database_url, setup_connection, TestResources};
-use crate::pg_replicate::clients::postgres::ReplicationClient;
+use super::test_utils::{
+    create_publication_for_table, create_replication_client_and_slot, database_url,
+    fetch_table_schema, set_replica_identity_full, setup_connection, spawn_sql_executor,
+    TestResources,
+};
 use crate::pg_replicate::conversions::cdc_event::CdcEvent;
 use crate::pg_replicate::conversions::Cell;
 use crate::pg_replicate::postgres_source::{CdcStreamConfig, PostgresSource};
-use crate::pg_replicate::table::{TableName, TableSchema};
 use futures::StreamExt;
 use serial_test::serial;
 use std::collections::VecDeque;
 use std::time::Duration;
-use tokio::sync::mpsc;
-use tokio_postgres::{connect, NoTls};
 
 const STREAM_NEXT_TIMEOUT_MS: u64 = 100;
 const EVENT_COLLECTION_SECS: u64 = 5;
-const PUBLIC_SCHEMA: &str = "public";
-
-async fn fetch_table_schema(publication: &str, table_name_str: &str) -> TableSchema {
-    let url = database_url();
-    let (schema_pg_client, schema_conn) = connect(&url, NoTls).await.unwrap();
-    tokio::spawn(async move {
-        if let Err(e) = schema_conn.await {
-            eprintln!("Schema connection error: {e}");
-        }
-    });
-    let mut schema_client = ReplicationClient::from_client(schema_pg_client);
-    let table_name = TableName {
-        schema: PUBLIC_SCHEMA.to_string(),
-        name: table_name_str.to_string(),
-    };
-    let src_table_id = schema_client
-        .get_src_table_id(&table_name)
-        .await
-        .unwrap()
-        .expect("missing table id for table {table_name}");
-    schema_client
-        .get_table_schema(src_table_id, table_name, Some(publication))
-        .await
-        .unwrap()
-}
-
-/// Spawns a background SQL executor that can be used to submit arbitrary SQL in desired order with optional delays between them
-fn spawn_sql_executor(database_url: String) -> mpsc::UnboundedSender<String> {
-    let (tx, mut rx) = mpsc::unbounded_channel::<String>();
-    tokio::spawn(async move {
-        let (bg_client, bg_connection) = connect(&database_url, NoTls).await.unwrap();
-        tokio::spawn(async move {
-            bg_connection.await.unwrap();
-        });
-
-        while let Some(sql) = rx.recv().await {
-            bg_client.simple_query(&sql).await.unwrap();
-        }
-    });
-    tx
-}
+// Common helpers moved to tests/test_utils.rs
 
 #[tokio::test(flavor = "multi_thread", worker_threads = 2)]
 #[serial]
@@ -102,20 +62,10 @@ async fn test_composite_types() {
         .unwrap();
 
     // Ensure replica identity is FULL so replication uses full row images
-    resources
-        .client()
-        .simple_query(&format!("ALTER TABLE {table_name} REPLICA IDENTITY FULL;"))
-        .await
-        .unwrap();
+    set_replica_identity_full(resources.client(), &table_name).await;
 
     // Create publication
-    resources
-        .client()
-        .simple_query(&format!(
-            "CREATE PUBLICATION {publication} FOR TABLE {table_name};"
-        ))
-        .await
-        .unwrap();
+    create_publication_for_table(resources.client(), &publication, &table_name).await;
     resources
         .client()
         .simple_query(&format!(
@@ -135,27 +85,14 @@ async fn test_composite_types() {
         .await
         .unwrap();
 
-    let mut replication_client = create_replication_client().await;
-
-    // Create replication slot using the replication client
-    // https://github.com/supabase/etl/blob/4da956c6b9be8476a1dbe87a4d88689e0671b7c1/etl/docs/Replication%20in%20Postgres.md?plain=1#L70
-    replication_client
-        .begin_readonly_transaction()
-        .await
-        .unwrap();
-    let slot_info = replication_client
-        .get_or_create_slot(&slot_name)
-        .await
-        .unwrap();
-
-    // Commit the transaction after creating the slot
-    replication_client.commit_txn().await.unwrap();
+    let (replication_client, confirmed_flush_lsn) =
+        create_replication_client_and_slot(&slot_name).await;
 
     // Create CDC stream configuration
     let cdc_config = CdcStreamConfig {
         publication: publication.clone(),
         slot_name: slot_name.clone(),
-        confirmed_flush_lsn: slot_info.confirmed_flush_lsn,
+        confirmed_flush_lsn: confirmed_flush_lsn,
     };
 
     // Create CDC stream (converted events) and attach table schema for conversion
@@ -411,37 +348,18 @@ async fn test_null() {
         ))
         .await
         .unwrap();
-    resources
-        .client()
-        .simple_query(&format!("ALTER TABLE {table_name} REPLICA IDENTITY FULL;"))
-        .await
-        .unwrap();
+    set_replica_identity_full(resources.client(), &table_name).await;
 
     // Publication
-    resources
-        .client()
-        .simple_query(&format!(
-            "CREATE PUBLICATION {publication} FOR TABLE {table_name};"
-        ))
-        .await
-        .unwrap();
+    create_publication_for_table(resources.client(), &publication, &table_name).await;
 
-    let mut replication_client = create_replication_client().await;
-    // https://github.com/supabase/etl/blob/4da956c6b9be8476a1dbe87a4d88689e0671b7c1/etl/docs/Replication%20in%20Postgres.md?plain=1#L70
-    replication_client
-        .begin_readonly_transaction()
-        .await
-        .unwrap();
-    let slot_info = replication_client
-        .get_or_create_slot(&slot_name)
-        .await
-        .unwrap();
-    replication_client.commit_txn().await.unwrap();
+    let (replication_client, confirmed_flush_lsn) =
+        create_replication_client_and_slot(&slot_name).await;
 
     let cdc_config = CdcStreamConfig {
         publication: publication.clone(),
         slot_name: slot_name.clone(),
-        confirmed_flush_lsn: slot_info.confirmed_flush_lsn,
+        confirmed_flush_lsn: confirmed_flush_lsn,
     };
 
     let mut cdc_stream = PostgresSource::create_cdc_stream(replication_client, cdc_config.clone())
--- a/src/moonlink_connectors/src/pg_replicate/tests/cdc_event_test.rs
+++ b/src/moonlink_connectors/src/pg_replicate/tests/cdc_event_test.rs
@@ -235,21 +235,19 @@ async fn test_composite_types() {
 
     assert!(!events.is_empty(), "No CDC events were received");
 
-    use crate::pg_replicate::conversions::cdc_event::CdcEvent;
-
     // the events might include other types of events, eg. BEGIN, PrimaryKeepAlive, etc.
-    // Filter & find insert for id=2
-    let inserted_row = events
-        .iter()
-        .find_map(|e| {
-            if let CdcEvent::Insert((_, row, _)) = e {
-                if matches!(row.values.get(0), Some(Cell::I32(2))) {
-                    return Some(row);
-                }
+    // so using deque to pop the events in the order of execution sequence
+    let mut q: VecDeque<_> = events.into_iter().collect();
+
+    // Expect Insert id=2 first
+    let inserted_row = loop {
+        let ev = q.pop_front().expect("expected insert for id=2");
+        if let CdcEvent::Insert((_, row, _)) = ev {
+            if matches!(row.values.get(0), Some(Cell::I32(2))) {
+                break row;
             }
-            None
-        })
-        .expect("expected insert for id=2");
+        }
+    };
 
     // basic_addr should be a composite (street, city, zip)
     match inserted_row.values.get(1) {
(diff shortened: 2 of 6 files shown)


peterxcli:
??
this should fail with:
```bash
running 1 test
Error in CDC stream: CdcEventConversion(FromBytes(InvalidComposite(MissingParentheses)))
Error in CDC stream: CdcEventConversion(FromBytes(InvalidComposite(MissingParentheses)))

thread 'pg_replicate::tests::cdc_event_test::test_composite_types_in_cdc_stream' panicked at src/moonlink_connectors/src/pg_replicate/tests/cdc_event_test.rs:374:10:
expected insert for id=3
test pg_replicate::tests::cdc_event_test::test_composite_types_in_cdc_stream ... FAILED

failures:

failures:
    pg_replicate::tests::cdc_event_test::test_composite_types_in_cdc_stream

test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 79 filtered out; finished in 5.36s
```

without the patch(handle doubled quotes inside quoted composite/array fields):
```diff
diff --git a/src/moonlink_connectors/src/pg_replicate/conversions/text.rs b/src/moonlink_connectors/src/pg_replicate/conversions/text.rs
index 18be04d..464bb5c 100644
--- a/src/moonlink_connectors/src/pg_replicate/conversions/text.rs
+++ b/src/moonlink_connectors/src/pg_replicate/conversions/text.rs
@@ -383,7 +383,7 @@ impl TextFormatConverter {
         let mut in_quotes = false;
         let mut in_escape = false;
         let mut val_quoted = false;
-        let mut chars = str.chars();
+        let mut chars = str.chars().peekable();
         let mut done = str.is_empty();
 
         while !done {
@@ -395,10 +395,19 @@ impl TextFormatConverter {
                             in_escape = false;
                         }
                         '"' => {
-                            if !in_quotes {
+                            if in_quotes {
+                                // support doubled quotes inside quoted value
+                                if let Some('"') = chars.peek().copied() {
+                                    // consume next quote and append a single quote to value
+                                    let _ = chars.next();
+                                    val_str.push('"');
+                                } else {
+                                    in_quotes = false;
+                                }
+                            } else {
                                 val_quoted = true;
+                                in_quotes = true;
                             }
-                            in_quotes = !in_quotes;
                         }
                         '\\' => in_escape = true,
                         ',' if !in_quotes => {
@@ -454,7 +463,7 @@ impl TextFormatConverter {
         let mut in_quotes = false;
         let mut in_escape = false;
         let mut val_quoted = false;
-        let mut chars = inner.chars();
+        let mut chars = inner.chars().peekable();
         let mut field_iter = fields.iter();
         let mut done = inner.is_empty();
 
@@ -467,10 +476,18 @@ impl TextFormatConverter {
                             in_escape = false;
                         }
                         '"' => {
-                            if !in_quotes {
+                            if in_quotes {
+                                // doubled quotes inside quoted composite field
+                                if let Some('"') = chars.peek().copied() {
+                                    let _ = chars.next();
+                                    val_str.push('"');
+                                } else {
+                                    in_quotes = false;
+                                }
+                            } else {
                                 val_quoted = true;
+                                in_quotes = true;
                             }
-                            in_quotes = !in_quotes;
                         }
                         '\\' if in_quotes => in_escape = true,
                         ',' if !in_quotes => {
@@ -547,7 +564,7 @@ impl TextFormatConverter {
         let mut val_str = String::with_capacity(50);
         let mut in_quotes = false;
         let mut in_escape = false;
-        let mut chars = inner.chars();
+        let mut chars = inner.chars().peekable();
         let mut done = false;
 
         while !done {
@@ -558,11 +575,18 @@ impl TextFormatConverter {
                             val_str.push(c);
                             in_escape = false;
                         }
-                        '\\' if in_quotes => {
-                            in_escape = true;
-                        }
+                        '\\' if in_quotes => in_escape = true,
                         '"' => {
-                            in_quotes = !in_quotes;
+                            if in_quotes {
+                                if let Some('"') = chars.peek().copied() {
+                                    let _ = chars.next();
+                                    val_str.push('"');
+                                } else {
+                                    in_quotes = false;
+                                }
+                            } else {
+                                in_quotes = true;
+                            }
                         }
                         ',' if !in_quotes => {
                             break;
```

at least on my local machine...

peterxcli:
@dentiny could you help me to checkout to this patch and run `cargo test --package moonlink_connectors --lib --features connector-pg`? thanks!

src/moonlink_connectors/src/pg_replicate/tests/cdc_event_test.rs
```diff
@@ -0,0 +1,460 @@
...
+        .await
+        .unwrap();
+    let slot_info = match tokio::time::timeout(
+        Duration::from_secs(10),
+        replication_client.get_or_create_slot(REPLICATION_SLOT),
+    )
+    .await
+    {
+        Ok(Ok(slot_info)) => slot_info,
+        Ok(Err(_)) => return,
```
dentiny: Do we return the whole function early if error?
I would suggest log or unwrap in tests, instead of silently proceed

peterxcli:
> ?? this should fail with:
> 
> ```shell
> running 1 test
> Error in CDC stream: CdcEventConversion(FromBytes(InvalidComposite(MissingParentheses)))
> Error in CDC stream: CdcEventConversion(FromBytes(InvalidComposite(MissingParentheses)))

Sorry this was a pg setting in ci issue and the wrongly early return in test.
Not the error in ci is expected: https://github.com/Mooncake-Labs/moonlink/actions/runs/17018798508/job/48244935785?pr=1525#step:10:372
👍

peterxcli:
Let's get https://github.com/Mooncake-Labs/moonlink/pull/1541 merged frist.
cc @dentiny

dentiny:
> Let's get #1541 merged frist. cc @dentiny

Do you mind rebasing? Thanks!

dentiny:
Ping me when ready, the chaos tests are still pretty broken, feel free to ignore it.

peterxcli:
@dentiny This PR is ready for review, please take another look. Thanks!
But if you feel like this patch can be spitted to 1. TestResources + pg_type_test adapt TestResources. 2. pure cdc event integration test. I'm ok for that.

src/moonlink_connectors/src/pg_replicate/tests/cdc_event_test.rs
```diff
@@ -0,0 +1,419 @@
...
+    let mut schema_client = ReplicationClient::from_client(schema_pg_client);
+    let table_name = TableName {
+        schema: "public".to_string(),
+        name: table_name_str.to_string(),
+    };
+    let src_table_id = schema_client
+        .get_src_table_id(&table_name)
+        .await
+        .unwrap()
+        .expect("missing table id");
```
dentiny:
```suggestion
        .expect("missing table id for table {table_name}");
```

src/moonlink_connectors/src/pg_replicate/tests/cdc_event_test.rs
```diff
@@ -0,0 +1,419 @@
...
+    schema_client
+        .get_table_schema(src_table_id, table_name, Some(publication))
+        .await
+        .unwrap()
```
dentiny: curious what's the purpose for getting table schema here? We're not using it

src/moonlink_connectors/src/pg_replicate/tests/cdc_event_test.rs
```diff
@@ -0,0 +1,419 @@
...
+}
+
+fn spawn_sql_executor(database_url: String) -> mpsc::UnboundedSender<String> {
```
dentiny: Let's add a doc

src/moonlink_connectors/src/pg_replicate/tests/test_utils.rs
```diff
@@ -0,0 +1,155 @@
...
+            eprintln!("Replication connection error: {e}");
+        }
+    });
+    replication_client
+}
+
+async fn drop_publication(client: &tokio_postgres::Client, publication: &str) {
+    let _ = client
+        .simple_query(&format!(
+            "DROP PUBLICATION IF EXISTS {publication} CASCADE;"
```
dentiny:
```suggestion
            "DROP PUBLICATION {publication} CASCADE;"
```
I'm thinking in the test, where we have fully controlled env, should we fail the test if expected publication and table doesn't exist? I'm afraid it could hide problems.

src/moonlink_connectors/src/pg_replicate/tests/test_utils.rs
```diff
@@ -0,0 +1,155 @@
...
+        ))
+        .await
+        .unwrap();
+}
+
+async fn drop_table(client: &tokio_postgres::Client, table_name: &str) {
+    let _ = client
+        .simple_query(&format!("DROP TABLE IF EXISTS {table_name} CASCADE;"))
```
dentiny:
```suggestion
        .simple_query(&format!("DROP TABLE {table_name} CASCADE;"))
```

src/moonlink_connectors/src/pg_replicate/tests/test_utils.rs
```diff
@@ -0,0 +1,155 @@
...
+        // Drop the SQL sender to stop background executor if still running
+        let _ = self.sql_tx.take();
+
+        // Use Arc to share the client without cloning
+        let client = Arc::clone(&self.client);
+        let tables = self.tables.clone();
+        let publications = self.publications.clone();
+        let slots = self.slots.clone();
+        let types = self.types.clone();
+        tokio::spawn(async move {
```
dentiny: I don't think this is ideal for unit test, which should be deterministic and self-contained.
One example you could follow is
https://github.com/Mooncake-Labs/moonlink/blob/86b1549be754345f179e03f4ae638a1e82939208/src/moonlink_backend/tests/common.rs#L104-L113
and you need to use multi-threads for all tests
https://github.com/Mooncake-Labs/moonlink/blob/86b1549be754345f179e03f4ae638a1e82939208/src/moonlink_backend/tests/test_backend.rs#L76

src/moonlink_connectors/src/pg_replicate/tests/cdc_event_test.rs
```diff
@@ -0,0 +1,409 @@
...
+    });
+    tx
+}
+
+#[tokio::test]
+#[serial]
+async fn test_composite_types_in_cdc_stream() {
+    let client = setup_connection().await;
+    // Per-test dynamic names
+    let suffix = std::time::SystemTime::now()
```
dentiny: You have declare the unit test to be executed in serial, so no concurrent access to our pg instance; curious why do we need the random suffix?

src/moonlink_connectors/src/pg_replicate/tests/test_utils.rs
```diff
@@ -0,0 +1,155 @@
...
+async fn drop_table(client: &tokio_postgres::Client, table_name: &str) {
+    let _ = client
+        .simple_query(&format!("DROP TABLE IF EXISTS {table_name} CASCADE;"))
+        .await
+        .unwrap();
+}
+
+async fn drop_type(client: &tokio_postgres::Client, type_name: &str) {
+    let _ = client
+        .simple_query(&format!("DROP TYPE IF EXISTS {type_name} CASCADE;"))
```
dentiny:
```suggestion
        .simple_query(&format!("DROP TYPE {type_name} CASCADE;"))
```

src/moonlink_connectors/src/pg_replicate/tests/test_utils.rs
```diff
@@ -0,0 +1,155 @@
...
+        .await
+        .unwrap();
+}
+
+async fn drop_replication_slot(client: &tokio_postgres::Client, slot_name: &str) {
+    let _ = client
+        .simple_query(&format!(
+            "SELECT pg_drop_replication_slot('{slot}');",
```
dentiny, line 143: it reads `select` statement but you name it `drop_slot`?

src/moonlink_connectors/src/pg_replicate/tests/test_utils.rs
```diff
@@ -0,0 +1,155 @@
...
+        self.types.push(name.into());
+    }
+
+    pub fn set_sql_tx(&mut self, tx: mpsc::UnboundedSender<String>) {
+        self.sql_tx = Some(tx);
+    }
+
+    pub async fn cleanup(&self) {
+        for slot in &self.slots {
+            drop_replication_slot(&self.client, slot).await;
```
dentiny: I'm actually curious, what's the motivation of having both clean up function and drop?
At least we could reuse the same cleanup?

src/moonlink_connectors/src/pg_replicate/tests/test_utils.rs
```diff
@@ -0,0 +1,155 @@
...
+
+    pub fn add_type(&mut self, name: impl Into<String>) {
+        self.types.push(name.into());
+    }
+
+    pub fn set_sql_tx(&mut self, tx: mpsc::UnboundedSender<String>) {
+        self.sql_tx = Some(tx);
+    }
+
+    pub async fn cleanup(&self) {
```
dentiny:
```suggestion
    pub async fn cleanup(self) {
```

src/moonlink_connectors/src/pg_replicate/tests/pg_type_tests.rs
```diff
@@ -59,13 +46,17 @@ async fn test_basic_composite_type() {
         .await
         .unwrap();
 
-    let table_id = get_table_id(&client, "test_basic_composite").await;
+    // Register resources for cleanup
+    resources.add_type("test_address");
+    resources.add_table("test_basic_composite");
```
dentiny: let's make both type and table constants, it's hard to tell what entity it is from their names.

src/moonlink_connectors/src/pg_replicate/tests/pg_type_tests.rs
```diff
@@ -119,8 +112,13 @@ async fn test_nested_composite_types() {
         .await
         .unwrap();
 
-    let table_id = get_table_id(&client, "test_nested").await;
-    let replication_client = ReplicationClient::from_client(client);
+    // Register resources for cleanup
+    resources.add_type("test_point");
```
dentiny: ditto

src/moonlink_connectors/src/pg_replicate/tests/pg_type_tests.rs
```diff
@@ -180,8 +180,12 @@ async fn test_array_of_composite_types() {
         .await
         .unwrap();
 
-    let table_id = get_table_id(&client, "test_array_composite").await;
-    let replication_client = ReplicationClient::from_client(client);
+    // Register resources for cleanup
```
dentiny, line 230: ditto

src/moonlink_connectors/src/pg_replicate/tests/pg_type_tests.rs
```diff
@@ -249,8 +254,13 @@ async fn test_composite_with_nested_array() {
         .await
         .unwrap();
 
-    let table_id = get_table_id(&client, "test_nested_array").await;
-    let replication_client = ReplicationClient::from_client(client);
+    // Register resources for cleanup
+    resources.add_type("test_skill");
```
dentiny: ditto

src/moonlink_connectors/src/pg_replicate/tests/pg_type_tests.rs
```diff
@@ -328,8 +339,15 @@ async fn test_deeply_nested_composites() {
         .await
         .unwrap();
 
-    let table_id = get_table_id(&client, "test_deep").await;
-    let replication_client = ReplicationClient::from_client(client);
+    // Register resources for cleanup
+    resources.add_type("test_coords");
```
dentiny: ditto

src/moonlink_connectors/src/pg_replicate/tests/pg_type_tests.rs
```diff
@@ -415,8 +434,14 @@ async fn test_complex_mixed_nesting() {
         .await
         .unwrap();
 
-    let table_id = get_table_id(&client, "test_mixed").await;
-    let replication_client = ReplicationClient::from_client(client);
+    // Register resources for cleanup
+    resources.add_type("test_tag");
```
dentiny: ditto

src/moonlink_connectors/src/pg_replicate/tests/cdc_event_test.rs
```diff
@@ -0,0 +1,409 @@
...
+                eprintln!("background SQL failed: {e}; sql: {sql}");
+            }
+        }
+    });
+    tx
+}
+
+#[tokio::test]
+#[serial]
+async fn test_composite_types_in_cdc_stream() {
```
dentiny: nitpick: do you mind adding another test case where there's stringified "null", so we could cover the bug fix at https://github.com/Mooncake-Labs/moonlink/pull/1541?
I assume it's much easier to setup in sql level than manually crafted pg types.

src/moonlink_connectors/src/pg_replicate/tests/cdc_event_test.rs
```diff
@@ -0,0 +1,409 @@
...
+             );"
+        ))
+        .await
+        .unwrap();
+
+    let mut replication_client = create_replication_client().await;
+
+    // Create replication slot using the replication client
+    replication_client
+        .begin_readonly_transaction()
```
dentiny: (not a merge blocker) this is sth I'm not sure about, why do we open a read-only transaction, but create a replication slot inside (which looks like a write operation)?

src/moonlink_connectors/src/pg_replicate/tests/cdc_event_test.rs
```diff
@@ -0,0 +1,409 @@
...
+    let table_schema = fetch_table_schema(&publication, &table_name).await;
+    use std::pin::Pin;
+    let mut pinned_stream = Box::pin(cdc_stream);
+    pinned_stream.as_mut().add_table_schema(table_schema);
+
+    // Start a background SQL executor and submit changes via channel
+    let sql_tx = spawn_sql_executor(database_url());
+    resources.set_sql_tx(sql_tx.clone());
+    // Developers can submit arbitrary SQL in desired order with optional delays between them
+    tokio::time::sleep(Duration::from_millis(100)).await;
```
dentiny: Actually why do we need the timeout between executions? we're blocking receiving their cdc events anyway.

src/moonlink_connectors/src/pg_replicate/tests/cdc_event_test.rs
```diff
@@ -0,0 +1,409 @@
...
+            assert!(matches!(fields.get(0), Some(Cell::String(s)) if s == "999 Test St"));
+            assert!(matches!(fields.get(1), Some(Cell::String(s)) if s == "Test City"));
+            assert!(matches!(fields.get(2), Some(Cell::I32(12345))));
+        }
+        other => panic!("unexpected basic_addr cell: {:?}", other),
+    }
+
+    // Optionally validate update for id=1 if present
+    if let Some(updated_row) = events.iter().find_map(|e| {
+        if let CdcEvent::Update((_, _old, new_row, _)) = e {
```
dentiny:
```suggestion
        if let CdcEvent::Update((_, _, new_row, _)) = e {
```

src/moonlink_connectors/src/pg_replicate/tests/cdc_event_test.rs
```diff
@@ -0,0 +1,409 @@
...
+            }
+        }
+    }
+
+    assert!(!events.is_empty(), "No CDC events were received");
+
+    use crate::pg_replicate::conversions::cdc_event::CdcEvent;
+
+    // the events might include other types of events, eg. BEGIN, PrimaryKeepAlive, etc.
+    // Filter & find insert for id=2
```
dentiny: just a nitpick (not a merge blocker), we're searching interesting events every time, so even if events are arriving the incorrect order tests will still pass; might just use a deque and pop front while search.

src/moonlink_connectors/src/pg_replicate/tests/cdc_event_test.rs
peterxcli: Actually I have the exact same question in the prod code:
https://github.com/Mooncake-Labs/moonlink/blob/86b1549be754345f179e03f4ae638a1e82939208/src/moonlink_connectors/src/pg_replicate/clients/postgres.rs#L764-L766

https://github.com/Mooncake-Labs/moonlink/blob/86b1549be754345f179e03f4ae638a1e82939208/src/moonlink_connectors/src/pg_replicate/clients/postgres.rs#L730-L753

I think we should use `BEGIN isolation level repeatable read` at [line 765](https://github.com/Mooncake-Labs/moonlink/blob/86b1549be754345f179e03f4ae638a1e82939208/src/moonlink_connectors/src/pg_replicate/clients/postgres.rs#L765)

dentiny: Sorry I don't have too much context and knowledge here; asked in team internal channel (simply rephrase and copy it here).

Two questions you might be curious about:
- why read-only transaction for write operations?
- why no commit after begin transaction in the same function?

The first answer is: https://github.com/supabase/etl/blob/4da956c6b9be8476a1dbe87a4d88689e0671b7c1/etl/docs/Replication%20in%20Postgres.md?plain=1#L70
The second answers is: due to the way we implement replication and initial copy, we split commit begin and commit in two places, likely will be refactored later
https://github.com/Mooncake-Labs/moonlink/pull/918#discussion_r2218087961

peterxcli:
Status update: addressed review comments and ready for re-review ✅

- Doubled-quote parsing: Implemented for text arrays, composites, and composite arrays in `src/moonlink_connectors/src/pg_replicate/conversions/text.rs` (uses `chars().peekable()` and consumes doubled quotes inside quoted fields).
- CI for Postgres: Workflow config now sets `wal_level=logical` and restarts the container before tests; this resolves the earlier mismatch noted in discussion.
- Integration tests: Added behind `connector-pg` feature in:
  - `src/moonlink_connectors/src/pg_replicate/tests/{test_utils.rs, pg_type_tests.rs, cdc_event_test.rs}`
  - Tests use `TestResources` for dynamic names + cleanup; CDC test covers composite, array-of-composite, nested composite, and arrays inside composites.
- Build/lint: `cargo fmt --check` and `cargo clippy --all-targets --all-features -- -D warnings` are clean.
- Compile check for reviewer ask: `cargo test --package moonlink_connectors --lib --features connector-pg --no-run` succeeds.
- Running the DB-backed tests: I can’t execute them in this environment; CI does via the new `PostgreSQL Integration Tests` job. Locally, run with `DATABASE_URL=postgresql://postgres:postgres@localhost:5432/postgres`.
- Rebase note: Rebased on `main` after #1541; includes the composite parsing fix.

If you prefer, I can split this into:
1) `TestResources` + adapting `pg_type_tests` and
2) CDC event integration test
—but kept together here since both are green and interrelated. Let me know your preference.

Ping @dentiny — ready for another look; ignoring chaos tests per your guidance.

src/moonlink_connectors/src/pg_replicate/tests/cdc_event_test.rs
peterxcli: raised one - https://github.com/Mooncake-Labs/moonlink/issues/1588
for the second Q/A

src/moonlink_connectors/src/pg_replicate/tests/test_utils.rs
peterxcli: This is how `pg_drop_replication_slot` is used.
ref: https://stackoverflow.com/a/30857037

src/moonlink_connectors/src/pg_replicate/tests/cdc_event_test.rs
peterxcli: We have to give CdcStream the schema of that table, otherwise it wont be able to parse the cdc content.
```rust
let table_schema = fetch_table_schema(&publication, &table_name).await;
    use std::pin::Pin;
    let mut pinned_stream = Box::pin(cdc_stream);
    pinned_stream.as_mut().add_table_schema(table_schema);
```

peterxcli: https://github.com/Mooncake-Labs/moonlink/blob/f0606173411498be24bbe87081463348b6fc5e6b/src/moonlink_connectors/src/pg_replicate/postgres_source.rs#L338-L344

https://github.com/Mooncake-Labs/moonlink/blob/f0606173411498be24bbe87081463348b6fc5e6b/src/moonlink_connectors/src/pg_replicate/postgres_source.rs#L363-L366

src/moonlink_connectors/src/pg_replicate/tests/cdc_event_test.rs
peterxcli: yeah, ur right

peterxcli:
@dentiny I've address all of your comments, thanks again for you thorough review, please take another look, thanks!
and sorry for this big patch 😭

src/moonlink_connectors/src/pg_replicate/tests/cdc_event_test.rs
dentiny: My question is less of pg, more of:
- this function is named as get_xxx, which serves for read-only
- if we don't need the return value, why do we call it?

peterxcli: isn't the `schema_client.get_table_schema(src_table_id, table_name, Some(publication)).await.unwrap()` return the value of `fetch_table_schema`?
```rust
async fn fetch_table_schema(publication: &str, table_name_str: &str) -> TableSchema {
    // ...

    // this will return the schema of the given table 
    schema_client
        .get_table_schema(src_table_id, table_name, Some(publication))
        .await
        .unwrap()
}
```

dentiny: Ahhh, sorry I must be mis-reading... sorry!

src/moonlink_connectors/src/pg_replicate/tests/cdc_event_test.rs
```diff
@@ -0,0 +1,511 @@
...
+    let url = database_url();
+    let (schema_pg_client, schema_conn) = connect(&url, NoTls).await.unwrap();
+    tokio::spawn(async move {
+        if let Err(e) = schema_conn.await {
+            eprintln!("Schema connection error: {e}");
+        }
+    });
+    let mut schema_client = ReplicationClient::from_client(schema_pg_client);
+    let table_name = TableName {
+        schema: "public".to_string(),
```
dentiny: nit: make it a constant

src/moonlink_connectors/src/pg_replicate/tests/cdc_event_test.rs
```diff
@@ -0,0 +1,511 @@
...
+}
+
+/// Spawns a background SQL executor that can be used to submit arbitrary SQL in desired order with optional delays between them
+fn spawn_sql_executor(database_url: String) -> mpsc::UnboundedSender<String> {
+    let (tx, mut rx) = mpsc::unbounded_channel::<String>();
+    tokio::spawn(async move {
+        let (bg_client, bg_connection) = connect(&database_url, NoTls).await.unwrap();
+        tokio::spawn(async move {
+            if let Err(e) = bg_connection.await {
+                eprintln!("Background connection error: {e}");
```
dentiny: I'm thinking if we should directly fail it in unit test? Since it could actually block and timeout the whole test suite

src/moonlink_connectors/src/pg_replicate/tests/cdc_event_test.rs
```diff
@@ -0,0 +1,511 @@
...
+            }
+        });
+
+        while let Some(sql) = rx.recv().await {
+            if let Err(e) = bg_client.simple_query(&sql).await {
+                eprintln!("background SQL failed: {e}; sql: {sql}");
```
dentiny: same here

src/moonlink_connectors/src/pg_replicate/tests/cdc_event_test.rs
```diff
@@ -0,0 +1,511 @@
...
+              ARRAY[ROW('789 Pine St', 'Chicago', 60601)::test_address,
+                    ROW('321 Elm St', 'Boston', 02101)::test_address]
+             );"
+        ))
+        .await
+        .unwrap();
+
+    let mut replication_client = create_replication_client().await;
+
+    // Create replication slot using the replication client
```
dentiny: Do you mind if we putting the link besides?
https://github.com/supabase/etl/blob/4da956c6b9be8476a1dbe87a4d88689e0671b7c1/etl/docs/Replication%20in%20Postgres.md?plain=1#L70
just for better clarity

src/moonlink_connectors/src/pg_replicate/tests/cdc_event_test.rs
```diff
@@ -0,0 +1,511 @@
...
+    let table_name = format!("test_cdc_string_null");
+    let publication = format!("test_cdc_string_null_pub");
+    let slot_name = format!("test_cdc_string_null_slot");
+
+    let mut resources = TestResources::new(client);
+    resources.add_table(table_name.clone());
+    resources.add_publication(publication.clone());
+    resources.add_slot(slot_name.clone());
+
+    // Simple schema with text columns to verify stringified "null"
```
dentiny, line 339: quite a few test setups could be extracted into common test utils, including replica identity, creation publication, create slot, etc.

src/moonlink_connectors/src/pg_replicate/tests/test_utils.rs
dentiny: TIL, thank you!

src/moonlink_connectors/src/pg_replicate/tests/cdc_event_test.rs
```diff
@@ -46,15 +47,11 @@ fn spawn_sql_executor(database_url: String) -> mpsc::UnboundedSender<String> {
...
-                eprintln!("Background connection error: {e}");
-            }
+            bg_connection.await.unwrap();
         });
 
         while let Some(sql) = rx.recv().await {
-            if let Err(e) = bg_client.simple_query(&sql).await {
-                eprintln!("background SQL failed: {e}; sql: {sql}");
-            }
+            bg_client.simple_query(&sql).await.unwrap();
```
dentiny: I'm not sure unwrap would print out the sql statement, would appreciate if we use expect here

peterxcli:
Thanks for the review!

src/moonlink_connectors/src/pg_replicate/tests/cdc_event_test.rs
dentiny: FYI: a small followup https://github.com/Mooncake-Labs/moonlink/pull/1626

.github/workflows/ci.yml
```diff
@@ -456,6 +457,18 @@ jobs:
           PGPASSWORD=postgres psql -h localhost -p 5432 -U postgres -d postgres -c "ALTER SYSTEM SET wal_level = 'logical';"
           PGPASSWORD=postgres psql -h localhost -p 5432 -U postgres -d postgres -c "SELECT pg_reload_conf();"
 
+      - name: Restart PostgreSQL to apply wal_level
+        run: |
```
dentiny, line 461: oops, we need to set timeout to all operations; I met situations where minio / fake GCS setup gets stuck.

PR #1526, opened 2025-08-16
[storage] Set parquet fpp after compaction

dentiny:
## Summary

Use the same config as we use in pg_mooncake v0.1 and duckdb default.
https://github.com/Mooncake-Labs/pg_mooncake/blob/d68f1e5f023ba36e0810dbcf7d8deecc275eccc8/src/columnstore/columnstore_table.cpp#L66-L71

https://github.com/duckdb/duckdb/blob/v1.3.2/extension/parquet/parquet_extension.cpp#L232

## Related Issues

Closes https://github.com/Mooncake-Labs/moonlink/issues/525

## Checklist

- [ ] Code builds correctly

### DIFF
--- a/src/moonlink/src/storage/parquet_utils.rs
+++ b/src/moonlink/src/storage/parquet_utils.rs
@@ -8,6 +8,9 @@ const DEFAULT_COMPRESSION: Compression = parquet::basic::Compression::SNAPPY;
 /// Default row group size from duckdb.
 const DEFAULT_ROW_GROUP_SIZE: usize = 122880;
 
+/// Default false positive probability (fpp).
+const DEFAULT_FPP: f64 = 0.01;
+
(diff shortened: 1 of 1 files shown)


dpxcc:
I believe we also need to set ndv but it would require some calculation
> DuckDB aims for fpp=0.01 (https://github.com/duckdb/duckdb/blob/v1.3.2/extension/parquet/parquet_extension.cpp#L232) and hit_ratio=0.01% (https://github.com/duckdb/duckdb/blob/v1.3.2/extension/parquet/parquet_statistics.cpp#L566)

dentiny:
Discussed offline, merge first, will create a separate issue for NDV.

dentiny:
failed chaos test are of known issues

Issue #1527, opened 2025-08-16
[BUG] moonlink hangs when schema is specified


dpxcc:
## Description

```
CREATE EXTENSION pg_mooncake;
CREATE SCHEMA s1;
CREATE TABLE s1.r (a int primary key, b text);
CREATE SCHEMA s2;
CALL mooncake.create_table('s2.c', 's1.r');
SELECT * FROM s2.c; -- hang
```

## Version

- moonlink version (if applicable): [9ca0736](https://github.com/Mooncake-Labs/pg_mooncake/commit/9ca07366d82871eae81111de1c121ffb0bf6a5d3)
- pg_mooncake version (if applicable):

## Expected Behavior

What did you expect to happen?

## Actual Behavior

What actually happened?

## Additional Context

Any other information (logs, screenshots, configs) that might help.

PR #1529, opened 2025-08-16
Attempt to fix blocked read

dentiny:
## Summary

Blocked SQL at pg_mooncake end:
```sql
CREATE EXTENSION pg_mooncake;
CREATE SCHEMA s1;
CREATE TABLE s1.r (a int primary key, b text);
CREATE SCHEMA s2;
CALL mooncake.create_table('s2.c', 's1.r');
SELECT * FROM s2.c; -- hang
```
The reason is when commit = 0 && snapshot = MAX, we consider snapshot unclean

## Related Issues

Closes https://github.com/Mooncake-Labs/moonlink/issues/1527

## Checklist

- [x] Code builds correctly

### DIFF
--- a/src/moonlink/src/union_read/read_state_manager.rs
+++ b/src/moonlink/src/union_read/read_state_manager.rs
@@ -226,7 +226,7 @@ mod tests {
             // Completely no updates for the snapshot.
             Case {
                 snapshot_lsn: NO_SNAPSHOT_LSN,
-                commit_lsn: 0,
+                commit_lsn: NO_COMMIT_LSN,
                 expected: true,
             },
             // All commits are sync-ed to the snapshot.


dpxcc:
Why does this case consistently hang, while the no-schema case works fine?

dentiny:
> Why does this case consistently hang, while the no-schema case works fine?

I don't think it has anything to do with schema, no schema still blocks on my end.
```sql
pg_mooncake (pid: 1636) =# CREATE EXTENSION pg_mooncake;
CREATE EXTENSION
pg_mooncake (pid: 1636) =# CREATE TABLE r (a int primary key, b text);
CREATE TABLE
pg_mooncake (pid: 1636) =# CALL mooncake.create_table('c', 'r');
CALL
pg_mooncake (pid: 1636) =# SELECT * FROM c;
<block>
```

dentiny:
Discussed offline, it's always blocking for a fresh new pg; the fix here is simply consider it clean.

dpxcc:
Discussed offline, the issue was with empty table but not schema:
1. My regular no-schema case always populates data
2. But when I test schema case, I accidentally forgot to populate and triggered the bug

Issue #1528, opened 2025-08-16
Consider using pg database name as iceberg namespace name


dpxcc:
## Description

We currently use `<schema_name>.<table_name>` to identify a table
This can cause collisions, since different databases may contain the same schema and table name

One possible solution is to use pg `database_name` as the Iceberg namespace in moonlink

Right now, we always use `default` as the namespace name. We should consider switching to the database name instead

```sql
CREATE EXTENSION pg_mooncake;
CREATE TABLE r (a int primary key, b text);
CALL mooncake.create_table('c', 'r');
INSERT INTO r VALUES (1, 'a'), (2, 'b'), (3, 'c');
SELECT * FROM c;
CALL mooncake.create_snapshot('c');
```
then
```bash
vscode ➜ /workspaces/pg_mooncake (main) $ tree ~/.pgrx/data-17/pg_mooncake/
/home/vscode/.pgrx/data-17/pg_mooncake/
├── default
│   ├── indicator.text
│   └── public.c
│       ├── data
│       │   ├── 0198b47d-050e-7483-85c3-8a6dee5cd1a4-hash-index-v1-puffin.bin
│       │   ├── data-0198b47d-02fe-75b1-acd7-5d835ca6feb6.parquet
│       │   └── index_block_0198b47d-0302-77e3-bb18-b5f64384e0a6.bin
│       └── metadata
│           ├── 0198b47d-0513-7952-aa17-6a670f832525-m0.avro
│           ├── 0198b47d-0522-7a12-aee4-acf1eada7299-m0.avro
│           ├── snap-9031922700891787243-0-0198b47d-0513-7952-aa17-6a670f832525.avro
│           ├── v0.metadata.json
│           ├── v1.metadata.json
│           └── version-hint.text
├── moonlink_metadata_store.sqlite
├── moonlink.sock
├── public.c
├── read_through_cache
├── temp
│   └── inmemory_public.c_0_26286632.parquet
└── _wal
    └── public.c
        └── metadata_wal.json

10 directories, 14 files
```

dpxcc:
Ideally, when client connects to moonlink_service, it connects to one database/namespace name, and subsequent API calls don't need to specify database/namespace name

dentiny:
On moonlink end, I'm always using mooncake table id as part of the iceberg path, same for wal, which is expected to be unique otherwise non deterministic behavior. 

[...]

PR #1533, opened 2025-08-17
[backend] Rename schema / table to database / table

dentiny:
## Summary

Current `<schema>.<table>` doesn't uniquely identify a table, since it could duplicate across databases;
this PR updates them to `<database>.<schema.table>`.

This PR is supposed to be a no-op change.

## Related Issues

Closes https://github.com/Mooncake-Labs/moonlink/issues/1528

## Checklist

- [ ] Code builds correctly

### DIFF
--- a/src/moonlink_backend/src/lib.rs
+++ b/src/moonlink_backend/src/lib.rs
@@ -107,10 +107,18 @@ impl MoonlinkBackend {
 
     /// Create an iceberg snapshot with the given LSN, return when the a snapshot is successfully created.
     /// If the requested database or table doesn't exist, return [`TableNotFound`] error.
-    pub async fn create_snapshot(&self, schema: String, table: String, lsn: u64) -> Result<()> {
+    pub async fn create_snapshot(
+        &self,
+        mooncake_database: String,
+        mooncake_table: String,
+        lsn: u64,
+    ) -> Result<()> {
         let rx = {
             let mut manager = self.replication_manager.write().await;
-            let mooncake_table_id = MooncakeTableId { schema, table };
+            let mooncake_table_id = MooncakeTableId {
+                mooncake_database,
+                mooncake_table,
+            };
             let writer = manager.get_table_event_manager(&mooncake_table_id)?;
             writer.initiate_snapshot(lsn).await
         };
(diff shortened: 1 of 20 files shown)


dpxcc:
I believe you also need to change [moonlink_rpc](https://github.com/Mooncake-Labs/moonlink/blob/main/src/moonlink_rpc/src/lib.rs#L28) and [moonlink_service](https://github.com/Mooncake-Labs/moonlink/blob/main/src/moonlink_service/src/rpc_server.rs#L80)

[moonlink_datafusion](https://github.com/Mooncake-Labs/moonlink/blob/main/src/moonlink_datafusion/src/table_provider.rs#L218) also needs to be changed, but I will do it myself later

src/moonlink_backend/src/lib.rs
```diff
@@ -107,10 +107,18 @@ impl MoonlinkBackend {
 
     /// Create an iceberg snapshot with the given LSN, return when the a snapshot is successfully created.
     /// If the requested database or table doesn't exist, return [`TableNotFound`] error.
-    pub async fn create_snapshot(&self, schema: String, table: String, lsn: u64) -> Result<()> {
+    pub async fn create_snapshot(
+        &self,
+        mooncake_database: String,
```
dpxcc, line 112: nit: would prefer shorter name, i.e. just `database` and `table` but we could do it in a future PR

dentiny:
> I believe you also need to change [moonlink_rpc](https://github.com/Mooncake-Labs/moonlink/blob/main/src/moonlink_rpc/src/lib.rs#L28) and [moonlink_service](https://github.com/Mooncake-Labs/moonlink/blob/main/src/moonlink_service/src/rpc_server.rs#L80)
> 
> [moonlink_datafusion](https://github.com/Mooncake-Labs/moonlink/blob/main/src/moonlink_datafusion/src/table_provider.rs#L218) also needs to be changed, but I will do it myself later

Thanks for the catch! rpc and service crate should be updated.

src/moonlink_backend/src/lib.rs
dentiny: I could followup with a global grep and sed later.

dentiny:
This PR is a no-op change, just renaming; chaos test failed due to a known issue.

Issue #1540, opened 2025-08-17
[BUG] Wrong complex composite parsing


peterxcli:
## Description

https://github.com/Mooncake-Labs/moonlink/pull/1525#issuecomment-3193803075

## Steps to Reproduce

1. 
2. 
3. 

## Version

- moonlink version (if applicable):
- pg_mooncake version (if applicable):

## Expected Behavior

What did you expect to happen?

## Actual Behavior

What actually happened?

## Additional Context

Any other information (logs, screenshots, configs) that might help.

PR #1541, opened 2025-08-17
fix: wrong complex composite parsing

peterxcli:
## Summary

https://github.com/Mooncake-Labs/moonlink/pull/1525#issuecomment-3193803075

```proto
Event 14: XLogData(XLogDataBody { 
  wal_start: 54529448, 
  wal_end: 54529448, 
  timestamp: 808677494744232, 
  data:
    Update(
      UpdateBody { 
        xid: None, 
        rel_id: 28017, 
        old_tuple: None, 
        key_tuple: None, 
        new_tuple: 
          Tuple(
            [
              Text(b"1"), 
              Text(b"(\"Updated St\",\"Updated City\",54321)"), 
              Text(b"(Home,\"(40.7,-74)\")"), 
              Text(b"(Alice,\"{\"\"(\\\\\"\"123 Main St\\\\\"\",NYC,10001)\"\",\"\"(\\\\\"\"456 Oak Ave\\\\\"\",LA,90210)\"\"}\",\"(Work,\"\"(34,-118)\"\")\")"), 
              Text(b"{\"(\\\"789 Pine St\\\",Chicago,60601)\",\"(\\\"321 Elm St\\\",Boston,2101)\"}")
            ]
          ) 
        }
      ) 
    }
  )
```

```
- Example 1 (quoted NULL in array of composites gets dropped):
  - Input: for an array-of-composites type, element text is "{\"null\"}".
  - Old behavior (parse_composite_array): treated any val_str == "null" as None, ignoring quotes → returned [None].
  - New behavior (via parse_array): quoted "null" is not NULL, so we pass "null" to parse_composite(...) → it errors (invalid composite), which is correct.
  - Why it matters: text arrays already handle quoted "null" as a literal (see existing test). Composite arrays were inconsistent and could silently lose an element.

- Example 2 (consistency with CDC tuple):
  - Element like "{\"(\\\"A St\\\",City,123)}\"":
    - Needs: commas inside the inner composite not to split the outer array, doubled quotes to collapse to a single quote, and no misclassification of values as NULL.
    - Old code and generic array had slightly different NULL handling paths; unifying through parse_array guarantees the same quote/escape/NULL rules as other array types, preventing edge-case divergence.
```

## Related Issues

Closes https://github.com/Mooncake-Labs/moonlink/issues/1540

## Changes

- 
- 
- 

## Checklist

- [x] Code builds correctly

### DIFF
--- a/src/moonlink_connectors/src/pg_replicate/conversions/text.rs
+++ b/src/moonlink_connectors/src/pg_replicate/conversions/text.rs
@@ -896,6 +896,28 @@ mod tests {
             }
             _ => panic!("expected array of composites, got: {cell:?}"),
         }
+
+        let addr_fields = vec![
+            Field::new("street".to_string(), Type::TEXT),
+            Field::new("city".to_string(), Type::TEXT),
+            Field::new("zip".to_string(), Type::INT4),
+        ];
+        let pgoutput_like = r#"{"(\"789 Pine St\",Chicago,60601)","(\"321 Elm St\",Boston,2101)"}"#;
+        let cell = TextFormatConverter::parse_composite_array(pgoutput_like, &addr_fields).unwrap();
+        match cell {
+            Cell::Array(ArrayCell::Composite(composites)) => {
+                assert_eq!(composites.len(), 2);
+                let first = composites[0].as_ref().unwrap();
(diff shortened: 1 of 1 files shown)


peterxcli:
cc @dentiny, PTAL, Thanks!

PR #1564, opened 2025-08-18
deployment: setup CI for standalone deployment

### LINKED ISSUE #1552: deployment: setup CI for standalone deployment
## Summary

Should be a followup PR for https://github.com/Mooncake-Labs/moonlink/pull/1551, to add it to CI.

machichima:
## Summary

Add moonlink service test to CI

## Related Issues

Closes #1552

## Changes

- Setup nginx in CI for running `moonlink_service` standalone test

## Checklist

- [x] Code builds correctly

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -469,19 +469,6 @@ jobs:
     name: Moonlink Service Integration Tests
     runs-on: ubuntu-latest
     
-    services:
-      nginx:
-        image: nginx:alpine
-        ports:
-          - 80:80
-        volumes:
-          - ${{ github.workspace }}/.shared-nginx:/usr/share/nginx/html:ro
-        options: >-
-          --health-cmd "nginx -t"
-          --health-interval 30s
-          --health-timeout 10s
-          --health-retries 3
-
     steps:
       - uses: actions/checkout@v5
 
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -463,3 +463,58 @@ jobs:
...
+    name: Moonlink Service Integration Tests
+    runs-on: ubuntu-latest
+    
+    services:
+      nginx:
+        image: nginx:alpine
+        ports:
+          - 80:80
+        volumes:
+          - /workspaces/moonlink/.shared-nginx:/usr/share/nginx/html:rw
```
dentiny: /workspace/moonlink is the path for devcontainer, not sure it's the same for ci

.github/workflows/ci.yml
```diff
@@ -463,3 +463,58 @@ jobs:
...
+              break
+            fi
+            echo "Waiting for nginx..."
+            sleep 2
+          done
+
+      - name: Run moonlink_service integration tests
+        timeout-minutes: 10
+        run: |
+          cargo test --package moonlink_service --lib --features standalone-test
```
dentiny: I suggest redirect the output to console (at least for now), because the server is running in the background thread, even if it crashes it's still possible for unit tests to pass (completely different than python/c++ thread); enabling the redirection gives us the opportunity to at least manual inspection.
```suggestion
          cargo test --package moonlink_service --lib --features standalone-test -- --nocapture
```

dentiny: maybe also `RUST_BACKTRACE=1` for easier debugging

.github/workflows/ci.yml
machichima: I think we need to explicitly create this directory as this path is hardcoded in the test

https://github.com/Mooncake-Labs/moonlink/blob/58054e3301ce11e923c6481e38a196a3f8b0d5de/src/moonlink_service/src/test.rs#L18

dentiny: You could change the directory name; say, take a env, if it exists then override

src/moonlink_service/src/test.rs
```diff
@@ -15,7 +15,13 @@ use crate::{start_with_config, ServiceConfig, READINESS_PROBE_PORT};
 use moonlink_rpc::{scan_table_begin, scan_table_end};
 
 /// Moonlink backend directory.
-const MOONLINK_BACKEND_DIR: &str = "/workspaces/moonlink/.shared-nginx";
+fn get_moonlink_backend_dir() -> String {
+    if let Ok(workspace) = std::env::var("GITHUB_WORKSPACE") {
```
dentiny: let's give it an explicit name; if we have other env and params in the future, we cannot have another `GITHUB_WORKSPACE`. :)

machichima: Sorry for the delay! I will make the env var name as "MOONLINK_BACKEND_DIR"

dentiny (approved):
Thanks!

PR #1581, opened 2025-08-19
[REST] Fix null field handle

dentiny:
## Summary

When json value input is null, and schema field allows null, we should treat it as null inside moonlink.

## Checklist

- [x] Code builds correctly


dentiny:
All chaos tests failure are due to known issues.

Issue #1518, opened 2025-08-16
Unpin cache if io error occur in read snapshot


peterxcli:
## Summary

https://github.com/Mooncake-Labs/moonlink/pull/1513#pullrequestreview-3125675255

- take_as_read_state tries to read object from remote and pin in the data file cache, that's why we have error status;
- for the current implementation, unwrap directly crashes moonlink, which is definitely bad but we could recover;
- after switching to error propagation, the cache handle is still pinned in cache, so we actually need to unpin before leave.
- add a unit test for error case. The easiest way is to mock IO failure after https://github.com/Mooncake-Labs/moonlink/pull/1335, or give an invalid filepath.

## Context

Why is this issue important? Link any relevant discussions, designs, or documents.

## Tasks

- [ ] 

## Notes

Any extra notes or details that might be useful.

dentiny:
@peterxcli The dependent PR for unit test has been merged.

PR #1335, opened 2025-08-10
refactor: use CacheTrait instead of ObjectStorageCache

CheyuWu:
## Summary

Following the instructions from [Link](https://docs.google.com/document/d/1BN8edRapsVMDomdHJH0GjE9LvBlChCaooNzf7rfahEk/edit?tab=t.0)

However, there is 

## Related Issues

Closes #<issue-number> or links to related issues.

## Changes
- Declare [CacheTrait](https://github.com/Mooncake-Labs/moonlink/blob/14f9165720c6db350eabaa07d4cc6470024e1525/src/moonlink/src/storage/cache/object_storage/base_cache.rs#L28) as Send + Sync
- Update all `&mut self` to `&self` for traits
Example: https://github.com/Mooncake-Labs/moonlink/blob/14f9165720c6db350eabaa07d4cc6470024e1525/src/moonlink/src/storage/cache/object_storage/base_cache.rs#L34
- Update from `ObjectStorageCache` to `Arc<dyn CacheTrait>` everywhere in the codebase
## Checklist

- [ ] Code builds correctly
- [ ] Tests have been added or updated
- [ ] Documentation updated if necessary
- [ ] I have reviewed my own changes

PR #1589, opened 2025-08-19
do cleanup(unpin/delete) before returning error

peterxcli:
## Summary

https://github.com/Mooncake-Labs/moonlink/pull/1513#pullrequestreview-3125675255

- take_as_read_state tries to read object from remote and pin in the data file cache, that's why we have error status;
- for the current implementation, unwrap directly crashes moonlink, which is definitely bad but we could recover;
- after switching to error propagation, the cache handle is still pinned in cache, so we actually need to unpin before leave.
- add a unit test for error case. The easiest way is to mock IO failure after https://github.com/Mooncake-Labs/moonlink/pull/1335, or give an invalid filepath.

## Related Issues

Closes https://github.com/Mooncake-Labs/moonlink/issues/1518

## Changes

- add a unit test to test if the cache handle would be released when error occur
- beside unpin the data cache handle, also unpin puffin file cache handle and remove associated_files when cleaning up
- 

## Checklist

- [x] Code builds correctly

### DIFF
--- a/src/moonlink/src/storage/mooncake_table/snapshot_read_output.rs
+++ b/src/moonlink/src/storage/mooncake_table/snapshot_read_output.rs
@@ -56,6 +56,21 @@ pub struct ReadOutput {
 }
 
 impl ReadOutput {
+    /// Helper to notify evicted files if non-empty.
+    async fn notify_evicted_files(&mut self, files: Vec<String>) {
+        if files.is_empty() {
+            return;
+        }
+        self.table_notifier
+            .as_mut()
+            .unwrap()
+            .send(TableEvent::EvictedFilesToDelete {
+                evicted_files: EvictedFiles { files },
+            })
+            .await
+            .unwrap();
+    }
+
     /// Resolve all remote filepaths and convert into [`ReadState`] for query usage.
     ///
     /// TODO(hjiang): Parallelize download and pin.
@@ -66,7 +81,8 @@ impl ReadOutput {
         // Resolve remote data files.
         let mut resolved_data_files = Vec::with_capacity(self.data_file_paths.len());
         let mut cache_handles = vec![];
-        for cur_data_file in self.data_file_paths.clone().into_iter() {
+        let data_file_paths = std::mem::take(&mut self.data_file_paths);
+        for cur_data_file in data_file_paths.into_iter() {
             match cur_data_file {
                 DataFileForRead::TemporaryDataFile(file) => resolved_data_files.push(file),
                 DataFileForRead::RemoteFilePath((file_id, remote_filepath)) => {
@@ -88,21 +104,10 @@ impl ReadOutput {
                                 resolved_data_files.push(remote_filepath);
                             }
 
-                            if !files_to_delete.is_empty() {
-                                self.table_notifier
-                                    .as_mut()
-                                    .unwrap()
-                                    .send(TableEvent::EvictedFilesToDelete {
-                                        evicted_files: EvictedFiles {
-                                            files: files_to_delete.to_vec(),
-                                        },
-                                    })
-                                    .await
-                                    .unwrap();
-                            }
+                            self.notify_evicted_files(files_to_delete.into_vec()).await;
                         }
                         Err(e) => {
-                            self.handle_resolution_error(&mut cache_handles).await;
+                            self.handle_resolution_error(cache_handles).await;
                             return Err(e);
                         }
                     }
@@ -125,8 +130,8 @@ impl ReadOutput {
     }
 
     /// Handle cleanup and notifications when resolving remote filepaths fails.
-    async fn handle_resolution_error(&mut self, cache_handles: &mut Vec<NonEvictableHandle>) {
-        let mut evicted_files_to_delete_on_error = vec![];
+    async fn handle_resolution_error(&mut self, mut cache_handles: Vec<NonEvictableHandle>) {
+        let mut evicted_files_to_delete_on_error: Vec<String> = vec![];
         // Unpin all previously pinned cache handles before propagating error.
         for mut handle in cache_handles.drain(..) {
             let files_to_delete = handle.unreference().await;
@@ -139,33 +144,10 @@ impl ReadOutput {
             evicted_files_to_delete_on_error.extend(files_to_delete);
         }
 
-        // Best-effort delete any temporary associated files created for this read.
-        if !self.associated_files.is_empty() {
-            self.table_notifier
-                .as_mut()
-                .unwrap()
-                .send(TableEvent::EvictedFilesToDelete {
-                    evicted_files: EvictedFiles {
-                        files: self.associated_files.to_vec(),
-                    },
-                })
-                .await
-                .unwrap();
-        }
-
-        // Best-effort notify deletions.
-        if !evicted_files_to_delete_on_error.is_empty() {
-            self.table_notifier
-                .as_mut()
-                .unwrap()
-                .send(TableEvent::EvictedFilesToDelete {
-                    evicted_files: EvictedFiles {
-                        files: evicted_files_to_delete_on_error.to_vec(),
-                    },
-                })
-                .await
-                .unwrap();
-        }
+        // Include any temporary associated files created for this read, and notify once.
+        evicted_files_to_delete_on_error.extend(std::mem::take(&mut self.associated_files));
+        self.notify_evicted_files(evicted_files_to_delete_on_error)
+            .await;
     }
 }
 


peterxcli:
@dentiny PTAL, Thanks!

src/moonlink/src/storage/cache/object_storage/base_cache.rs
```diff
@@ -28,10 +30,10 @@ pub struct CacheEntry {
 }
 
 #[async_trait]
+#[cfg_attr(test, automock)]
 pub trait CacheTrait: std::fmt::Debug + Send + Sync {
     /// Import cache entry to the cache. If there's no enough disk space, panic directly.
     /// Precondition: the file is not managed by cache.
-    #[must_use]
```
dentiny, line 34: Curious why remove this attribute? I think it's extremely useful to prevent bugs.

src/moonlink/src/storage/mooncake_table/snapshot_read_output.rs
```diff
@@ -65,40 +65,82 @@ impl ReadOutput {
...
+                                                    files: files_to_delete.to_vec(),
+                                                },
+                                            })
+                                            .await;
+                                    }
+                                }
+                            }
+                            Err(e) => {
+                                // Unpin all previously pinned cache handles before propagating error.
+                                for mut handle in cache_handles.drain(..) {
```
dentiny: I'm wondering if we could extract the error handling into a separate function? I want to focus on the happy path when reading.

src/moonlink/src/storage/mooncake_table/snapshot_read_output.rs
```diff
@@ -65,40 +65,82 @@ impl ReadOutput {
...
-                        .get_cache_entry(
-                            file_id,
-                            &remote_filepath,
-                            self.filesystem_accessor.as_ref().unwrap().as_ref(),
-                        )
-                        .await?;
-                    if let Some(cache_handle) = cache_handle {
-                        resolved_data_files.push(cache_handle.get_cache_filepath().to_string());
-                        cache_handles.push(cache_handle);
+                    // If cache or filesystem accessor not provided, fall back to remote filepath.
```
dentiny: curious what's the issue with old implementation? It's an invariant that these two optional fields are assigned? I would prefer the old one since it reduces indentation.

src/moonlink/src/storage/mooncake_table/snapshot_read_output.rs
```diff
@@ -65,40 +65,82 @@ impl ReadOutput {
...
+
+                                if !files_to_delete.is_empty() {
+                                    if let Some(table_notifier) = self.table_notifier.as_mut() {
+                                        let _ = table_notifier
+                                            .send(TableEvent::EvictedFilesToDelete {
+                                                evicted_files: EvictedFiles {
+                                                    files: files_to_delete.to_vec(),
+                                                },
+                                            })
+                                            .await;
```
dentiny: shall we unwrap since it's not expected to fail?

src/moonlink/src/storage/mooncake_table/snapshot_read_output.rs
```diff
@@ -65,40 +65,82 @@ impl ReadOutput {
...
+
+                                // Also unpin any puffin cache handles included in this read output.
+                                for mut handle in self.puffin_cache_handles.drain(..) {
+                                    let files_to_delete = handle.unreference().await;
+                                    evicted_files_to_delete_on_error.extend(files_to_delete);
+                                }
+
+                                // Best-effort delete any temporary associated files created for this read.
+                                for file in self.associated_files.drain(..) {
+                                    let _ = tokio::fs::remove_file(&file).await;
```
dentiny: curious why not have the same handling for all files to delete (I mean, send to channel)?
one benefit is we could check it by listening to the table notifier.

src/moonlink/src/storage/mooncake_table/snapshot_read_output.rs
```diff
@@ -65,40 +65,82 @@ impl ReadOutput {
...
+                                }
+
+                                // Best-effort notify deletions.
+                                if !evicted_files_to_delete_on_error.is_empty() {
+                                    if let Some(table_notifier) = self.table_notifier.as_mut() {
+                                        let _ = table_notifier
```
dentiny: unwrap the error? I don't think send error should happen here.
My concern for ignoring error is it could hide real problems.

src/moonlink/src/storage/mooncake_table/snapshot_read_output.rs
```diff
@@ -117,3 +159,99 @@ impl ReadOutput {
...
+        let read_output = ReadOutput {
+            data_file_paths: vec![
+                DataFileForRead::RemoteFilePath((FAKE_FILE_ID, fake_remote_path.clone())),
+                DataFileForRead::RemoteFilePath((FAKE_FILE_ID, fake_remote_path)),
+            ],
+            puffin_cache_handles: Vec::new(),
+            deletion_vectors: Vec::new(),
+            position_deletes: Vec::new(),
+            associated_files: Vec::new(),
+            table_notifier: None,
```
dentiny: I would suggest passing a table notifier and a temporary file, to see if files to delete meets our expectation.

src/moonlink/src/storage/cache/object_storage/base_cache.rs
peterxcli: I remember it cause clippy check error, but I forgot the reason...

peterxcli: https://github.com/Mooncake-Labs/moonlink/actions/runs/17064091210/job/48377062341?pr=1589#step:7:16
```
error: `#[must_use]` has no effect when applied to a provided trait method
  --> src/moonlink/src/storage/cache/object_storage/base_cache.rs:37:5
   |
37 |     #[must_use]
   |     ^^^^^^^^^^^
   |
   = note: `-D unused-attributes` implied by `-D warnings`
   = help: to override `-D warnings` add `#[allow(unused_attributes)]`

error: `#[must_use]` has no effect when applied to a provided trait method
  --> src/moonlink/src/storage/cache/object_storage/base_cache.rs:46:5
   |
46 |     #[must_use]
   |     ^^^^^^^^^^^

error: `#[must_use]` has no effect when applied to a provided trait method
  --> src/moonlink/src/storage/cache/object_storage/base_cache.rs:55:5
   |
55 |     #[must_use]
   |     ^^^^^^^^^^^
```

dentiny: interesting... I didn't see it before, do you mind adding it to implementation side?

src/moonlink/src/storage/mooncake_table/snapshot_read_output.rs
peterxcli: make sense, thanks.

src/moonlink/src/storage/cache/object_storage/base_cache.rs
peterxcli: I got the same result though...

src/moonlink/src/storage/mooncake_table/snapshot_read_output.rs
```diff
@@ -66,38 +66,45 @@ impl ReadOutput {
...
+                                        evicted_files: EvictedFiles {
+                                            files: files_to_delete.to_vec(),
+                                        },
+                                    })
+                                    .await
+                                    .unwrap();
+                            }
+                        }
+                        Err(e) => {
+                            self.handle_resolution_error(&mut cache_handles).await;
```
dentiny: curious why not transfer the ownership to `handle_resolution_error`?

src/moonlink/src/storage/mooncake_table/snapshot_read_output.rs
```diff
@@ -116,4 +123,245 @@ impl ReadOutput {
...
+        }
+
+        // Best-effort delete any temporary associated files created for this read.
+        if !self.associated_files.is_empty() {
+            self.table_notifier
+                .as_mut()
+                .unwrap()
+                .send(TableEvent::EvictedFilesToDelete {
+                    evicted_files: EvictedFiles {
+                        files: self.associated_files.to_vec(),
```
dentiny:
```suggestion
                        files: std::mem::take(&mut self.associated_files),
```

src/moonlink/src/storage/mooncake_table/snapshot_read_output.rs
```diff
@@ -116,4 +123,245 @@ impl ReadOutput {
...
+        }
+
+        // Best-effort notify deletions.
+        if !evicted_files_to_delete_on_error.is_empty() {
+            self.table_notifier
+                .as_mut()
+                .unwrap()
+                .send(TableEvent::EvictedFilesToDelete {
+                    evicted_files: EvictedFiles {
+                        files: evicted_files_to_delete_on_error.to_vec(),
```
dentiny:
```suggestion
                        files: std::mem::take(&mut evicted_files_to_delete_on_error),
```

src/moonlink/src/storage/mooncake_table/snapshot_read_output.rs
dentiny: why not add them into `evicted_files_to_delete_on_error` and notify together?

src/moonlink/src/storage/mooncake_table/snapshot_read_output.rs
```diff
@@ -66,38 +66,45 @@ impl ReadOutput {
...
-                            })
-                            .await
-                            .unwrap();
+                            if !files_to_delete.is_empty() {
+                                self.table_notifier
+                                    .as_mut()
+                                    .unwrap()
+                                    .send(TableEvent::EvictedFilesToDelete {
+                                        evicted_files: EvictedFiles {
+                                            files: files_to_delete.to_vec(),
```
dentiny:
```suggestion
                                            files: std::mem::take(&mut files_to_delete),
```

src/moonlink/src/storage/mooncake_table/snapshot_read_output.rs
```diff
@@ -66,38 +66,45 @@ impl ReadOutput {
         // Resolve remote data files.
         let mut resolved_data_files = Vec::with_capacity(self.data_file_paths.len());
         let mut cache_handles = vec![];
-        for cur_data_file in self.data_file_paths.into_iter() {
+        for cur_data_file in self.data_file_paths.clone().into_iter() {
```
dentiny: curious do you mind telling me why this change? Thanks

src/moonlink/src/storage/mooncake_table/snapshot_read_output.rs
dentiny:
```suggestion
                                            files: std::mem::take(&mut files_to_delete),
```

src/moonlink/src/storage/mooncake_table/snapshot_read_output.rs
dentiny:
```suggestion
                        files: std::mem::take(&mut self.associated_files),
```

src/moonlink/src/storage/mooncake_table/snapshot_read_output.rs
```diff
@@ -66,38 +66,45 @@ impl ReadOutput {
...
-                            .unwrap()
-                            .send(TableEvent::EvictedFilesToDelete {
-                                evicted_files: EvictedFiles {
-                                    files: files_to_delete.to_vec(),
-                                },
-                            })
-                            .await
-                            .unwrap();
+                            if !files_to_delete.is_empty() {
+                                self.table_notifier
```
dentiny: I think sending deleted files to table notifier take too many lines, maybe we extract it into a util function?

src/moonlink/src/storage/mooncake_table/snapshot_read_output.rs
```diff
@@ -116,4 +123,245 @@ impl ReadOutput {
...
+
+        // Invoke and expect success.
+        let res = read_output
+            .take_as_read_state(Arc::new(|p: String| p))
+            .await;
+        assert!(res.is_ok());
+
+        // Receive exactly one notification and validate its content.
+        if let Some(TableEvent::EvictedFilesToDelete { evicted_files }) = rx.recv().await {
+            assert!(evicted_files
```
dentiny:
```suggestion
            assert_eq!(evicted_files, vec!["old_data_file_cache_file".to_string()]);
```
I think your implementation doesn't check only one element in the vector?

src/moonlink/src/storage/mooncake_table/snapshot_read_output.rs
```diff
@@ -116,4 +123,245 @@ impl ReadOutput {
...
+        // Collect all deletion notifications from the channel and validate contents.
+        let mut notified_file_sets: Vec<Vec<String>> = Vec::new();
+        while let Some(event) = rx.recv().await {
+            if let TableEvent::EvictedFilesToDelete { evicted_files } = event {
+                notified_file_sets.push(evicted_files.files);
+            }
+        }
+
+        // Should contain the associated file from error-path cleanup notification.
+        let expected_associated = get_fake_file_path(&associated_temp_dir);
```
dentiny: I usually don't use any for unit test, since it doesn't guarantee uniqueness.
What making `notified_file_sets` a vector<string> and sort it?

src/moonlink/src/storage/mooncake_table/snapshot_read_output.rs
peterxcli: ah yes, thanks

src/moonlink/src/storage/mooncake_table/snapshot_read_output.rs
peterxcli: Cause it cause some "partial move" complication error:
```rust
borrow of partially moved value: `self`
partial move occurs because `self.data_file_paths` has type `Vec<DataFileForRead>`, which does not implement the `Copy` trait
```
but i will change it to use mem:take to avoid necessary clone call.

src/moonlink/src/storage/mooncake_table/snapshot_read_output.rs
```diff
@@ -116,4 +128,217 @@ impl ReadOutput {
...
+
+        // Also unpin any puffin cache handles included in this read output.
+        for mut handle in self.puffin_cache_handles.drain(..) {
+            let files_to_delete = handle.unreference().await;
+            evicted_files_to_delete_on_error.extend(files_to_delete);
+        }
+
+        // Include any temporary associated files created for this read, and notify once.
+        evicted_files_to_delete_on_error.extend(std::mem::take(&mut self.associated_files));
+        self.notify_evicted_files(evicted_files_to_delete_on_error)
```
dentiny, line 149: just nitpick, not merge blocker: in theory associated files are not "evicted files"

src/moonlink/src/storage/mooncake_table/snapshot_read_output.rs
```diff
@@ -116,4 +128,217 @@ impl ReadOutput {
...
+        let _ = real_cache.try_delete_cache_entry(FAKE_FILE_ID).await;
+
+        // Prepare a separate cache/handle to simulate puffin cache behavior, and
+        // also mark it requested-to-delete so unreference returns files.
+        let puffin_temp_dir = tempdir().unwrap();
+        let mut puffin_cache: ObjectStorageCache = create_infinite_object_storage_cache(
+            &puffin_temp_dir,
+            /*optimize_local_filesystem=*/ false,
+        );
+        let puffin_handle = { import_fake_cache_entry(&puffin_temp_dir, &mut puffin_cache).await };
```
dentiny, line 282:
```suggestion
        let puffin_handle = import_fake_cache_entry(&puffin_temp_dir, &mut puffin_cache).await;
```
curious what's the difference?

src/moonlink/src/storage/mooncake_table/snapshot_read_output.rs
```diff
@@ -116,4 +128,217 @@ impl ReadOutput {
...
+            }
+        }
+
+        // Expect exactly these three files regardless of ordering.
+        let mut notified_files_sorted = notified_files;
+        notified_files_sorted.sort();
+
+        let mut expected_files = vec![
+            get_fake_file_path(&associated_temp_dir),
+            get_fake_file_path(&temp_dir),
```
dentiny, line 337:
```suggestion
            get_fake_file_path(&data_temp_dir),
```

Issue #859, opened 2025-07-14
[ISSUE] Support TLS connection to PG


dpxcc:
## Summary

Briefly describe the task, feature, or problem.

We are currently passing NoTls to make PG connection everywhere

Need to support TLS connection

## Context

Why is this issue important? Link any relevant discussions, designs, or documents.

## Tasks

- [ ] 

## Notes

Any extra notes or details that might be useful.

lohpaul9:
Closed by #1604, but still requires testing with a remote fully SSL-enabled instance. #1632

PR #1604, opened 2025-08-20
pg tls

lohpaul9:
## Summary
Ensures that PG TLS is supported with moonlink. Adds devcontainer script to also regenerate TLS certs on each build and load it to the postgres instance.
## Checklist

- [X] Code builds correctly
- [X] Tests have been added or updated

### DIFF
--- a/.devcontainer/certs/.gitignore
+++ b/.devcontainer/certs/.gitignore
@@ -1,0 +1,4 @@
+*.cnf
+*.key
+*.crt
+*.csr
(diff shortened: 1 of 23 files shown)


lohpaul9:
Closes #859

.devcontainer/certs/.gitignore
```diff
@@ -0,0 +1,4 @@
+*.cnf
```
dentiny, line 1: curious why don't we directly ignore the whole folder? it's never impossible we don't have other cert files under the directory, right?

Issue #1571, opened 2025-08-18
Set NDV value for compacted parquet files.


dentiny:
## Summary

See comment for details: https://github.com/Mooncake-Labs/moonlink/pull/1526#issuecomment-3193858719

fweilun:
I'd like to work on this, thanks!

dentiny:
> I'd like to work on this, thanks!

Thanks for the generous help!

fweilun:
Should I implement a function like `calculate_default_ndv`, which takes the original data as input and outputs a recommended NDV value?

dentiny:
> Should I implement a function like `calculate_default_ndv`, which takes the original data as input and outputs a recommended NDV value?

Yes.

PR #1619, opened 2025-08-20
set ndv value for compacted parquet files

fweilun:
## Summary

set ndv value for compacted parquet files

https://github.com/duckdb/duckdb/blob/0b83e5d2f68bc02dfefde74b846bd039f078affa/extension/parquet/parquet_statistics.cpp#L570-L571

> double n = LossyNumericCast<double>(num_entries);
> double m = -k * n / std::log(1 - std::pow(f, 1 / k));

n is the NDV, which basically is the row count in DuckDB’s implementation.
m is the recommended bitmap size required to achieve the target ffp, but there is no corresponding property in `WriterPropertiesBuilder` options.

Closes [Related Issues](https://github.com/Mooncake-Labs/moonlink/issues/1571)

## Changes
- set ndv to row number

## Checklist

- [✓] Code builds correctly

### DIFF
--- a/src/moonlink/src/storage/compaction/compactor.rs
+++ b/src/moonlink/src/storage/compaction/compactor.rs
@@ -133,7 +133,7 @@ impl CompactionBuilder {
         self.cur_new_data_file = Some(self.create_new_data_file());
         let write_file =
             tokio::fs::File::create(self.cur_new_data_file.as_ref().unwrap().file_path()).await?;
-        let properties = parquet_utils::get_parquet_properties_for_compaction();
+        let properties = parquet_utils::get_parquet_properties_for_compaction(self.cur_row_num);
         let writer: AsyncArrowWriter<tokio::fs::File> =
             AsyncArrowWriter::try_new(write_file, self.schema.clone(), Some(properties))?;
         self.cur_arrow_writer = Some(writer);
--- a/src/moonlink/src/storage/parquet_utils.rs
+++ b/src/moonlink/src/storage/parquet_utils.rs
@@ -22,7 +22,7 @@ pub(crate) fn get_default_parquet_properties() -> WriterProperties {
 }
 
 /// Parquet option for already compacted files.
-pub(crate) fn get_parquet_properties_for_compaction() -> WriterProperties {
+pub(crate) fn get_parquet_properties_for_compaction(row_num: usize) -> WriterProperties {
     let builder = get_default_parquet_properties_builder();
-    builder.set_bloom_filter_enabled(true).build()
+    builder.set_bloom_filter_ndv(row_num as u64).build()
 }


dentiny:
Do you mind rebasing main to resolve conflicts? Thanks!

fweilun:
> Do you mind rebasing main to resolve conflicts? Thanks!

I resolved this one, thanks.

dentiny:
@dpxcc could you please help me review this PR? Thanks!

dpxcc:
There are two issues with the implementation, when comparing to DuckDB..
1. DuckDB only adds bloom filter when dictionary encoding is being used: https://github.com/duckdb/duckdb/blob/v1.3.2/extension/parquet/include/writer/templated_column_writer.hpp#L285
2. DuckDB's ndv is the size of its dictionary, i.e. number of distinct values, not the row count: https://github.com/duckdb/duckdb/blob/v1.3.2/extension/parquet/include/writer/templated_column_writer.hpp#L289

fweilun:
> There are two issues with the implementation, when comparing to DuckDB..
> 
> 1. DuckDB only adds bloom filter when dictionary encoding is being used: https://github.com/duckdb/duckdb/blob/v1.3.2/extension/parquet/include/writer/templated_column_writer.hpp#L285
> 2. DuckDB's ndv is the size of its dictionary, i.e. number of distinct values, not the row count: https://github.com/duckdb/duckdb/blob/v1.3.2/extension/parquet/include/writer/templated_column_writer.hpp#L289

1. From the codebase, I noticed that two get_parquet_properties functions use get_default_parquet_properties_builder(), which enables dictionary encoding, so this seems handled correctly.

2. Is there any recommended way to estimate the number of distinct values? I've checked (this paper mentioned in duckdb)[http://tfk.mit.edu/pdf/bloom.pdf], and it seems that the NDV is a prerequisite. 
 My current understanding is that we’d need to parse the SingleFileToCompact object and iterate through all values using hyperloglog algorithm, but I might be missing something.

Appreciate your guidance!

dpxcc:
1. Are you saying that moonlink always uses dictionary encoding? If so, that’s a bigger issue. We need to fall back to other encodings when the dictionary gets too large. DuckDB does so when the dictionary exceeds [1/5th of row group size](https://github.com/duckdb/duckdb/blob/4768277a98072c9e8900fce3372ee20ed3321149/extension/parquet/parquet_extension.cpp#L219-L223), and it was even [1/100th](https://github.com/duckdb/duckdb/blob/v1.2.0/extension/parquet/parquet_extension.cpp#L194-L195) before
2. Hmm, DuckDB scans the row group and populates the dictionary before constructing the bloom filter, so the dictionary size is known [at that point](https://github.com/duckdb/duckdb/blob/0b83e5d2f68bc02dfefde74b846bd039f078affa/extension/parquet/include/writer/templated_column_writer.hpp#L283). This approach seems more natural to me. Can't we do something similar with arrow-rs?

fweilun:
1.
> Are you saying that moonlink always uses dictionary encoding? 

he default is dictionary encoding if we don’t specify additional settings. IMO, the implementation of compaction only relies on this default setting.

> DuckDB only adds bloom filter when dictionary encoding is being used:

The WriterPropertiesBuilder doesn’t support configuring bloom filters based on the encoding method. It’s tricky to replicate DuckDB’s behavior. I think we should check the settings whenever we use it.

2.
> DuckDB scans the row group and populates the dictionary before constructing the bloom filter

We can probably achieve this by parsing `tokio::fs::File`. Are there any packages in moonlink for this?

dpxcc:
Just to clarify, the issue here isn’t just about diverging from DuckDB’s behavior. It’s that always using dictionary encoding can be pretty inefficient. For example, with an integer column, dictionary encoding can actually double the original uncompressed size, whereas choosing the right encoding typically results in significant compression

In the worst case, we could count the number of distinct values first and only use dictionary encoding when that number is small. The overhead pays for itself for read-heavy workload. I was hoping that arrow-rs would have native support for this

fweilun:
The issue is that Arrow’s compute module doesn’t support distinct value counts for `RecordBatch`. They provide [this function](https://arrow.apache.org/docs/python/generated/pyarrow.compute.unique.html) in Python, but it’s not available in the [Rust compute kernel](https://github.com/apache/arrow-rs/blob/main/arrow/src/compute/mod.rs).

I think we should implement this feature.

dpxcc:
Yes, if arrow-rs doesn’t support this natively, we’ll need to handle it ourselves
The scope of this PR has been expanded, and a new issue has been created for further investigation: https://github.com/Mooncake-Labs/moonlink/issues/1732

pohanhuang:
After discuss with @dentiny: the current writer works in a **streaming / flush-based mode**. It only sees one `RecordBatch` at a time and flushes when hitting the target file size. Because of this, the writer:

* knows the schema and values of the current batch,
* but does **not** know the total number of rows that will come, nor the final NDV for a column across the whole row group/file.

That’s why we can’t pre-decide the “correct” encoding purely at writer creation time. We either need a pre-scan phase (like DuckDB’s `Analyze`/`FinalizeAnalyze`) or accept that NDV decisions are made with only partial knowledge at flush boundaries.

cc: @dpxcc

fweilun:
This issue has been resolved by https://github.com/Mooncake-Labs/moonlink/pull/2031.

Issue #1624, opened 2025-08-20
Avoid crash the rest server when json fail


dentiny:
## Summary

https://github.com/Mooncake-Labs/moonlink/blob/e2957b5a06c2cfd21e5c3a2a4a2b2125a132918b/src/moonlink_service/src/rest_api.rs#L223

fweilun:
I'd like to handle this issue, thanks.

PR #1625, opened 2025-08-20
add error handling for json fail condition

fweilun:
## Summary

Avoid crashing the REST server when JSON serialization fails by replacing `unwrap()` with proper error handling.

## Related Issues

Closes https://github.com/Mooncake-Labs/moonlink/issues/1624

## Checklist

- [✓] Code builds correctly

### DIFF
--- a/src/moonlink_service/src/rest_api.rs
+++ b/src/moonlink_service/src/rest_api.rs
@@ -210,7 +210,20 @@ async fn create_table(
     let arrow_schema = Schema::new(fields);
 
     // Serialization not expect to fail.
-    let serialized_table_config = serde_json::to_string(&payload.table_config).unwrap();
+    let serialized_table_config = match serde_json::to_string(&payload.table_config) {
+        Ok(cfg) => cfg,
+        Err(e) => {
+            error!("serialize table_config failed: {}", e);
+            return Err((
+                StatusCode::INTERNAL_SERVER_ERROR,
+                Json(ErrorResponse {
+                    error: "table_creation_failed".to_string(),
+                    message: format!("Failed to create table: {e}"),
+                }),
+            ));
+        }
+    };
+    // let serialized_table_config = serde_json::to_string(&table_config).unwrap();
 
     // Create table in backend
     match state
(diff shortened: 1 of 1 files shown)


dentiny:
could you please rebase and resolve conflict? thanks!

src/moonlink_service/src/rest_api.rs
```diff
@@ -220,7 +220,19 @@ async fn create_table(
         iceberg_config: None,
     };
     // Serialization not expect to fail.
-    let serialized_table_config = serde_json::to_string(&table_config).unwrap();
+    let serialized_table_config = match serde_json::to_string(&table_config) {
+        Ok(cfg) => cfg,
+        Err(e) => {
+            error!("serialize table_config failed: {}", e);
+            return Err((
+                StatusCode::INTERNAL_SERVER_ERROR,
```
dentiny, line 217: Bad error message, it's the same as table creation

src/moonlink_service/src/rest_api.rs
```diff
@@ -220,7 +220,19 @@ async fn create_table(
         iceberg_config: None,
     };
     // Serialization not expect to fail.
-    let serialized_table_config = serde_json::to_string(&table_config).unwrap();
+    let serialized_table_config = match serde_json::to_string(&table_config) {
+        Ok(cfg) => cfg,
+        Err(e) => {
+            error!("serialize table_config failed: {}", e);
```
dentiny: For error handling, logging / propagation / panic / swallow are the general ways to deal with.
Usually one of them is enough, and here we use propagation;
Logging is fine is we provide extra context, but doesn't seem to necessary in this case.

fweilun: Thanks for the explanation.

src/moonlink_service/src/rest_api.rs
```diff
@@ -210,7 +210,18 @@ async fn create_table(
     let arrow_schema = Schema::new(fields);
 
     // Serialization not expect to fail.
```
dentiny, line 212: we should remove the comment, I will followup in the next PR

dentiny (approved):
Thanks!

Issue #1597, opened 2025-08-19
[BUG] moonlink_service panics when client closes connection


dpxcc:
## Description

Please describe the issue clearly.

https://github.com/Mooncake-Labs/moonlink/blob/f6a7706f7bc9dddba712641dcf0e5a28cd0ae80b/src/moonlink_service/src/rpc_server.rs#L33-L43

After the refactoring, the first arm is not catching closed connection any more, resulting in moonlink_service panics when client closes connection

## Steps to Reproduce

Run anything on pg_mooncake that will connect to moonlink, e.g.
```sql
CREATE EXTENSION pg_mooncake;
CREATE TABLE r (a int primary key, b text);
CALL mooncake.create_table('c', 'r');
INSERT INTO r VALUES (1, 'a'), (2, 'b'), (3, 'c');
SELECT * FROM c;
```
Now Ctrl-C to close client. In pg log (/home/vscode/.pgrx/17.log), we have
```
thread 'tokio-runtime-worker' panicked at /workspaces/pg_mooncake/moonlink/src/moonlink_service/src/rpc_server.rs:42:27:
Unexpected Unix RPC server error: RPC error (permanent) at /workspaces/pg_mooncake/moonlink/src/moonlink_service/src/rpc_server.rs:80:23, source: IO error (permanent) at /workspaces/pg_mooncake/moonlink/src/moonlink_rpc/src/lib.rs:49:5, source: early eof
stack backtrace:
   0: __rustc::rust_begin_unwind
             at /rustc/6b00bc3880198600130e1cf62b8f8a93494488cc/library/std/src/panicking.rs:697:5
   1: core::panicking::panic_fmt
             at /rustc/6b00bc3880198600130e1cf62b8f8a93494488cc/library/core/src/panicking.rs:75:14
   2: moonlink_service::rpc_server::start_unix_server::{{closure}}::{{closure}}
             at /workspaces/pg_mooncake/moonlink/src/moonlink_service/src/rpc_server.rs:42:27
   3: <core::pin::Pin<P> as core::future::future::Future>::poll
             at /usr/local/rustup/toolchains/1.88.0-aarch64-unknown-linux-gnu/lib/rustlib/src/rust/library/core/src/future/future.rs:124:9
   4: tokio::runtime::task::core::Core<T,S>::poll::{{closure}}
[...]

Issue #1431, opened 2025-08-13
Propogate moonlink ErrorStruct to client


dpxcc:
## Summary

Briefly describe the task, feature, or problem.

moonlink_backend APIs can error, e.g. calling `scan_table` on a non-existent table etc

Currently, such errors are unwrapped in moonlink_service, e.g. https://github.com/Mooncake-Labs/moonlink/blob/a5f71e2aef749a6ee4670aee00a030c78bfa24ca/src/moonlink_service/src/rpc_server.rs#L82

We should instead propagate such errors from moonlink_service to moonlink_rpc. Specifically, each moonlink_rpc API should return Result<T, ErrorStruct> instead of T: https://github.com/Mooncake-Labs/moonlink/blob/a5f71e2aef749a6ee4670aee00a030c78bfa24ca/src/moonlink_rpc/src/lib.rs#L28-L35

## Context

Why is this issue important? Link any relevant discussions, designs, or documents.

## Tasks

- [ ] 

## Notes

Any extra notes or details that might be useful.

dpxcc:
This one is of top priority. moonlink_service unwraps the error and kills the connection, client side will only see a connection broken error which is bad. We need to fix it ASAP - since we also need to finish client changes before end of August

PR #1627, opened 2025-08-20
fix: moonlink_service panics when client closes connection

peterxcli:
## Summary

Run anything on pg_mooncake that will connect to moonlink, e.g.

```sql
CREATE EXTENSION pg_mooncake;
SELECT * FROM c;
```

Now Ctrl-C to close client. In pg log (/home/vscode/.pgrx/17.log), we have

```
thread 'tokio-runtime-worker' panicked at /workspaces/pg_mooncake/moonlink/src/moonlink_service/src/rpc_server.rs:42:27:
Unexpected Unix RPC server error: RPC error (permanent) at /workspaces/pg_mooncake/moonlink/src/moonlink_service/src/rpc_server.rs:80:23, source: IO error (permanent) at /workspaces/pg_mooncake/moonlink/src/moonlink_rpc/src/lib.rs:49:5, source: early eof
stack backtrace:
   0: __rustc::rust_begin_unwind
             at /rustc/6b00bc3880198600130e1cf62b8f8a93494488cc/library/std/src/panicking.rs:697:5
   1: core::panicking::panic_fmt
             at /rustc/6b00bc3880198600130e1cf62b8f8a93494488cc/library/core/src/panicking.rs:75:14
   2: moonlink_service::rpc_server::start_unix_server::{{closure}}::{{closure}}
             at /workspaces/pg_mooncake/moonlink/src/moonlink_service/src/rpc_server.rs:42:27
   3: <core::pin::Pin<P> as core::future::future::Future>::poll
             at /usr/local/rustup/toolchains/1.88.0-aarch64-unknown-linux-
....
```

**The root cause is the error is actually a two level depth error - `Error::Rpc(moonlink_rpc::Error(std::io::Error))`. 
The original error handling path can only handle the `moonlink_rpc::Error(std::io::Error)` or `Error::Rpc(std::io::Error)`
So that's why the panic happend.**

## Related Issues

Closes https://github.com/Mooncake-Labs/moonlink/issues/1597

## Changes

- If source is IO directly, means it's the tokio error or service-level IO like `accept/bind/fs`
- All error return from inside of the `handle_stream` are basically PRC error
- Add a test that start a unix server and then close its client to check no panic

## Checklist

- [x] Code builds correctly

### DIFF
--- a/src/moonlink_service/src/rpc_server.rs
+++ b/src/moonlink_service/src/rpc_server.rs
@@ -12,6 +12,37 @@ use tokio::io::{AsyncRead, AsyncWrite};
 use tokio::net::{TcpListener, UnixListener};
 use tracing::info;
 
+fn is_closed_connection(err: &Error) -> bool {
+    // Direct IO error path
+    if let Error::Io(error_struct) = err {
+        if let Some(io_err) = error_struct
+            .source()
+            .and_then(|e| e.downcast_ref::<std::io::Error>())
+        {
+            return matches!(io_err.kind(), BrokenPipe | ConnectionReset | UnexpectedEof);
+        }
+    }
+
+    // RPC wraps an RPC error which can wrap an IO error
+    if let Error::Rpc(error_struct) = err {
+        if let Some(rpc_err) = error_struct
+            .source()
+            .and_then(|e| e.downcast_ref::<moonlink_rpc::Error>())
+        {
+            if let moonlink_rpc::Error::Io(inner) = rpc_err {
+                if let Some(io_err) = inner
+                    .source()
+                    .and_then(|e| e.downcast_ref::<std::io::Error>())
+                {
+                    return matches!(io_err.kind(), BrokenPipe | ConnectionReset | UnexpectedEof);
+                }
+            }
+        }
+    }
+
+    false
+}
+
 /// Start the Unix socket RPC server and serve requests until the task is aborted.
 pub async fn start_unix_server(
     backend: Arc<MoonlinkBackend>,
@@ -31,14 +62,7 @@ pub async fn start_unix_server(
         let backend = Arc::clone(&backend);
         tokio::spawn(async move {
             match handle_stream(backend, stream).await {
-                Err(Error::Rpc(error_struct))
-                    if error_struct
-                        .source()
-                        .and_then(|src| src.downcast_ref::<std::io::Error>())
-                        .map(|io_err| {
-                            matches!(io_err.kind(), BrokenPipe | ConnectionReset | UnexpectedEof)
-                        })
-                        .unwrap_or(false) => {}
+                Err(e) if is_closed_connection(&e) => {}
                 Err(e) => panic!("Unexpected Unix RPC server error: {e}"),
                 Ok(()) => {}
             }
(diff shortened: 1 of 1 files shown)


src/moonlink_service/src/rpc_server.rs
```diff
@@ -3,7 +3,7 @@ use arrow_ipc::writer::StreamWriter;
 use moonlink_backend::MoonlinkBackend;
 use moonlink_rpc::{read, write, Request, Table};
 use std::collections::HashMap;
-use std::error::Error as StdError;
+use std::error::Error as _;
```
dentiny, line 6: doesn't seem to be necessary?

src/moonlink_service/src/rpc_server.rs
```diff
@@ -12,6 +12,39 @@ use tokio::io::{AsyncRead, AsyncWrite};
...
+fn is_disconnect(io: &std::io::Error) -> bool {
+    matches!(io.kind(), BrokenPipe | ConnectionReset | UnexpectedEof)
+}
+
+fn is_closed_connection(err: &Error) -> bool {
+    // Direct IO error path
+    if let Error::Io(error_struct) = err {
+        if let Some(io_err) = error_struct
+            .source()
+            .and_then(|e| e.downcast_ref::<std::io::Error>())
```
dentiny, line 24: curious what if we have multiple layers of error?

src/moonlink_service/src/rpc_server.rs
peterxcli: yeah, just a drive-by cleanup

src/moonlink_service/src/rpc_server.rs
peterxcli: you prefer recursive error matching?

peterxcli: It have covered all cases, If there are any layers/cases it need to cover in the future, then we can switch to use recursion.

dentiny: I'm afraid it's easy to forget --- when people add new crates and new error structs, it's not straightforward to recall they need to update rpc server. Then we meet the same error again. :(

If you can add certain assertion to prevent, it's also fine.

dpxcc: This code is meant to only catch closed connection errors from rpc clients:
https://github.com/Mooncake-Labs/moonlink/blob/034ed00ff262648197cd56a45ee47edeee41004a/src/moonlink_rpc/src/lib.rs#L42-L43
https://github.com/Mooncake-Labs/moonlink/blob/034ed00ff262648197cd56a45ee47edeee41004a/src/moonlink_rpc/src/lib.rs#L49
https://github.com/Mooncake-Labs/moonlink/blob/034ed00ff262648197cd56a45ee47edeee41004a/src/moonlink_rpc/src/lib.rs#L52
All other errors should be surfaced back to rpc clients once https://github.com/Mooncake-Labs/moonlink/issues/1431 is done

So it's always fixed error struct that we need to catch

dentiny (approved):
Thanks @peterxcli for the quick fix! It's urgently needed so I plan to merge it for now.

There're several things:
- (preferred) we stream back backend error (i.e., table creation error) to client side, so the error returned by `handle_stream` is always IO error
- For error identification, I'm thinking of a recursively way to get the most inner error, like backend error -> moonlink error -> IO error -> check if connection error

Issue #1630, opened 2025-08-20
moonlink connector use error struct in all places


dentiny:
## Summary

https://github.com/Mooncake-Labs/moonlink/blob/69f6b4c9b2aba1bb9455084e791963ecdadb4b50/src/moonlink_connectors/src/error.rs#L37-L47

https://github.com/Mooncake-Labs/moonlink/blob/69f6b4c9b2aba1bb9455084e791963ecdadb4b50/src/moonlink_connectors/src/error.rs#L53-L59

fweilun:
I'd like to work on this, thanks!

dentiny:
Thanks!

PR #1641, opened 2025-08-21
Refine error propagation for all types in moonlink_connector

fweilun:
## Summary

Refactor error propagation in moonlink_connector by unifying all error types through function-based mapping.

## Related Issues

Closes https://github.com/Mooncake-Labs/moonlink/issues/1630

## Checklist

- [✓] Code builds correctly

### DIFF
--- a/src/moonlink_connectors/src/error.rs
+++ b/src/moonlink_connectors/src/error.rs
@@ -36,32 +36,33 @@ pub enum Error {
     MpscChannelSendError(ErrorStruct),
 
     // Requested database table not found.
-    #[error("Table {0} not found")]
-    TableNotFound(String),
+    #[error("{0}")]
+    TableNotFound(ErrorStruct),
 
     // Invalid source type for operation.
-    #[error("Invalid source type: {0}")]
-    InvalidSourceType(String),
+    #[error("{0}")]
+    InvalidSourceType(ErrorStruct),
 
     // Table replication error: duplicate table.
     #[error("Duplicate table added to replication: {0}")]
     ReplDuplicateTable(String),
 
     // REST API error.
-    #[error("REST API error: {0}")]
-    RestApi(String),
+    #[error("{0}")]
+    RestApi(ErrorStruct),
 
     // REST source error.
     #[error("{0}")]
     RestSource(ErrorStruct),
 
     // REST source error: duplicate source table to add.
-    #[error("REST source error: duplicate source table to add with table id {0}")]
-    RestDuplicateTable(SrcTableId),
+    #[error("{0}")]
+    RestDuplicateTable(ErrorStruct),
 
     // REST source error: non-existent source table to remove.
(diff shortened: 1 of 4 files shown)


src/moonlink_connectors/src/error.rs
```diff
@@ -74,6 +74,70 @@ pub enum Error {
...
+        Error::ReplDuplicateTable(ErrorStruct {
+            message: format!("Duplicate table added to replication: {table_name}"),
+            status: ErrorStatus::Permanent,
+            source: None,
+            location: Some(Location::caller().to_string()),
+        })
+    }
+
+    #[track_caller]
+    pub fn rest_api(e: String) -> Self {
```
dentiny: let's take source error as well, which provides structured error
```suggestion
    pub fn rest_api(err_msg: &str, err: Error) -> Self {
```

fweilun: There are only two possible sources of errors: `SendError<RestCommand>` and `SendError<TableEvent>`. Since neither of them can be converted into our Error type, I think it’s unnecessary to include them as sources?

dentiny: oh sorry I actually mean a `Arc<anyhow::Error>`

fweilun: I added anyhow::Error, but TableEvent can’t be converted because it doesn’t implement the Sync trait.

hhcs9527 (approved):
LGTM

src/moonlink_connectors/src/error.rs
```diff
@@ -36,32 +37,32 @@ pub enum Error {
...
-    #[error("Table {0} not found")]
-    TableNotFound(String),
+    #[error("{0}")]
+    TableNotFound(ErrorStruct),
 
     // Invalid source type for operation.
-    #[error("Invalid source type: {0}")]
-    InvalidSourceType(String),
+    #[error("{0}")]
+    InvalidSourceType(ErrorStruct),
```
dentiny, line 45: Actually it's not used anywhere (the naming is not clear as well)

src/moonlink_connectors/src/error.rs
```diff
@@ -3,6 +3,7 @@ use crate::pg_replicate::postgres_source::{
 };
 use crate::rest_ingest::rest_source::RestSourceError;
 use crate::rest_ingest::{json_converter, SrcTableId};
+use anyhow;
```
dentiny, line 6: curious what's the purpose for `use anyhow` here? since we only use `anyhow::Error` and `anyhow::anyhow!`, the later is not used here

fweilun: Sorry, I didn't notice that. I thought these kind of errors would be handled by pre-commit?

dentiny: Yeah I think it should, even compilation should capture unused import. Sorry not sure about that

Issue #1638, opened 2025-08-21
BUG: support recovery for REST API


dentiny:
## Summary

```sh
thread 'main' panicked at /home/ubuntu/moonlink/src/moonlink_connectors/src/replication_manager.rs:195:9:
assertion failed: self.connections.contains_key(src_uri)
stack backtrace:
   0: __rustc::rust_begin_unwind
   1: core::panicking::panic_fmt
   2: core::panicking::panic
   3: moonlink_connectors::replication_manager::ReplicationManager<T>::start_replication::{{closure}}
   4: moonlink_service::start_with_config::{{closure}}
   5: tokio::runtime::park::CachedParkThread::block_on
   6: tokio::runtime::context::runtime::enter_runtime
   7: tokio::runtime::runtime::Runtime::block_on
   8: moonlink_service::main
note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.
```

PR #1643, opened 2025-08-21
[REST] Implement rest table recovery

dentiny:
## Summary

This PR implements recovery for rest ingestion table, it currently only works for content persisted into iceberg.

## Related Issues

Closes https://github.com/Mooncake-Labs/moonlink/issues/1638

## Checklist

- [x] Code builds correctly


src/moonlink_connectors/src/replication_manager.rs
```diff
@@ -161,11 +161,18 @@ impl<T: Clone + Eq + Hash + std::fmt::Display> ReplicationManager<T> {
 
     /// Initialize event API connection for data ingestion.
     /// Returns the event request sender channel for the API to use.
-    pub async fn initialize_event_api(
+    pub async fn initialize_event_api_for_once(
```
zhousun, line 164: why do we need to call it explicitly?
when we already have:
if config.rest_api_port.is_some() {
backend.initialize_event_api().await?;
}

src/moonlink_backend/src/recovery_utils.rs
```diff
@@ -40,6 +83,31 @@ async fn recover_table(
...
 }
 
+/// Recovery the given table.
+async fn recover_table(
+    backend_attributes: BackendAttributes,
+    metadata_entry: TableMetadataEntry,
+    replication_manager: &mut ReplicationManager<MooncakeTableId>,
+    read_state_filepath_remap: ReadStateFilepathRemap,
+) -> Result<()> {
+    // Table created by REST API doesn't support recovery.
```
zhousun, line 93: remove

src/moonlink_connectors/src/replication_manager.rs
dentiny: In normal use case, its always service initiator who initiate event api

But no such role for unit tests

PR #1644, opened 2025-08-21
[storage] Fix file deletion for file upload

dentiny:
## Summary

Before REST file upload API, there're only two types of files in mooncake snapshot: write-through caches and iceberg files.
The former will be deleted if (1) they get uploaded to iceberg table; (2) iceberg table lives on local filesystem; (3) unpinned.

After we introduce file upload API, we have third type of file: user uploaded files, which serves as the counterpart for write-through cache and will attempted to be deleted if the above three requirements met.
This PR only attempts deletion only if they're local files.

## Checklist

- [x] Code builds correctly

### DIFF
--- a/src/moonlink/src/storage/io_utils.rs
+++ b/src/moonlink/src/storage/io_utils.rs
@@ -1,10 +1,14 @@
 use crate::Result;
 
 /// Util function to delete local file in parallel.
+/// If the file doesn't exists on local filesystem, ignore and proceed.
 pub(crate) async fn delete_local_files(local_files: &[String]) -> Result<()> {
-    let delete_futures = local_files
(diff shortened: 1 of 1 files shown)


dpxcc:
> After we introduce file upload API, we have third type of file: user uploaded files, which serves as the counterpart for write-through cache and will attempted to be deleted if the above three requirements met.
This PR only attempts deletion only if they're local files.

Don't we copy user uploaded files to write through cache?

dpxcc:
As discussed offline, we are not copying user-uploaded files to our own storage, only temporarily for this POC build
In the long term, we should always copy user-uploaded files into our own storage so that no special cases are required

PR #1657, opened 2025-08-22
[storage] Get table cardinality

dentiny:
## Summary

This PR exposes cardinality to table status, so we could leverage to optimize.
Return rows count includes:
- all row counts, it's worth noting deleted rows are not accounted
- all committed records, even not persisted, is accounted

## Checklist

- [x] Code builds correctly

### DIFF
--- a/src/moonlink/src/storage/mooncake_table/table_status.rs
+++ b/src/moonlink/src/storage/mooncake_table/table_status.rs
@@ -7,6 +7,8 @@ pub struct TableSnapshotStatus {
     pub commit_lsn: u64,
     /// Iceberg flush LSN.
     pub flush_lsn: Option<u64>,
+    /// Cardinality.
+    pub cardinality: u64,
     /// Iceberg warehouse location.
     pub iceberg_warehouse_location: String,
 }
--- a/src/moonlink/src/storage/mooncake_table/table_status.rs
+++ b/src/moonlink/src/storage/mooncake_table/table_status.rs
@@ -3,12 +3,12 @@ use serde::{Deserialize, Serialize};
 
 #[derive(Clone, Debug, Deserialize, PartialEq, Serialize)]
 pub struct TableSnapshotStatus {
+    /// Cardinality.
+    pub cardinality: u64,
     /// Mooncake table commit LSN.
     pub commit_lsn: u64,
     /// Iceberg flush LSN.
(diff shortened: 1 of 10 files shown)


src/moonlink/src/storage/mooncake_table/table_status.rs
```diff
@@ -7,6 +7,8 @@ pub struct TableSnapshotStatus {
     pub commit_lsn: u64,
     /// Iceberg flush LSN.
     pub flush_lsn: Option<u64>,
+    /// Cardinality.
+    pub cardinality: u64,
```
dpxcc: nit: can we move cardinality to the first field

src/moonlink_rpc/src/lib.rs
```diff
@@ -61,5 +61,6 @@ pub struct Table {
     pub table: String,
     pub commit_lsn: u64,
     pub flush_lsn: Option<u64>,
+    pub cardinality: u64,
```
dpxcc: nit: same, move after table but before commit_lsn

dpxcc (approved):

src/moonlink_backend/src/lib.rs
```diff
@@ -235,6 +235,7 @@ impl MoonlinkBackend {
                 table: mooncake_table_id.table.clone(),
                 commit_lsn: table_snapshot_status.commit_lsn,
                 flush_lsn: table_snapshot_status.flush_lsn,
+                cardinality: table_snapshot_status.cardinality,
```
dpxcc, line 238: your order is still not consistent, esp when you are assigning them in test code..
but the public interface in moonlink_rpc is fixed, so accept for now

src/moonlink/src/storage/mooncake_table/table_status.rs
dentiny: updated

src/moonlink_rpc/src/lib.rs
dentiny: updated

src/moonlink_backend/src/lib.rs
dentiny: yes I only reorder the struct but not assignment

dpxcc: but this is a struct that's also inconsistent..

PR #1670, opened 2025-08-22
[REST] Provide sync mode for write operations

dentiny:
## Summary

To implement `create_snapshot`, client-side needs to have LSN concept; one possible implementation is block wait operation completion and return back LSN for the completed insertions.

Current REST API server does everything async, which means client response 200 only means request has been queued at server side, but never indicates success.
In this PR, I add a SYNC mode to block wait its completion.

We could also add timeout and retry mechanism.

## Checklist

- [ ] Code builds correctly

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4306,6 +4306,7 @@ dependencies = [
  "iceberg",
  "moonlink",
  "moonlink_error",
+ "more-asserts",
  "native-tls",
  "num-traits",
  "parquet",
--- a/src/moonlink_backend/src/lib.rs
+++ b/src/moonlink_backend/src/lib.rs
@@ -11,9 +11,11 @@ pub use error::{Error, Result};
 use mooncake_table_id::MooncakeTableId;
 pub use moonlink::ReadState;
 use moonlink::{ReadStateFilepathRemap, TableEventManager};
-pub use moonlink_connectors::rest_ingest::rest_source::{
+pub use moonlink_connectors::rest_ingest::event_request::{
     EventRequest, FileEventOperation, FileEventRequest, RowEventOperation, RowEventRequest,
 };
+pub use moonlink_connectors::rest_ingest::rest_event::RestEvent;
+pub use moonlink_connectors::rest_ingest::rest_source::RestSource;
 use moonlink_connectors::ReplicationManager;
 pub use moonlink_connectors::REST_API_URI;
 use moonlink_metadata_store::base_metadata_store::MetadataStoreTrait;
--- a/src/moonlink_backend/tests/test_backend.rs
+++ b/src/moonlink_backend/tests/test_backend.rs
@@ -9,11 +9,9 @@ mod tests {
         current_wal_lsn, get_serialized_table_config, smoke_create_and_insert, TestGuard,
         TestGuardMode, DATABASE, TABLE,
     };
-    use moonlink_backend::EventRequest;
-    use moonlink_backend::MoonlinkBackend;
-    use moonlink_backend::{
-        table_status::TableStatus, RowEventOperation, RowEventRequest, REST_API_URI,
-    };
+    use moonlink_backend::RowEventOperation;
+    use moonlink_backend::{table_status::TableStatus, REST_API_URI};
+    use moonlink_backend::{EventRequest, MoonlinkBackend, RowEventRequest};
     use moonlink_metadata_store::{base_metadata_store::MetadataStoreTrait, SqliteMetadataStore};
 
     use arrow::datatypes::Schema as ArrowSchema;
@@ -464,6 +462,7 @@ mod tests {
                 "age": 30
             }),
             timestamp: SystemTime::now(),
+            tx: None,
         };
         let rest_event_request = EventRequest::RowRequest(row_event_request);
         backend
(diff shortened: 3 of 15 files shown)


dpxcc:
We don't really support async mode, like there are no way to see all ongoing async jobs, polling the result of async jobs etc
For this issue, the fix should simply just be to make RestAPI waits for its operation
Don't complicate the stack unnecessarily. If we were to introduce async mode in the future, it would need proper work to support it, and for both RestAPI and RPC

dentiny:
> We don't really support async mode, like there are no way to see all ongoing async jobs, polling the result of async jobs etc For this issue, the fix should simply just be to make RestAPI waits for its operation Don't complicate the stack unnecessarily. If we were to introduce async mode in the future, it would need proper work to support it, and for both RestAPI and RPC

No everything in async mode

zhousun:
@dpxcc Actually we do, event api (espeically when called through RestAPI) are meant to be async (remember they are usually called by web-apps who won't wait for completion). It is very important this is the default behavior.
And it is common to have an argument 'wait_for_response' to wait until the role is written, which is what's implemented in the current diff.

dpxcc:
When a web-app makes an http request, it must wait for the response. Whether the request is made synchronously or asynchronously is orthogonal - it only determines whether the web-app can do other work while waiting. Ultimately, it's the server implementation that decides whether to respond immediately after queuing the request (async mode) or after completing the task (sync mode); both approaches are valid

However, there are two issues, as mentioned in my previous comment:

1. moonlink currently lacks proper async mode support. To implement async mode correctly, users should be able to list ongoing tasks and poll for results. In async mode, the http response code when the server queues the request should also be 202 (Accepted), not 200 (OK)

2. Async support should be consistent across RPC and REST APIs. What I dislike about this diff is that it introduces further divergence between the two

dentiny:
> When a web-app makes an http request, it must wait for the response. Whether the request is made synchronously or asynchronously is orthogonal - it only determines whether the web-app can do other work while waiting. Ultimately, it's the server implementation that decides whether to respond immediately after queuing the request (async mode) or after completing the task (sync mode); both approaches are valid
> 
> However, there are two issues, as mentioned in my previous comment:
> 
> 1. moonlink currently lacks proper async mode support. To implement async mode correctly, users should be able to list ongoing tasks and poll for results. In async mode, the http response code when the server queues the request should also be 202 (Accepted), not 200 (OK)
> 2. Async support should be consistent across RPC and REST APIs. What I dislike about this diff is that it introduces further divergence between the two

Rpc has no data plane support for now, only control plane

dpxcc:
But async vs sync is not limited to ingestion, it also applies to other APIs, e.g. create_table, create_snapshot, optimize_table etc

dentiny:
> But async vs sync is not limited to ingestion, it also applies to other APIs, e.g. create_table, create_snapshot, optimize_table etc

I didnt change anything in the control plane in this pr

dpxcc:
Then you are just chase fixing

dentiny:
> Then you are just chase fixing

Could you please elaborate? Fixing what?

dpxcc:
This is a typical chase fix. You want to add create_snapshot, but it requires an LSN in order to test correctness. Okay, let's work around it by introducing a sync mode to writes, and then make sure tests use that mode to obtain an LSN. But then
How should create_snapshot work when writes are async?
How should async mode be handled more generally?
How do we resolve the divergence between the RPC and REST API?
Good luck to future contributors

dentiny:
> How should create_snapshot work when writes are async?
How should async mode be handled more generally?

It's intentional, async mode is expected to be best effort.
Sync mode gets LSN back and guarantees recovery.

> How do we resolve the divergence between the RPC and REST API?

I don't see any divergence between these two sets of APIs; 
as I mentioned, RPC is supposed to work on control plane, while REST = connector / data plane + control plane.

If you do care about API consistency, we should have better abstraction on connection, rather than RPC.

dpxcc:
> It's intentional, async mode is expected to be best effort.
Sync mode gets LSN back and guarantees recovery.

Are you suggesting we test mostly in sync mode but deploy in production with async mode? That’s a red flag

Also, you still haven’t addressed the fact that the current code has no proper async support, e.g. listing ongoing async jobs or polling their results

> I don't see any divergence between these two sets of APIs;
as I mentioned, RPC is supposed to work on control plane, while REST = connector / data plane + control plane.

That’s just incorrect. RPC is already used to manage connectors in create_table, and we will also ingest data using RPC to support event tables in pg_mooncake

dentiny:
> Are you suggesting we test mostly in sync mode but deploy in production with async mode? That’s a red flag

Not sure what do you mean? Core implementation is the same.

> Also, you still haven’t addressed the fact that the current code has no proper async support, e.g. listing ongoing async jobs or polling their results

I think I already answered: 
> It's intentional, async mode is expected to be best effort.

in other words I'm thinking of fire-and-forget style.

If people needs synchronization, they should use sync mode.

dpxcc:
> Not sure what do you mean? Core implementation is the same

You can't test in one mode but deploy in another mode. That's why bugs are undiscovered

> I think I already answered:

That doesn't answer my question. People need to list ongoing async jobs and polling their results in async mode, you can't just say use sync mode..

Issue #1673, opened 2025-08-22
[ISSUE] Optimize `CdcStream`


nbiscaro:
## Summary

Leverage the batching API we implemented in rust-postgres to pull batches of messages off the channel.

## Context

Why is this issue important? Link any relevant discussions, designs, or documents.

## Tasks

- [ ] 

## Notes

Any extra notes or details that might be useful.

PR #1677, opened 2025-08-22
Expose batched-streaming API to `CdcStream`

nbiscaro:
## Summary

Exposes new batch-based API from the underlying stream: https://github.com/Mooncake-Labs/rust-postgres/pull/12

The previous moonlink CDC stream awaited and parsed one backend frame at a time, forcing a wakeup for every message which inflated scheduler churn and produced long poll times under load. 

`CdcStream` now provides `next_batch_msgs(out, max)`, and the replication loop drains up to `MAX_EVENTS_PER_WAKE` in a single await, converts in place, and pushes successful items into an ok_events queue. We also dedup per-table schema changes before fetching and applying them, updating the local schema map once per table. 

Batching collapses wakeups and cuts per-message scheduling and allocation overhead. The attached [tokio-console](https://github.com/tokio-rs/console) graphs illustrate this performance improvement. In each case we do a 60s run of moonlink ingesting 100M appends:

Previously, 
_about 211k wakes_ with 99% self-wakes and poll latency around 
p50 ≈ 23ms, 
p95 ≈ 97ms, 
p99 ≈ 158ms, 
scheduling times around
p50 ≈ 46ms, 
p95 ≈ 71ms, 
p99 ≈ 181ms, 

with a broad, multi-modal histogram. 

Screenshot 2025-08-22 at 9 38 26 AM

Currently, 
_total wakes drop to roughly 9.8k_ with 71% self-wakes and a much tighter poll distribution at 
p50 ≈ 8.7ms, 
p95 ≈ 11.9ms, 
p99 ≈ 18.5ms. 
and scheduling times around
p50 ≈ 14.0ms, 
p95 ≈ 25.8ms, 
p99 ≈ 35.8ms. 

Busy time rises from ~86% to ~92%, 
Idle time drops from ~14% to 7.7%

indicating less time lost to scheduler overhead, and more time doing actual work. 

Screenshot 2025-08-22 at 9 44 42 AM

As an experiment, changing the table handler [mpsc channel](https://github.com/Mooncake-Labs/moonlink/blob/266a7f6514b585ce86942e70a8c3e9535a7a909b/src/moonlink/src/table_handler.rs#L30) to unbounded shows its size balloon to around 300k events at its max, implying moonlink cannot consume fast enough from the channel that the `CdcStream` is appending to! Of course, in the actual implementation we keep this channel bounded and will apply back-pressure when full. In order to fully take advantage of the end-to-end performance gain we should look next to moonlink and see how we can drain this channel faster. 

References:

Tokio-console blog: https://hegdenu.net/posts/task-scheduled-time-in-console/?utm_source=chatgpt.com
Waker: https://docs.rs/futures/latest/futures/task/struct.Waker.html

## Related Issues

Closes https://github.com/Mooncake-Labs/moonlink/issues/1673

## Changes

- 
- 
- 

## Checklist

- [ ] Code builds correctly


src/moonlink_connectors/src/pg_replicate/moonlink_sink.rs
```diff
@@ -716,4 +722,42 @@ mod tests {
             matches!(rx.recv().await.unwrap(), TableEvent::Append { .. });
         }
     }
+
+    #[tokio::test]
+    async fn test_send_table_event_ok() {
+        let (tx, mut rx) = mpsc::channel(1);
+        let res = Sink::send_table_event(&tx, TableEvent::DropTable).await;
+        assert!(res.is_ok());
```
dentiny, line 730: You should unwrap it directly, otherwise error status and message get lost

Issue #1687, opened 2025-08-24
BUG: iceberg sync not allow same filename


dentiny:
## Summary

```sh
thread 'tokio-runtime-worker' panicked at src/moonlink/src/storage/iceberg/iceberg_table_syncer.rs:221:13:
assertion failed: self.remote_data_file_to_file_id.insert(iceberg_data_file.file_path().to_string(),
        local_data_file.file_id()).is_none()
stack backtrace:
   0: __rustc::rust_begin_unwind
   1: core::panicking::panic_fmt
   2: core::panicking::panic
   3: moonlink::storage::iceberg::iceberg_table_syncer::<impl moonlink::storage::iceberg::iceberg_table_manager::IcebergTableManager>::sync_data_files::{{closure}}
   4: moonlink::storage::iceberg::iceberg_table_syncer::<impl moonlink::storage::iceberg::iceberg_table_manager::IcebergTableManager>::sync_snapshot_impl::{{closure}}
   5: <moonlink::storage::iceberg::iceberg_table_manager::IcebergTableManager as moonlink::storage::iceberg::table_manager::TableManager>::sync_snapshot::{{closure}}
```

PR #1688, opened 2025-08-24
[storage] Iceberg sync allows duplicate filename

dentiny:
## Summary

Current remote file (aka, iceberg filepath) is named as `warehouse/namespace/table/filename`; before file upload API it's not a problem, since disk slice writer always suffixes with a random string (internal implementation gurantee); but that's not the case for uploaded files.

There're a few solutions:
- (this PR) suffixes with a uuid v7, which keeps the filename to easier inspection, also maintain ingestion order
  + It's also the consistent handling for moonlink index files
- do some truncation and include certain (sub-)directory names; the concern is to which level

I envision when we implement partition in moonlink (should be soon), the filepath naming will be updated accordingly.

## Related Issues

Closes https://github.com/Mooncake-Labs/moonlink/issues/1687

## Checklist

- [x] Code builds correctly

### DIFF
--- a/src/moonlink/src/storage/iceberg/io_utils.rs
+++ b/src/moonlink/src/storage/iceberg/io_utils.rs
@@ -28,7 +28,8 @@ pub(crate) async fn write_record_batch_to_iceberg(
         .unwrap()
         .to_string();
     let location_generator = DefaultLocationGenerator::new(table.metadata().clone())?;
-    let remote_filepath = location_generator.generate_location(&filename);
+    let remote_filepath =
+        location_generator.generate_location(&format!("{}-{}", filename, uuid::Uuid::now_v7()));
     // Import local parquet file to remote.
     filesystem_accessor
         .copy_from_local_to_remote(local_filepath, &remote_filepath)
--- a/src/moonlink/src/storage/iceberg/tests.rs
+++ b/src/moonlink/src/storage/iceberg/tests.rs
@@ -464,14 +464,14 @@ async fn test_store_and_load_snapshot_impl(iceberg_table_config: IcebergTableCon
         ));
 
         // Check second data file and its deletion vector.
-        if file_path.ends_with(data_filename_2) {
+        if file_path.contains(data_filename_2) {
             assert_eq!(loaded_arrow_batch, batch_2,);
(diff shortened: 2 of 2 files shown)


dpxcc:
> suffixes with a uuid v7, which keeps the filename to easier inspection, also maintain ingestion order

We can’t preserve the original filename once compaction is enabled
So just name it the same way as a non-event table

dentiny:
> > suffixes with a uuid v7, which keeps the filename to easier inspection, also maintain ingestion order
> 
> We can’t preserve the original filename once compaction is enabled So just name it the same way as a non-event table

Not sure what do you mean, iceberg layer shares the same implementation for both event vs non-event table.
Keeping the old filename as part of the iceberg file name is for better debuggability.

dpxcc:
Lol, data files in non-event tables will now be named `data-<uuid>.parquet-<uuid>`…
This is bad! We should name them consistently as `data-<uuid>.parquet` for both event and non-event tables. The original filename can’t be preserved anyway once compaction is enabled

dentiny:
So when I was working on the PR, I followed two guidelines:
- prefix / suffix should be added at sink,  instead of source
- Uploaded filenames are more important to keep than moonlink internal filenames, since it's possible to contain human-readable information

The reason being:
- Moonlink internal filenames are already random string, which is not made for human to read; so having multiple UUID in the filename doesn't cause any issue
- If you really want, we could rename disk slice writer, just like what we do for temporary parquet files

I would like to know why you think it's bad, what's the justification, or it's your personal preference.

src/moonlink/src/storage/iceberg/io_utils.rs
```diff
@@ -28,7 +28,8 @@ pub(crate) async fn write_record_batch_to_iceberg(
         .unwrap()
         .to_string();
     let location_generator = DefaultLocationGenerator::new(table.metadata().clone())?;
-    let remote_filepath = location_generator.generate_location(&filename);
+    let remote_filepath =
```
dentiny, line 31: The real issue is here: `generate_location` doesn't consider suffix (another function did it), will fix in the followup :hear_no_evil:

PR #1711, opened 2025-08-25
[REST] improve error propagation for REST API 

### LINKED ISSUE #1652: REST: error propagation for rest API
## Summary

https://github.com/Mooncake-Labs/moonlink/blob/3130da9912a74ef4398fcc76f3eaf60af41133f0/src/moonlink_service/src/rest_api.rs#L261

for now we're always using INTERNAL error, which is supposed to return only severe invariant breaks.

Andrewtangtang:
## Summary

Improve REST API error propagation by replacing hardcoded `INTERNAL_SERVER_ERROR` with appropriate HTTP status codes based on error type and context. This ensures that 500 Internal Server Error is reserved for severe invariant breaks as intended, while client errors return more appropriate 4xx status codes.

## Related Issues

Closes #1652

## Changes

- Added `get_status()` to `moonlink_backend::Error` for retrieving `ErrorStatus`
- Implemented:

  - `get_backend_error_status_code()` — maps backend errors to correct HTTP codes
  - `get_context_error_status_code()` — maps context-specific errors to correct HTTP codes
- Updated all REST API endpoints to replace hardcoded `INTERNAL_SERVER_ERROR` with proper mappings:

  - `invalid_schema` / `invalid_operation` → **400 Bad Request**
  - Temporary backend errors → **503 Service Unavailable**
  - Permanent system errors → **500 Internal Server Error**

## Checklist

- [x] Code builds correctly

### DIFF
--- a/src/moonlink_service/src/rest_api.rs
+++ b/src/moonlink_service/src/rest_api.rs
@@ -9,6 +9,7 @@ use moonlink::StorageConfig;
 use moonlink_backend::{table_config::TableConfig, table_status::TableStatus};
 use moonlink_backend::{EventRequest, FileEventOperation, RowEventOperation};
 use moonlink_backend::{FileEventRequest, RowEventRequest, REST_API_URI};
+use moonlink_error::ErrorStatus;
 use serde::{Deserialize, Serialize};
 use std::collections::HashMap;
 use std::sync::Arc;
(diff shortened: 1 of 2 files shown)


src/moonlink_service/src/rest_api.rs
```diff
@@ -161,7 +162,31 @@ pub struct HealthResponse {
...
+
+        _ => match error.get_status() {
+            ErrorStatus::Temporary => StatusCode::SERVICE_UNAVAILABLE,
+            ErrorStatus::Permanent => StatusCode::INTERNAL_SERVER_ERROR,
+        },
+    }
+}
+
+/// Map specific error contexts that don't have backend errors
+fn get_context_error_status_code(error_type: &str) -> StatusCode {
```
dentiny: I'm not sure how useful the util function is, what about we directly assigning status code?

dentiny (commented):
Thanks for the change!

The ideal state I'm praying for is we have a centralized error code system and a number of adaptors (i.e., with http status code), but don't expect it any time soon.

I think the error translation part looks ok to me, at least it's better than returning INTERNAL_ERROR whatever error it is.

But could you please revert all changes for data ingestion and file upload?
The reason is:
- http error status is served to indicate retryable or not
- so it's ok for read and table creation and drop, since they're intrinsically idempotent; but we haven't implemented such idempotency for ingestion.

src/moonlink_service/src/rest_api.rs
Andrewtangtang: Sure, no problem — I also think directly assigning the status code here would be more straightforward. I can update it accordingly.

Andrewtangtang:
@dentiny I’ve reverted the changes for data ingestion and file upload as requested. Please take another look when you have time. Thanks.

src/moonlink_service/src/rest_api.rs
```diff
@@ -161,7 +162,21 @@ pub struct HealthResponse {
     pub timestamp: u64,
 }
 
-/// Create the router with all API endpoints
+/// Map backend error to appropriate HTTP status code based on error type
```
dentiny, line 165: It's fine for now to translate via backend error, but I'm not 100% sure that's the case in the future. Subject to change in the future.

dentiny (approved):
Thanks!

PR #1720, opened 2025-08-26
REST: expose endpoints to optimize table

### LINKED ISSUE #1651: REST: expose endpoints to optimize table
## Summary

```sh
POST /tables/{table}/optimize
optimize_table(&self, database: String, table: String, mode: &str)
```
reference: https://github.com/Mooncake-Labs/moonlink/pull/1649
backend rpc: https://github.com/Mooncake-Labs/moonlink/blob/3130da9912a74ef4398fcc76f3eaf60af41133f0/src/moonlink_backend/src/lib.rs#L255-L282

Standing-Man:
## Summary

add endpoints to optimize table

## Related Issues

- Closes #1651.

## Changes

- 
- 
- 

## Checklist

- [x] Code builds correctly

### DIFF
--- a/src/moonlink_service/src/rest_api.rs
+++ b/src/moonlink_service/src/rest_api.rs
@@ -470,14 +470,12 @@ async fn optimize_table(
         Ok(_) => Ok(Json(OptimizeTableResponse {})),
         Err(e) => {
             error!("Failed to optimize table '{}': {}", table, e);
+            let status_code = get_backend_error_status_code(&e);
             Err((
-                StatusCode::BAD_REQUEST,
+                status_code,
                 Json(ErrorResponse {
-                    error: "invalid_mode".to_string(),
-                    message: format!(
-                        "Invalid mode '{}'. Must be 'data', 'index' or 'full'",
-                        payload.mode
-                    ),
+                    error: "table_optimize_failed".to_string(),
+                    message: format!("Failed to optimize table: {e}"),
                 }),
             ))
         }
--- a/src/moonlink_service/src/rest_api.rs
+++ b/src/moonlink_service/src/rest_api.rs
@@ -107,6 +107,22 @@ pub struct ListTablesResponse {
 }
 
 /// ====================
+/// Optimize table
+/// ====================
+///
+/// Request structure for table optimize.
+#[derive(Debug, Serialize, Deserialize)]
+pub struct OptimizeTableRequest {
+    pub database: String,
+    pub table: String,
+    pub mode: String,
+}
+
+/// Response structure for table optimize.
+#[derive(Debug, Serialize, Deserialize)]
+pub struct OptimizeTableResponse {}
+
+/// ====================
 /// Data ingestion
 /// ====================
 ///
@@ -185,6 +201,7 @@ pub fn create_router(state: ApiState) -> Router {
         .route("/tables/{table}", delete(drop_table))
         .route("/ingest/{table}", post(ingest_data))
         .route("/upload/{table}", post(upload_files))
+        .route("/tables/{table}/optimize", post(optimize_table))
         .with_state(state)
         .layer(
             CorsLayer::new()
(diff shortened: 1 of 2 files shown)


src/moonlink_service/src/rest_api.rs
```diff
@@ -421,6 +438,37 @@ async fn upload_files(
...
+    match state
+        .backend
+        .optimize_table(payload.database, payload.table, payload.mode.as_str())
+        .await
+    {
+        Ok(_) => Ok(Json(OptimizeTableResponse {})),
+        Err(e) => {
+            error!("Failed to optimize table '{}': {}", table, e);
+            Err((
+                StatusCode::BAD_REQUEST,
```
dentiny: How do we differentiate user error and maintenance operation failure?

src/moonlink_service/src/test.rs
```diff
@@ -214,6 +224,46 @@ async fn test_schema() {
...
+        start_with_config(config).await.unwrap();
+    });
+    test_readiness_probe().await;
+
+    // Create test table.
+    let client = reqwest::Client::new();
+    create_table(&client, DATABASE, TABLE, /*append_only=*/ false).await;
+
+    // test for optimize table
+    optimize_table(&client, DATABASE, TABLE, "full").await;
```
dentiny: We need to at least populate some data into the table, otherwise maintenance operations are no-op

src/moonlink_service/src/rest_api.rs
Standing-Man:
```
  Error::ParseIntError(es)
  Error::PostgresSource(es) 
  Error::Io(es)
  Error::MoonlinkConnectorError(es) 
  Error::MoonlinkError(es)
  Error::MoonlinkMetadataStoreError(es)
  Error::InvalidArgumentError(es) 
  Error::TokioWatchRecvError(es) 
  Error::Json(es) 
  Error::MpscChannelSendError(es)
```
Do all errors need to be individually wrapped into `ErrorResponse`?

src/moonlink_service/src/test.rs
Standing-Man: I'm a bit confused — after inserting some data into the table, what's the recommended way to verify that `optimize_table` has run successfully?

src/moonlink_service/src/rest_api.rs
dentiny: My ideal way of error propagation is to have a centralized error code ([example](https://grpc.io/docs/guides/status-codes/)), and have multiple adaptors with external systems, but I don't envision that in the next future months.

Do you think we could reuse https://github.com/Mooncake-Labs/moonlink/blob/096ab2eaddc1b3a8e2a4bcf4799b67216e35ca8f/src/moonlink_service/src/rest_api.rs#L165-L177?

src/moonlink_service/src/test.rs
dentiny: Thanks for asking!

I'm thinking about some easy ways for testing, two ways I could think of:
- there's no file flush / snapshot control at rest API, I'm thinking whether we could randomly ingest some data, and call these maintenance functions and make sure content is correct after maintenance
- or we could use file upload, to see whether data files are correctly compacted (i.e., two files into one)

Standing-Man: I have properly handled the error returns and tested `optimize_table` function using the first approach. Many thanks for your suggestion.

src/moonlink_service/src/rest_api.rs
```diff
@@ -436,6 +453,35 @@ async fn upload_files(
...
+        table, payload
+    );
+    match state
+        .backend
+        .optimize_table(payload.database, payload.table, payload.mode.as_str())
+        .await
+    {
+        Ok(_) => Ok(Json(OptimizeTableResponse {})),
+        Err(e) => {
+            error!("Failed to optimize table '{}': {}", table, e);
```
dentiny, line 472: A note to myself, I need to revisit whether error logging here is good and needed.

Issue #1617, opened 2025-08-20
iceberg: implement table functions


dentiny:
## Summary

we could use iceberg-rust and make a simple wrapper (apart from `update_table`)
- create: https://github.com/apache/iceberg-rust/blob/fa5f2ed8dcd60be7438c5a3bfe098490ecd03e3a/crates/iceberg/src/catalog/mod.rs#L88
- drop: https://github.com/apache/iceberg-rust/blob/fa5f2ed8dcd60be7438c5a3bfe098490ecd03e3a/crates/iceberg/src/catalog/mod.rs#L98
- existence: https://github.com/apache/iceberg-rust/blob/fa5f2ed8dcd60be7438c5a3bfe098490ecd03e3a/crates/iceberg/src/catalog/mod.rs#L101
- load: https://github.com/apache/iceberg-rust/blob/fa5f2ed8dcd60be7438c5a3bfe098490ecd03e3a/crates/iceberg/src/catalog/mod.rs#L95

fweilun:
I'd like to deal with it, thanks.

dentiny:
> I'd like to deal with it, thanks.

Thanks for the help!

PR #1725, opened 2025-08-26
Implement rest catalog table functions

fweilun:
## Summary

- Wrap the `iceberg_catalog_rest::RestCatalog` interface
- Implement the create, drop, exists, and load methods

## Related Issues

Closes https://github.com/Mooncake-Labs/moonlink/issues/1617

## Checklist

- [✓] Code builds correctly


Cargo.toml
```diff
@@ -40,6 +40,7 @@ hashbrown = "0.15"
 iceberg = { git = "https://github.com/apache/iceberg-rust.git", rev = "7cde26a7d793905bbbe4d7264b215ca9dc7e9567", default-features = false, features = [
   "storage-fs",
 ] }
+iceberg-catalog-rest = { git = "https://github.com/apache/iceberg-rust.git", rev = "7cde26a7d793905bbbe4d7264b215ca9dc7e9567" }
```
dentiny: Wait, why do we have two versions of iceberg-rust?

dentiny:
CC @peterxcli for review as well

Cargo.toml
dentiny: my bad, iceberg-rest only contains `iceberg` crate, and rest seems to live another crate.

dentiny:
Do you mind rebasing the PR? Thanks!

src/moonlink/src/storage/iceberg/rest_catalog.rs
```diff
@@ -0,0 +1,292 @@
...
+        let builder = IcebergRestCatalogBuilder::default().with_client(Client::new());
+        let catalog = builder
+            .load(
+                "test",
+                HashMap::from([
+                    (
+                        REST_CATALOG_PROP_URI.to_string(),
+                        "http://localhost:8181".to_string(),
+                    ),
+                    ("a".to_string(), "b".to_string()),
```
peterxcli: Curious what's this a, b for?

fweilun: I added it for some tests, but it’s not functioning. Thanks for reminding.

Cargo.toml
```diff
@@ -41,8 +41,10 @@ hashbrown = "0.15"
 iceberg = { git = "https://github.com/apache/iceberg-rust.git", rev = "4aacad756250a5a1bbc44842ea4b608e3213b641", default-features = false, features = [
   "storage-fs",
 ] }
+iceberg-catalog-rest = { git = "https://github.com/apache/iceberg-rust.git", rev = "4aacad756250a5a1bbc44842ea4b608e3213b641" }
 itertools = { version = "0.14" }
 lru = { version = "0.14" }
+mockito = { version = "1.7.0" }
```
dentiny: 1. please place test only dependency under dev dependency
2. we already have `mockall` in the code base

src/moonlink/src/storage/iceberg/rest_catalog.rs
```diff
@@ -0,0 +1,272 @@
...
+    async fn register_table(
+        &self,
+        __table: &TableIdent,
+        _metadata_location: String,
+    ) -> IcebergResult<Table> {
+        todo!("register existing table is not supported")
+    }
+}
+
+#[cfg(feature = "storage-gcs")]
```
dentiny: Emm how is the test case has anything to do with GCS?

src/moonlink/src/storage/iceberg/rest_catalog.rs
```diff
@@ -0,0 +1,272 @@
...
+pub struct RestCatalog {
+    catalog: IcebergRestCatalog,
+}
+
+impl RestCatalog {
+    #[allow(dead_code)]
+    pub async fn new(
+        builder: IcebergRestCatalogBuilder,
+        name: impl Into<String>,
+        props: HashMap<String, String>,
```
dentiny: I would prefer a schema-tized struct instead of a map here.
To make it clear, you need to add proper documentation on what are the required key-value pairs, which is essentially the same as a struct.

src/moonlink/src/storage/iceberg/rest_catalog.rs
```diff
@@ -0,0 +1,272 @@
...
+#[derive(Debug)]
+pub struct RestCatalog {
+    catalog: IcebergRestCatalog,
+}
+
+impl RestCatalog {
+    #[allow(dead_code)]
+    pub async fn new(
+        builder: IcebergRestCatalogBuilder,
+        name: impl Into<String>,
```
dentiny: Please add a documentation on `name`

src/moonlink/src/storage/iceberg/rest_catalog.rs
```diff
@@ -0,0 +1,272 @@
...
+
+#[derive(Debug)]
+pub struct RestCatalog {
+    catalog: IcebergRestCatalog,
+}
+
+impl RestCatalog {
+    #[allow(dead_code)]
+    pub async fn new(
+        builder: IcebergRestCatalogBuilder,
```
dentiny: curious why do we pass catalog builder from caller, instead of constructing it internally?
Are you planning to do dependency injection for testing?

src/moonlink/src/storage/iceberg/rest_catalog.rs
```diff
@@ -0,0 +1,272 @@
...
+        name: impl Into<String>,
+        props: HashMap<String, String>,
+    ) -> Self {
+        Self {
+            catalog: builder.load(name, props).await.unwrap(),
```
dentiny: Let's do error propagation properly.

src/moonlink/src/storage/iceberg/rest_catalog.rs
```diff
@@ -0,0 +1,272 @@
...
+        let builder = IcebergRestCatalogBuilder::default().with_client(Client::new());
+        let catalog = builder
+            .load(
+                "test",
+                HashMap::from([(
+                    REST_CATALOG_PROP_URI.to_string(),
+                    "http://localhost:8181".to_string(),
+                )]),
+            )
+            .await;
```
dentiny: for unit tests, I usually `unwrap` directly. but not a requirement

src/moonlink/src/storage/iceberg/rest_catalog.rs
```diff
@@ -0,0 +1,272 @@
...
+        assert!(catalog.is_ok());
+        let table_creation = TableCreation::builder()
+            .name("test1".to_string())
```
dentiny: I'm afraid it's very easy to get flaky test. rust runs all unit tests parallelly, unless you explicitly declare them to be serial. I would suggest using a random table name here.

src/moonlink/src/storage/iceberg/rest_catalog.rs
```diff
@@ -0,0 +1,272 @@
...
+                    .unwrap(),
+            )
+            .build();
+
+        let namespace = NamespaceIdent::new("ns1".to_string());
+        let result = catalog
+            .unwrap()
+            .create_table(&namespace, table_creation)
+            .await;
+        match result {
```
dentiny: I'm not sure if we ignore creation errors what's the point of the unit test..
I would suggest:
- Use a random string as table name
- Write a RAII-style class for table drop

dentiny: Example: 
- https://github.com/Mooncake-Labs/moonlink/blob/main/src/moonlink/src/storage/filesystem/gcs/test_guard.rs
- https://github.com/Mooncake-Labs/moonlink/blob/main/src/moonlink/src/storage/filesystem/gcs/gcs_test_utils.rs

src/moonlink/src/storage/iceberg.rs
```diff
@@ -20,6 +20,7 @@ pub(super) mod parquet_stats_utils;
 pub(super) mod parquet_utils;
 pub(super) mod puffin_utils;
 pub(super) mod puffin_writer_proxy;
+pub(super) mod rest_catalog;
```
dentiny, line 26: let's guard with features, so it's not enabled by default.
The reason being, for iceberg-rust, all catalogs are all independent crates;
I would like to follow the similar style to avoid unnecessary dependencies.

fweilun: I understand the reason, but why is this related to independent crates.

dentiny: That's the impl for iceberg-rust: https://github.com/apache/iceberg-rust/tree/main/crates/catalog

Standing-Man:
Hi @fweilun, thanks for your contributions! To keep the git history clean, you may want to consider using `rebase` instead of `merge` when syncing with the main branch, and then push safely with `--force-with-lease`.

dentiny:
> error[E0433]: failed to resolve: use of unresolved module or unlinked crate `reqwest`
  --> src/moonlink/src/storage/filesystem/s3/s3_test_utils.rs:59:18
   |
59 |     let client = reqwest::Client::new();
   |                  ^^^^^^^ use of unresolved module or unlinked crate `reqwest`
   |
   = help: if you wanted to use a crate named `reqwest`, use `cargo add reqwest` to add it to your `Cargo.toml`

error[E0433]: failed to resolve: use of unresolved module or unlinked crate `reqwest`
   --> src/moonlink/src/storage/filesystem/s3/s3_test_utils.rs:103:18
    |
103 |     let client = reqwest::Client::new();
    |                  ^^^^^^^ use of unresolved module or unlinked crate `reqwest`
    |
    = help: if you wanted to use a crate named `reqwest`, use `cargo add reqwest` to add it to your `Cargo.toml`

Let's fix the object storage unit tests

src/moonlink/src/storage/iceberg/rest_catalog.rs
fweilun: Working on it.

fweilun:
> Hi @fweilun, thanks for your contributions! To keep the git history clean, you may want to consider using `rebase` instead of `merge` when syncing with the main branch, and then push safely with `--force-with-lease`.

Thanks for the advice ~

.github/workflows/ci.yml
```diff
@@ -108,6 +108,47 @@ jobs:
             echo "Waiting for Fake GCS..."
             sleep 2
           done
+      
+      # ─────────────────────── Start iceberg-rest ──────────────────────────
```
dentiny: Why don't we create a separate test suite for iceberg-rest?

src/moonlink/src/storage/iceberg.rs
```diff
@@ -58,6 +71,9 @@ mod file_catalog_test_utils;
 #[cfg(test)]
 mod file_catalog_test;
 
+#[cfg(test)]
+mod rest_catalog_test;
```
dentiny, line 76: Should it be guarded by `rest-catalog` feature?

src/moonlink/src/storage/iceberg/rest_catalog.rs
```diff
@@ -0,0 +1,291 @@
...
+        config.props.insert(
+            REST_CATALOG_PROP_URI.to_string(),
+            config.rest_catalog_prop_uri,
+        );
+        if let Some(warehouse) = config.rest_catalog_prop_warehouse {
+            config
+                .props
+                .insert(REST_CATALOG_PROP_WAREHOUSE.to_string(), warehouse);
+        }
+        let catalog = builder.load(config.name, config.props).await.unwrap();
```
dentiny: let's do error propagation properly

src/moonlink/src/storage/iceberg/rest_catalog.rs
```diff
@@ -0,0 +1,291 @@
...
+    pub(crate) catalog: IcebergRestCatalog,
+}
+
+#[derive(Debug, Default)]
+pub struct RestCatalogConfig {
+    // catalog name
+    pub name: String,
+    pub rest_catalog_prop_uri: String,
+    pub rest_catalog_prop_warehouse: Option<String>,
+    pub props: HashMap<String, String>,
```
dentiny: I'm still confused why we need properties field here; to me, what we need is warehouse and URI.

Think from catalog library users, what should we put in the properties map?
How does rest catalog handle these key-value pairs?
For unknown properties, are they ignored or errors?
All these needs proper documentation.

My preference is to get necessary information into catalog field, and only add arbitrary key-value pairs when needed.

src/moonlink/src/storage/iceberg/rest_catalog.rs
```diff
@@ -0,0 +1,291 @@
...
+    pub client: Option<Client>,
```
dentiny: Need documentation, what if user doesn't provide a HTTP client?

src/moonlink/src/storage/iceberg/rest_catalog.rs
```diff
@@ -0,0 +1,291 @@
...
+#[derive(Debug)]
+pub struct RestCatalog {
+    pub(crate) catalog: IcebergRestCatalog,
+}
+
+#[derive(Debug, Default)]
+pub struct RestCatalogConfig {
+    // catalog name
+    pub name: String,
+    pub rest_catalog_prop_uri: String,
```
dentiny: Curious, what if user doesn't provide a warehouse string? Documentation needed here.

src/moonlink/src/storage/iceberg/rest_catalog.rs
```diff
@@ -0,0 +1,291 @@
...
+use std::collections::HashMap;
+
+#[derive(Debug)]
+pub struct RestCatalog {
+    pub(crate) catalog: IcebergRestCatalog,
+}
+
+#[derive(Debug, Default)]
+pub struct RestCatalogConfig {
+    // catalog name
```
dentiny: Is the name used for informational purpose or debugging purpose?

src/moonlink/src/storage/iceberg/rest_catalog.rs
```diff
@@ -0,0 +1,291 @@
...
+    async fn register_table(
+        &self,
+        __table: &TableIdent,
+        _metadata_location: String,
+    ) -> IcebergResult<Table> {
+        todo!("register existing table is not supported")
+    }
+}
+
+#[cfg(feature = "rest-catalog")]
```
dentiny: curious we've already declare the whole `rest_catalog` module under the feature `rest-catalog`, is it needed here?

src/moonlink/src/storage/iceberg/rest_catalog.rs
```diff
@@ -0,0 +1,291 @@
...
+                            .into(),
+                    ])
+                    .build()
+                    .unwrap(),
+            )
+            .build();
+
+        let namespace = NamespaceIdent::new("ns1".to_string());
+        let result = catalog
+            .unwrap()
```
dentiny: catalog is unwrapped here anyway, curious why don't we directly unwrap L169?

src/moonlink/src/storage/iceberg/rest_catalog.rs
```diff
@@ -0,0 +1,291 @@
...
+            Ok(_) => {}
+            Err(e) if e.to_string().contains("does not exist") => {}
+            Err(e) => {
+                panic!("load table error: {e:?}");
+            }
+        }
+    }
+
+    #[tokio::test]
+    async fn test_table_exists() {
```
dentiny: Curious why do we split unit tests in two files?

src/moonlink/src/storage/iceberg/rest_catalog_test_utils.rs
```diff
@@ -0,0 +1,39 @@
...
+    rng.sample_iter(&Alphanumeric)
+        .take(10)
+        .map(char::from)
+        .collect()
+}
+
+pub(crate) fn default_rest_catalog_config() -> RestCatalogConfig {
+    RestCatalogConfig {
+        name: "test".to_string(),
+        rest_catalog_prop_uri: "http://localhost:8181".to_string(),
```
dentiny: nit: make it a constant

src/moonlink/src/storage/iceberg/rest_catalog_test_utils.rs
```diff
@@ -0,0 +1,39 @@
+use crate::storage::iceberg::rest_catalog::RestCatalogConfig;
+use iceberg::spec::{NestedField, PrimitiveType, Schema, Type};
+use iceberg::TableCreation;
+use rand::{distr::Alphanumeric, Rng};
+use std::collections::HashMap;
+
+pub(crate) fn random_string() -> String {
```
dentiny:
```suggestion
pub(crate) fn get_random_string() -> String {
```
verb + noun

src/moonlink/src/storage/iceberg/rest_catalog_test_guard.rs
```diff
@@ -0,0 +1,53 @@
...
+
+impl Drop for RestCatalogTestGuard {
+    fn drop(&mut self) {
+        let table = self.table.take();
+        let config = default_rest_catalog_config();
+        tokio::task::block_in_place(|| {
+            tokio::runtime::Handle::current().block_on(async move {
+                let catalog = RestCatalog::new(config).await;
+                if let Some(t) = table {
+                    if let Err(e) = catalog.drop_table(&t).await {
```
dentiny: What about directly unwrap on errors here? The reason is:
- We don't need error propagation for unit tests
- It could lead to flaky test for the following unit tests

src/moonlink/src/storage/iceberg/rest_catalog_test.rs
```diff
@@ -0,0 +1,83 @@
...
+    use crate::storage::iceberg::rest_catalog_test_utils::*;
+    use iceberg::{Catalog, NamespaceIdent};
+
+    #[tokio::test(flavor = "multi_thread", worker_threads = 2)]
+    async fn test_create_table() {
+        let namespace = random_string();
+        let table = random_string();
+        let _guard = RestCatalogTestGuard::new(namespace.clone(), None)
+            .await
+            .expect("Namespace creation fail");
```
dentiny:
```suggestion
            .expect("Rest catalog test guard creation fail");
```
namespace or table creation is an internal detail

src/moonlink/src/storage/iceberg/rest_catalog_test.rs
```diff
@@ -0,0 +1,83 @@
...
+        let config = default_rest_catalog_config();
+        let catalog = RestCatalog::new(config).await;
+        let namespace = NamespaceIdent::new(namespace);
+        let table_creation = default_table_creation(table);
+        catalog
+            .create_table(&namespace, table_creation)
+            .await
+            .expect("Table creation fail");
```
dentiny: nit: I won't be too picky here, but if you want to do proper error handling with `expect`, might be better to include namespace and table name here; because we could still debug after test failure with local rest catalog container.

Same for all below.

src/moonlink/Cargo.toml
```diff
@@ -59,6 +60,7 @@ futures = { workspace = true }
 hashbrown = { workspace = true }
 hmac = { version = "0.12", optional = true }
 iceberg = { workspace = true }
+iceberg-catalog-rest = { workspace = true }
```
dentiny:
```suggestion
iceberg-catalog-rest = { workspace = true, optional = true }
```

src/moonlink/src/storage/iceberg/rest_catalog.rs
fweilun: The error would never occurs, since both name and uri are given. Should I still propagate it?

dentiny: The general practice is to do error propagation, since the interface clearly shows it's possible to encounter error; no one guarantees the implementation won't be changed in the future.

If you really want to unwrap here, you need to do two things:
- Leave a comment why it's safe and ok to unwrap
- Prefer to use expect and provide some context here, rather than directly unwrap

src/moonlink/src/storage/iceberg/rest_catalog.rs
fweilun: The prop variable provides information for `Iceberg RestCatalogConfig`, such as "prefix", "oauth2-server-uri", "token"... [ref](https://github.com/apache/iceberg-rust/blob/3e6779ee09c7e54d7ae8101f8477f5d04486ecc3/crates/catalog/rest/src/catalog.rs#L147C1-L160C21)

They all handle these values by self.props.get("specific_key")
I think they ignore them.

I think adding some description for the specific keys & the questions you've asked  is a good idea.

fweilun: > The prop variable provides information for `Iceberg RestCatalogConfig`, such as "prefix", "oauth2-server-uri", "token"... [ref](https://github.com/apache/iceberg-rust/blob/3e6779ee09c7e54d7ae8101f8477f5d04486ecc3/crates/catalog/rest/src/catalog.rs#L147C1-L160C21)
> 
> They all handle these values by self.props.get("specific_key") I think they ignore them.
> 
> I think adding some description for the specific keys & the questions you mentioned is a good idea.

dentiny: Sounds good, thank you!

src/moonlink/src/storage/iceberg/rest_catalog_test_utils.rs
fweilun: Sure, I’ll also rename them to uri and warehouse for cleaner code.

src/moonlink/src/storage/iceberg/rest_catalog.rs
```diff
@@ -0,0 +1,291 @@
...
+use reqwest::Client;
+use std::collections::HashMap;
+
+#[derive(Debug)]
+pub struct RestCatalog {
+    pub(crate) catalog: IcebergRestCatalog,
+}
+
+#[derive(Debug, Default)]
+pub struct RestCatalogConfig {
```
dentiny: Wait I think we already have one
https://github.com/Mooncake-Labs/moonlink/blob/94656f903a8fb91565b6fe41b7ebcdef6d9729e8/src/moonlink/src/storage/iceberg/iceberg_table_config.rs#L68

src/moonlink/src/storage/iceberg/rest_catalog.rs
fweilun: I think It's better not to use warehouse as a option<>.

src/moonlink/src/storage/iceberg/iceberg_table_config.rs
```diff
@@ -4,17 +4,30 @@ use std::collections::HashMap;
...
+
+    /// Optional configuration properties:
+    ///
+    /// - prefix:          Optional URL path prefix to insert after the base URI and API version.
+    /// - oauth2-server-uri: Custom OAuth2 server URI. Defaults to: [uri, PATH_V1:"v1", "oauth", "tokens"].join("/")
+    /// - token:           Static authentication token used by the client for sending requests.
+    /// - credentials:     Client credentials used to fetch a new token.
+    ///     - None: No credentials provided.
+    ///     - Some(None, client_secret): Only client_secret is provided.
+    ///     - Some(Some(client_id), client_secret): Both client_id and client_secret are provided.
```
dentiny, line 27: My personal preference is, if there're not too many fields, I would put all of them as field member.
But I don't have a strong opinion here, as long as we have documentation.

nit: worth calling out other unknown fields will be ignored.

src/moonlink/src/storage/iceberg/rest_catalog_test.rs
```diff
@@ -0,0 +1,106 @@
+#[cfg(test)]
```
dentiny, line 1: actually we don't need the annotation again

src/moonlink/src/storage/iceberg/rest_catalog_test.rs
```diff
@@ -0,0 +1,106 @@
...
+    use crate::storage::iceberg::rest_catalog::RestCatalog;
+    use crate::storage::iceberg::rest_catalog_test_guard::RestCatalogTestGuard;
+    use crate::storage::iceberg::rest_catalog_test_utils::*;
+    use iceberg::{Catalog, NamespaceIdent, TableIdent};
+
+    #[tokio::test(flavor = "multi_thread", worker_threads = 2)]
+    async fn test_create_table() {
+        let namespace = get_random_string();
+        let table = get_random_string();
+        let mut guard = RestCatalogTestGuard::new(namespace.clone(), None)
```
dentiny, line 12:
```suggestion
        let mut guard = RestCatalogTestGuard::new(namespace.clone(), /*table=*/None)
```

src/moonlink/Cargo.toml
```diff
@@ -8,6 +8,7 @@ license = { workspace = true }
 default = ["storage-fs"]
 storage-all = ["storage-fs", "storage-s3", "storage-gcs"]
 test-utils = []
+rest-catalog = ["iceberg-catalog-rest"]
```
dentiny, line 11: We should have one unified rest catalog feature: https://github.com/Mooncake-Labs/moonlink/blob/94656f903a8fb91565b6fe41b7ebcdef6d9729e8/src/moonlink/Cargo.toml#L34

src/moonlink/src/storage/iceberg/rest_catalog_test_guard.rs
```diff
@@ -0,0 +1,53 @@
...
+
+impl Drop for RestCatalogTestGuard {
+    fn drop(&mut self) {
+        let table = self.table.take();
+        let config = default_rest_catalog_config();
+        tokio::task::block_in_place(|| {
+            tokio::runtime::Handle::current().block_on(async move {
+                let catalog = RestCatalog::new(config)
+                    .await
+                    .expect("Catalog creation fail");
```
dentiny, line 45: nit: I think we should better keep code consistent, to use expect or unwrap.

src/moonlink/src/storage/iceberg/rest_catalog_test.rs
```diff
@@ -0,0 +1,106 @@
...
+        let config = default_rest_catalog_config();
+        let catalog = RestCatalog::new(config)
+            .await
+            .expect("Catalog creation fail");
+        let table_ident = guard.table.clone().unwrap();
+        let result = catalog.load_table(&table_ident).await.unwrap_or_else(|_| {
+            panic!("Load table function fail, namespace={namespace} table={table}")
+        });
+        let result_table_ident = result.identifier().clone();
+        assert!(table_ident == result_table_ident);
```
dentiny, line 104: You should do `assert_eq!`

src/moonlink/src/storage/iceberg/rest_catalog.rs
```diff
@@ -0,0 +1,112 @@
...
+
+    async fn drop_namespace(&self, namespace_ident: &NamespaceIdent) -> IcebergResult<()> {
+        self.catalog.drop_namespace(namespace_ident).await
+    }
+
+    async fn list_tables(
+        &self,
+        _namespace_ident: &NamespaceIdent,
+    ) -> IcebergResult<Vec<TableIdent>> {
+        todo!("list tables is not supported");
```
dentiny, line 66: This API should be delegated as well

src/moonlink/src/storage/iceberg/rest_catalog_test.rs
```diff
@@ -0,0 +1,106 @@
...
+            .await
+            .expect("Catalog creation fail");
+        let namespace = NamespaceIdent::new(namespace);
+        let table_creation = default_table_creation(table.clone());
+        let table_name = table_creation.name.clone();
+        catalog
+            .create_table(&namespace, table_creation)
+            .await
+            .unwrap_or_else(|_| panic!("Table creation fail, namespace={namespace} table={table}"));
+        guard.table = Some(TableIdent::new(namespace, table_name));
```
dentiny, line 28: Are we testing anything here?

fweilun: I added this information so the guard can drop the table.

Standing-Man: It looks a bit unusual to insert the same namespace and catalog into both catalogs for the test. 🤔

Issue #1691, opened 2025-08-24
iceberg: provide config for more catalogs


dentiny:
## Summary

Current iceberg table config only supports file catalog: https://github.com/Mooncake-Labs/moonlink/blob/main/src/moonlink/src/storage/iceberg/iceberg_table_config.rs

PR #1729, opened 2025-08-26
Add Rest and Glue iceberg catalog configuration to support multiple catalog

peterxcli:
## Summary

Current iceberg table config only supports file catalog: [main/src/moonlink/src/storage/iceberg/iceberg_table_config.rs](https://github.com/Mooncake-Labs/moonlink/blob/main/src/moonlink/src/storage/iceberg/iceberg_table_config.rs?rgh-link-date=2025-08-24T08%3A46%3A19.000Z)

## Related Issues

Closes https://github.com/Mooncake-Labs/moonlink/issues/1691

## Changes

- Add RestCatalogConfig and GlueCatalogConfig for iceberg-rust CatalogConfig Builders
  - [GlueCatalogBuilder](https://github.com/apache/iceberg-rust/blob/4aacad756250a5a1bbc44842ea4b608e3213b641/crates/catalog/rest/src/catalog.rs#L73-L119)
  - [RestCatalogBuilder](https://github.com/apache/iceberg-rust/blob/4aacad756250a5a1bbc44842ea4b608e3213b641/crates/catalog/glue/src/catalog.rs#L66-L119)
- 

## Checklist

- [x] Code builds correctly

### DIFF
--- a/src/moonlink/Cargo.toml
+++ b/src/moonlink/Cargo.toml
@@ -31,6 +31,9 @@ storage-gcs = [
     "reqwest",
 ]
 
+catalog-rest = []
+catalog-glue = []
+
 bench = []
 
 chaos-test = ["function_name"]
--- a/src/moonlink/src/storage/iceberg/catalog_utils.rs
+++ b/src/moonlink/src/storage/iceberg/catalog_utils.rs
@@ -1,7 +1,7 @@
 #[cfg(test)]
 use crate::storage::filesystem::accessor::base_filesystem_accessor::BaseFileSystemAccess;
-use crate::storage::filesystem::accessor_config::AccessorConfig;
 use crate::storage::iceberg::file_catalog::FileCatalog;
+use crate::storage::iceberg::iceberg_table_config::IcebergCatalogConfig;
 use crate::storage::iceberg::moonlink_catalog::MoonlinkCatalog;
 
 use iceberg::spec::Schema as IcebergSchema;
@@ -12,17 +12,37 @@ use iceberg::Result as IcebergResult;
 /// It's worth noting catalog and warehouse uri are not 1-1 mapping; for example, rest catalog could handle warehouse.
 /// Here we simply deduce catalog type from warehouse because both filesystem and object storage catalog are only able to handle certain scheme.
 pub fn create_catalog(
-    accessor_config: AccessorConfig,
+    catalog: IcebergCatalogConfig,
     iceberg_schema: IcebergSchema,
 ) -> IcebergResult<Box<dyn MoonlinkCatalog>> {
-    Ok(Box::new(FileCatalog::new(accessor_config, iceberg_schema)?))
+    match catalog {
+        IcebergCatalogConfig::File { accessor_config } => {
+            Ok(Box::new(FileCatalog::new(accessor_config, iceberg_schema)?))
+        }
+        IcebergCatalogConfig::Rest { .. } | IcebergCatalogConfig::Glue { .. } => {
+            Err(iceberg::Error::new(
+                iceberg::ErrorKind::Unexpected,
+                "Only File catalog is supported currently",
+            ))
+        }
+    }
 }
 
 /// Create a catalog with no schema provided.
 pub fn create_catalog_without_schema(
-    accessor_config: AccessorConfig,
+    catalog: IcebergCatalogConfig,
 ) -> IcebergResult<Box<dyn MoonlinkCatalog>> {
-    Ok(Box::new(FileCatalog::new_without_schema(accessor_config)?))
+    match catalog {
+        IcebergCatalogConfig::File { accessor_config } => {
+            Ok(Box::new(FileCatalog::new_without_schema(accessor_config)?))
+        }
+        IcebergCatalogConfig::Rest { .. } | IcebergCatalogConfig::Glue { .. } => {
+            Err(iceberg::Error::new(
+                iceberg::ErrorKind::Unexpected,
+                "Only File catalog is supported currently",
+            ))
+        }
+    }
 }
 
 /// Test util function to create catalog with provided filesystem accessor.
--- a/src/moonlink/src/storage/iceberg/catalog_utils.rs
+++ b/src/moonlink/src/storage/iceberg/catalog_utils.rs
@@ -19,12 +19,16 @@ pub fn create_catalog(
         IcebergCatalogConfig::File { accessor_config } => {
             Ok(Box::new(FileCatalog::new(accessor_config, iceberg_schema)?))
         }
-        IcebergCatalogConfig::Rest { .. } | IcebergCatalogConfig::Glue { .. } => {
-            Err(iceberg::Error::new(
-                iceberg::ErrorKind::FeatureUnsupported,
-                "Only File catalog is supported currently",
-            ))
-        }
+        #[cfg(feature = "catalog-rest")]
+        IcebergCatalogConfig::Rest { .. } => Err(iceberg::Error::new(
+            iceberg::ErrorKind::FeatureUnsupported,
+            "Only File catalog is supported currently",
+        )),
+        #[cfg(feature = "catalog-glue")]
+        IcebergCatalogConfig::Glue { .. } => Err(iceberg::Error::new(
+            iceberg::ErrorKind::FeatureUnsupported,
+            "Only File catalog is supported currently",
+        )),
     }
 }
 
@@ -36,12 +40,16 @@ pub fn create_catalog_without_schema(
         IcebergCatalogConfig::File { accessor_config } => {
             Ok(Box::new(FileCatalog::new_without_schema(accessor_config)?))
         }
-        IcebergCatalogConfig::Rest { .. } | IcebergCatalogConfig::Glue { .. } => {
-            Err(iceberg::Error::new(
-                iceberg::ErrorKind::FeatureUnsupported,
-                "Only File catalog is supported currently",
-            ))
-        }
+        #[cfg(feature = "catalog-rest")]
+        IcebergCatalogConfig::Rest { .. } => Err(iceberg::Error::new(
+            iceberg::ErrorKind::FeatureUnsupported,
+            "Only File catalog is supported currently",
+        )),
+        #[cfg(feature = "catalog-glue")]
+        IcebergCatalogConfig::Glue { .. } => Err(iceberg::Error::new(
+            iceberg::ErrorKind::FeatureUnsupported,
+            "Only File catalog is supported currently",
+        )),
     }
 }
 
--- a/src/moonlink/src/storage/iceberg/catalog_utils.rs
+++ b/src/moonlink/src/storage/iceberg/catalog_utils.rs
@@ -21,7 +21,7 @@ pub fn create_catalog(
         }
         IcebergCatalogConfig::Rest { .. } | IcebergCatalogConfig::Glue { .. } => {
             Err(iceberg::Error::new(
-                iceberg::ErrorKind::Unexpected,
+                iceberg::ErrorKind::FeatureUnsupported,
                 "Only File catalog is supported currently",
             ))
         }
@@ -38,7 +38,7 @@ pub fn create_catalog_without_schema(
         }
         IcebergCatalogConfig::Rest { .. } | IcebergCatalogConfig::Glue { .. } => {
             Err(iceberg::Error::new(
-                iceberg::ErrorKind::Unexpected,
+                iceberg::ErrorKind::FeatureUnsupported,
                 "Only File catalog is supported currently",
             ))
         }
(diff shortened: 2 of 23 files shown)


peterxcli:
I want to leave the persisted config parsing into another PR.

peterxcli:
cc @dentiny PTAL, thanks!

src/moonlink_metadata_store/src/config_utils.rs
```diff
@@ -119,7 +119,11 @@ pub(crate) fn parse_moonlink_table_config(
     let mooncake_config = moonlink_table_config.mooncake_table_config;
     let persisted = MoonlinkTableConfigForPersistence {
         iceberg_table_config: IcebergTableConfigForPersistence {
-            warehouse_uri: iceberg_config.accessor_config.get_root_path(),
+            warehouse_uri: iceberg_config
+                .catalog
+                .get_file_catalog_accessor_config()
```
dentiny: just heads up, be sure NOT to use unwrap when you work on persistence

dentiny (commented):
Before this PR, I'm wondering if it's better to differentiate iceberg metadata file access vs iceberg data file access?

dentiny:
Some context:
- when we do iceberg sync, we first upload data files to storage, then reflect metadata change to catalog
- when we do iceberg load, we first ask catalog for metadata, then access data files
- so to me, there're actually two configs here, one for data file storage, another for catalog
- i think I made sth bad in the current implementation, I simply assume data file and metadata file have the same storage access

src/moonlink/src/storage/iceberg/catalog_utils.rs
```diff
@@ -12,17 +12,37 @@ use iceberg::Result as IcebergResult;
...
     iceberg_schema: IcebergSchema,
 ) -> IcebergResult<Box<dyn MoonlinkCatalog>> {
-    Ok(Box::new(FileCatalog::new(accessor_config, iceberg_schema)?))
+    match catalog {
+        IcebergCatalogConfig::File { accessor_config } => {
+            Ok(Box::new(FileCatalog::new(accessor_config, iceberg_schema)?))
+        }
+        IcebergCatalogConfig::Rest { .. } | IcebergCatalogConfig::Glue { .. } => {
+            Err(iceberg::Error::new(
+                iceberg::ErrorKind::Unexpected,
```
dentiny:
```suggestion
                iceberg::ErrorKind::FeatureUnsupported,
```

src/moonlink/src/storage/iceberg/catalog_utils.rs
```diff
@@ -12,17 +12,37 @@ use iceberg::Result as IcebergResult;
...
+    catalog: IcebergCatalogConfig,
 ) -> IcebergResult<Box<dyn MoonlinkCatalog>> {
-    Ok(Box::new(FileCatalog::new_without_schema(accessor_config)?))
+    match catalog {
+        IcebergCatalogConfig::File { accessor_config } => {
+            Ok(Box::new(FileCatalog::new_without_schema(accessor_config)?))
+        }
+        IcebergCatalogConfig::Rest { .. } | IcebergCatalogConfig::Glue { .. } => {
+            Err(iceberg::Error::new(
+                iceberg::ErrorKind::Unexpected,
```
dentiny:
```suggestion
                iceberg::ErrorKind::FeatureUnsupported,
```

src/moonlink/src/storage/iceberg/iceberg_table_config.rs
```diff
@@ -1,13 +1,82 @@
...
+    },
+}
+
+impl IcebergCatalogConfig {
+    pub fn get_warehouse_uri(&self) -> String {
+        match self {
+            IcebergCatalogConfig::File { accessor_config } => accessor_config.get_root_path(),
+            IcebergCatalogConfig::Rest {
+                rest_catalog_config,
+            } => rest_catalog_config.warehouse.clone().unwrap_or_default(),
```
dentiny: curious what's the default value?

I'm actually thinking add a validation when we create iceberg table manager, to check if rest / glue catalog certain fields have to properly assigned, so later we could unwrap it without any concern

src/moonlink/src/storage/iceberg/catalog_utils.rs
dentiny: - unexpected is generally interpreted as "unknown" and a retryable error
- but feature unsupported is a clearly non-retryable error

src/moonlink/src/storage/iceberg/compaction_tests.rs
```diff
@@ -504,8 +505,8 @@ async fn test_compaction_1_2_1() {
         &snapshot,
         &iceberg_table_manager_to_load
             .config
-            .accessor_config
-            .get_root_path(),
+            .metadata_accessor_config
```
dentiny, line 508: sorry I think it's my bad, if you take a look at the implementation, it's actually asking for data files "root directory", but not "metadata warehouse uri".

could you please do me a favor to update the `validate_recovered_snapshot` param name? thanks!

dentiny: ditto all below

src/moonlink/src/storage/iceberg/iceberg_table_config.rs
```diff
@@ -1,13 +1,82 @@
 use crate::{storage::filesystem::accessor_config::AccessorConfig, StorageConfig};
+use std::collections::HashMap;
+
+#[derive(Clone, Debug, PartialEq)]
```
dentiny: You need serde feature as well, since these configs need to be persisted into metadata storage

dentiny: You also need to think about schema evolution mechanisms, like how to handle new columns, columns to drop, column rename, etc; ping me if you want a discussion :)

src/moonlink/src/storage/iceberg/iceberg_table_config.rs
```diff
@@ -1,13 +1,82 @@
...
+    Rest {
+        rest_catalog_config: RestCatalogConfig,
+    },
+    Glue {
+        glue_catalog_config: GlueCatalogConfig,
+    },
+}
+
+impl IcebergCatalogConfig {
+    pub fn get_warehouse_uri(&self) -> String {
```
dentiny, line 64: I strongly suggest to add a `validate` function to check all fields are properly assigned, which gets invoked at iceberg table manager creation.

src/moonlink/src/storage/iceberg/iceberg_table_config.rs
```diff
@@ -1,13 +1,82 @@
...
+    name: Option<String>,
+    uri: Option<String>,
+    catalog_id: Option<String>,
+    warehouse: String,
+    props: HashMap<String, String>,
+}
+
+#[derive(Clone, Debug, PartialEq)]
+pub enum IcebergCatalogConfig {
+    File {
```
dentiny: let's hide rest and glue into features, so they're not enabled by default (to reduce necessary dependency).

src/moonlink/src/storage/iceberg/iceberg_table_config.rs
```diff
@@ -1,13 +1,82 @@
...
 
 #[derive(Clone, Debug, PartialEq)]
 pub struct IcebergTableConfig {
     /// Namespace for the iceberg table.
     pub namespace: Vec<String>,
     /// Iceberg table name.
     pub table_name: String,
-    // Accessor config.
-    pub accessor_config: AccessorConfig,
+    /// Accessor config for writing data files.
```
dentiny: not only write, but not also read, what about just "accessing"?

src/moonlink/src/storage/iceberg/compaction_tests.rs
dentiny: Correction:
- warehouse uri is indeed the data file storage: https://iceberg.apache.org/docs/1.6.0/aws/?h=warehou#warehouse-location
- it's a required catalog property

src/moonlink/src/storage/iceberg/iceberg_table_config.rs
```diff
@@ -1,13 +1,118 @@
...
+
+#[derive(Clone, Debug, PartialEq, Serialize, Deserialize)]
+pub struct GlueCatalogConfig {
+    #[serde(rename = "name")]
+    #[serde(default)]
+    name: Option<String>,
+
+    #[serde(rename = "uri")]
+    #[serde(default)]
+    uri: Option<String>,
```
dentiny, line 28: nit: add a newliner in between to keep consistency

src/moonlink/src/storage/iceberg/iceberg_table_config.rs
```diff
@@ -1,13 +1,118 @@
...
+
+    #[serde(rename = "props")]
+    #[serde(default)]
+    props: HashMap<String, String>,
+}
+
+#[derive(Clone, Debug, PartialEq, Serialize, Deserialize)]
+pub enum IcebergCatalogConfig {
+    #[serde(rename = "file")]
+    File { accessor_config: AccessorConfig },
```
dentiny: nit:
```rs
pub type FileCatalogConfig = AccessorConfig;
```
to keep naming consistency with other catalogs.

src/moonlink_metadata_store/src/config_utils.rs
```diff
@@ -141,7 +141,9 @@ pub(crate) fn parse_moonlink_table_config(
 
     // Extract table secret entry.
     let iceberg_secret_entry = iceberg_config
-        .accessor_config
+        .metadata_accessor_config
```
dentiny, line 144: We should leave a github issue on secret update: currently we only have secret for a mixed concept of iceberg config, but we will split them into two.

src/moonlink_connectors/src/pg_replicate/table_init.rs
```diff
@@ -167,7 +167,10 @@ pub async fn build_table_components(
         table_components.read_state_filepath_remap,
     );
     let table_status_reader = TableStatusReader::new(
-        &table_components.moonlink_table_config.iceberg_table_config,
+        &table_components
```
dentiny: the constructor takes a reference, so we don't need to take clone here?

src/moonlink/src/storage/iceberg/iceberg_table_config.rs
```diff
@@ -1,13 +1,118 @@
...
+    pub fn get_file_catalog_accessor_config(&self) -> Option<AccessorConfig> {
+        match self {
+            IcebergCatalogConfig::File { accessor_config } => Some(accessor_config.clone()),
+            #[cfg(any(feature = "catalog-rest", feature = "catalog-glue"))]
+            _ => None,
+        }
+    }
+
+    #[cfg(feature = "catalog-rest")]
+    pub fn get_rest_catalog_config(&self) -> Option<RestCatalogConfig> {
```
dentiny, line 87: curious how do we plan to use the function?

peterxcli: Sorry actually it's currently useless

peterxcli: I just adding them with get_file_catalog_accessor_config by the way...

src/moonlink_metadata_store/src/config_utils.rs
peterxcli: done. https://github.com/Mooncake-Labs/moonlink/issues/1753

src/moonlink_connectors/src/pg_replicate/table_init.rs
peterxcli: sure, thanks

src/moonlink/Cargo.toml
```diff
@@ -31,6 +31,9 @@ storage-gcs = [
     "reqwest",
 ]
 
+catalog-rest = []
```
dentiny, line 34: FYI: I assume these two features need to be used in metadata storage

src/moonlink/src/storage/iceberg/iceberg_table_config.rs
dentiny: I would prefer mark it as `#[allow(dead_code)]`, since it's easy to grep and revisit; otherwise public functions never generates unused warning and no one knows whether it's actually used without inspecting the code.

src/moonlink/src/storage/iceberg/iceberg_table_config.rs
```diff
@@ -1,13 +1,121 @@
...
+
+    #[serde(rename = "props")]
+    #[serde(default)]
+    props: HashMap<String, String>,
+}
+
+pub type FileCatalogConfig = AccessorConfig;
+
+#[derive(Clone, Debug, PartialEq, Serialize, Deserialize)]
+pub enum IcebergCatalogConfig {
```
dentiny, line 46: I don't think it's a problem for now, just heads up.
For all serde-able classes, it could be part of other serde-able classes, when unspecified in, say, json format, default value will be filled in on deserialization.

If you have any field needs customized default value, you have to provide your own `default` override.

Issue #1240, opened 2025-08-06
[ISSUE] Parallelize initial copy


nbiscaro:
## Summary

Instead of copying in one large chunk we should segment the table and copy each part in parallel. 

## Context

Why is this issue important? Link any relevant discussions, designs, or documents.

## Tasks

- [ ] 

## Notes

Any extra notes or details that might be useful.

Issue #1239, opened 2025-08-06
[ISSUE] Initial copy direct to file


nbiscaro:
## Summary

Currently we use the existing append path during initial copy. There's not really any need for this and we should write rows directly to file. 

Cur

## Context

Why is this issue important? Link any relevant discussions, designs, or documents.

## Tasks

- [ ] 

## Notes

Any extra notes or details that might be useful.

PR #1736, opened 2025-08-26
Initial load direct-to-disk optimization

nbiscaro:
## Summary

Instead of sending each row from the initial load through the table handler we can write these directly to parquet and use the batch_ingest API to integrate the files with moonlink.

We consume rows from the initial copy stream and add them to an arrow batch builder. When a new batch is ready, it's sent over an `mpsc` channel to a parquet writer. 

The reason for the `mpsc` channel is two-fold: 
First, it gives us natural backpressure for free. In the event that our copy stream is producing batches much faster than our writer(s) can consume them, the channel will apply back pressure on the copy stream and prevent memory bloat. 
Second, it decouples the producers from the consumers making parallelization easier. We can have a collection of producers appending to the queue, and a pool of writers grabbing batches off the channel and writing them (to be implemented as part of https://github.com/Mooncake-Labs/moonlink/issues/1240

When the initial load stream completes, we send a `TableEvent::LoadFiles` with the newly written parquets. We then reuse the `batch_ingest` API to integrate the files. During the ingest we check if the table is not append only, and if not we rescan the parquet files and build the index, attaching this to the commit for integration with moonlink.

## Related Issues

Closes [#1239](https://github.com/Mooncake-Labs/moonlink/issues/1239)
## Changes

- 
- 
- 

## Checklist

- [x] Code builds correctly

### DIFF
--- a/src/moonlink_connectors/src/pg_replicate/initial_copy.rs
+++ b/src/moonlink_connectors/src/pg_replicate/initial_copy.rs
@@ -1,6 +1,8 @@
+use crate::pg_replicate::conversions::table_row::TableRow;
 use crate::pg_replicate::initial_copy_writer::{
     create_batch_channel, ArrowBatchBuilder, InitialCopyWriterConfig, ParquetFileWriter,
 };
+use crate::pg_replicate::postgres_source::TableCopyStreamError;
 use crate::pg_replicate::postgres_source::{PostgresSource, TableCopyStream};
 use crate::pg_replicate::table::{ColumnSchema, LookupKey, SrcTableId, TableName, TableSchema};
 use crate::pg_replicate::util::postgres_schema_to_moonlink_schema;
@@ -34,12 +36,7 @@ pub async fn copy_table_stream<S>(
     config: Option<InitialCopyWriterConfig>,
 ) -> Result<CopyProgress>
 where
-    S: Stream<
-        Item = std::result::Result<
-            crate::pg_replicate::conversions::table_row::TableRow,
-            crate::pg_replicate::postgres_source::TableCopyStreamError,
-        >,
-    >,
+    S: Stream<Item = std::result::Result<TableRow, TableCopyStreamError>>,
 {
     // Convert PostgreSQL schema to Arrow schema
     let (arrow_schema, _identity_prop) = postgres_schema_to_moonlink_schema(&table_schema);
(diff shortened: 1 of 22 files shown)


src/moonlink_connectors/src/pg_replicate/initial_copy.rs
```diff
@@ -20,87 +28,109 @@ pub struct CopyProgress {
...
+    config: Option<InitialCopyWriterConfig>,
+) -> Result<CopyProgress>
+where
+    S: Stream<Item = std::result::Result<TableRow, TableCopyStreamError>>,
+{
+    // Convert PostgreSQL schema to Arrow schema
+    let (arrow_schema, _identity_prop) = postgres_schema_to_moonlink_schema(&table_schema);
+    let arrow_schema = Arc::new(arrow_schema);
+
+    // Create output directory for initial copy files
```
dentiny, line 45: It doesn't work with standalone deployment, which requires all files to be placed under backend `base_path`.

dentiny: It passes CI since we don't have UT for initial copy here: https://github.com/Mooncake-Labs/moonlink/blob/main/src/moonlink_service/src/test.rs

src/moonlink_connectors/src/pg_replicate/initial_copy_writer.rs
```diff
@@ -0,0 +1,379 @@
...
+        while let Some(batch) = rx.recv().await {
+            if batch.num_columns() == 0 || batch.num_rows() == 0 {
+                continue;
+            }
+
+            if writer.is_none() {
+                let path = self.next_file_path();
+                // Ensure directory exists
+                if let Some(parent) = path.parent() {
+                    tokio::fs::create_dir_all(parent).await.ok();
```
dentiny, line 148: Emmm shouldn't we handle possible error?

src/moonlink_connectors/src/pg_replicate/initial_copy.rs
```diff
@@ -162,69 +192,152 @@ mod tests {
...
+        // LoadFiles with multiple files
+        let evt = timeout(Duration::from_millis(500), rx.recv())
+            .await
+            .expect("channel lag")
+            .expect("sender closed");
+        let files = match evt {
+            TableEvent::LoadFiles { files, lsn } => {
+                assert_eq!(lsn, start_lsn);
+                assert!(
+                    files.len() >= 2,
```
dentiny, line 300: `assert_ge!`

Issue #1737, opened 2025-08-27
error improvement: propagate rpc server error to client


dentiny:
## Summary

Example: https://github.com/Mooncake-Labs/moonlink/blob/7de7f5bf386106ab0a9ba107d5e0ef4289db3a5a/src/moonlink_service/src/rpc_server.rs#L107
Need to do the same thing for other places as well

fweilun:
I'll take this issue, thanks.

dentiny:
> I'll take this issue, thanks.

Thanks for saving my life!

fweilun:
Should I replace all instances of panic! with the ? operator to propagate errors instead?

dpxcc:
This is a duplicate to https://github.com/Mooncake-Labs/moonlink/issues/1431, that issue has more context

PR #1746, opened 2025-08-27
Improvement propagate rpc server error

fweilun:
## Summary

Change return type from T to Result<T> in moonlink_rpc.

## Related Issues

Closes https://github.com/Mooncake-Labs/moonlink/issues/1737

## Changes

- Change the return type
- silence all the warnings and errors

## Checklist

- [✓] Code builds correctly

### DIFF
--- a/src/moonlink_datafusion/src/table_provider.rs
+++ b/src/moonlink_datafusion/src/table_provider.rs
@@ -41,7 +41,7 @@ impl MooncakeTableProvider {
     pub async fn try_new(uri: &str, schema: String, table: String, lsn: u64) -> Result<Self> {
         let mut stream = UnixStream::connect(uri).await?;
         let table_schema = get_table_schema(&mut stream, schema.clone(), table.clone()).await?;
-        let table_schema = StreamReader::try_new(table_schema.as_slice(), None)?.schema();
+        let table_schema = StreamReader::try_new(table_schema?.as_slice(), None)?.schema();
         let scan = Arc::new(MooncakeTableScan::try_new(stream, schema, table, lsn).await?);
         Ok(Self {
             schema: table_schema,
@@ -218,7 +218,7 @@ impl MooncakeTableScan {
     ) -> Result<Self> {
         let metadata = scan_table_begin(&mut stream, schema.clone(), table.clone(), lsn).await?;
         let metadata: MooncakeTableMetadata =
-            bincode::decode_from_slice(&metadata, config::standard())?.0;
+            bincode::decode_from_slice(&metadata?, config::standard())?.0;
         Ok(Self {
             stream: Some(stream),
             schema,
--- a/src/moonlink_rpc/src/lib.rs
+++ b/src/moonlink_rpc/src/lib.rs
@@ -16,7 +16,7 @@ macro_rules! rpcs {
                 },)*
             }
 
-            $(pub async fn $func<S: AsyncRead + AsyncWrite + Unpin>(stream: &mut S, $($name: $type),*) -> Result<$res> {
+            $(pub async fn $func<S: AsyncRead + AsyncWrite + Unpin>(stream: &mut S, $($name: $type),*) -> Result<Result<$res>> {
                 write(stream, &Request::[<$func:camel>] { $($name),* }).await?;
                 read(stream).await
             })*
--- a/src/moonlink_service/src/rpc_server.rs
+++ b/src/moonlink_service/src/rpc_server.rs
@@ -1,6 +1,7 @@
 use crate::{error::Error, Result};
 use arrow_ipc::writer::StreamWriter;
 use moonlink_backend::MoonlinkBackend;
+use moonlink_backend::Result as BackendResult;
 use moonlink_rpc::{read, write, Request, Table};
 use std::collections::HashMap;
 use std::error::Error as _;
@@ -103,8 +104,8 @@ where
                 table,
                 lsn,
             } => {
-                backend.create_snapshot(database, table, lsn).await?;
-                write(&mut stream, &()).await?;
+                let resp = backend.create_snapshot(database, table, lsn).await;
+                write(&mut stream, &resp).await?;
             }
             Request::CreateTable {
                 database,
@@ -114,7 +115,7 @@ where
                 table_config,
             } => {
                 // Use default mooncake config, and local filesystem for storage layer.
-                backend
+                let resp = backend
                     .create_table(
                         database,
                         table,
@@ -123,45 +124,55 @@ where
                         table_config,
                         None, /* input_database */
                     )
-                    .await?;
-                write(&mut stream, &()).await?;
+                    .await;
+                write(&mut stream, &resp).await?;
             }
             Request::DropTable { database, table } => {
-                backend.drop_table(database, table).await?;
-                write(&mut stream, &()).await?;
+                let resp = backend.drop_table(database, table).await;
+
+                write(&mut stream, &resp).await?;
             }
             Request::GetParquetMetadatas { data_files } => {
-                let metadata = backend.get_parquet_metadatas(data_files).await?;
-                write(&mut stream, &metadata).await?;
+                let metadata_resp = backend.get_parquet_metadatas(data_files).await;
+                write(&mut stream, &metadata_resp).await?;
             }
             Request::GetTableSchema { database, table } => {
-                let database = backend.get_table_schema(database, table).await?;
-                let writer = StreamWriter::try_new(vec![], &database)?;
-                let data = writer.into_inner()?;
-                write(&mut stream, &data).await?;
+                let resp = async {
+                    let schema = backend.get_table_schema(database, table).await?;
+                    let writer = StreamWriter::try_new(vec![], &schema)?;
+                    writer.into_inner().map_err(Error::from)
+                }
+                .await;
+
+                write(&mut stream, &resp).await?;
             }
             Request::ListTables {} => {
-                let tables = backend.list_tables().await?;
-                let tables: Vec<Table> = tables
-                    .into_iter()
-                    .map(|table| Table {
-                        database: table.database,
-                        table: table.table,
-                        cardinality: table.cardinality,
-                        commit_lsn: table.commit_lsn,
-                        flush_lsn: table.flush_lsn,
-                        iceberg_warehouse_location: table.iceberg_warehouse_location,
-                    })
-                    .collect();
-                write(&mut stream, &tables).await?;
+                let tables_resp = backend.list_tables().await;
+                if let Err(e) = &tables_resp {
+                    write(&mut stream, &Err::<Vec<Table>, &moonlink_backend::Error>(e)).await?;
+                }
+                let tables_resp: BackendResult<Vec<Table>> = tables_resp.map(|table| {
+                    table
+                        .into_iter()
+                        .map(|table| Table {
+                            database: table.database,
+                            table: table.table,
+                            cardinality: table.cardinality,
+                            commit_lsn: table.commit_lsn,
+                            flush_lsn: table.flush_lsn,
+                            iceberg_warehouse_location: table.iceberg_warehouse_location,
+                        })
+                        .collect()
+                });
+                write(&mut stream, &tables_resp).await?;
             }
             Request::OptimizeTable {
                 database,
                 table,
                 mode,
             } => {
-                backend.optimize_table(database, table, &mode).await?;
-                write(&mut stream, &()).await?;
+                let resp = backend.optimize_table(database, table, &mode).await;
+                write(&mut stream, &resp).await?;
             }
             Request::ScanTableBegin {
                 database,
@@ -170,8 +181,9 @@ where
             } => {
                 let state = backend
                     .scan_table(database.to_string(), table.to_string(), Some(lsn))
-                    .await?;
-                write(&mut stream, &state.data).await?;
+                    .await;
+
+                write(&mut stream, &state.as_ref().map(|s| s.data.clone())).await?;
                 assert!(map.insert((database, table), state).is_none());
             }
             Request::ScanTableEnd { database, table } => {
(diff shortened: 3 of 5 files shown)


fweilun:
> I think you include rest catalog change in this PR :)

Yeah.. fixed it already.

src/moonlink_service/src/test.rs
```diff
@@ -597,7 +597,7 @@ async fn test_bulk_ingest_files() {
 
     // A dummy stub-level interface testing.
     let mut moonlink_stream = TcpStream::connect(MOONLINK_ADDR).await.unwrap();
-    load_files(
+    let _ = load_files(
```
dentiny: what about we directly unwrap?

src/moonlink_service/src/test.rs
```diff
@@ -462,7 +462,7 @@ async fn test_moonlink_standalone_file_insert() {
     assert!(puffin_deletion.is_empty());
     assert!(positional_deletion.is_empty());
 
-    scan_table_end(
+    let _ = scan_table_end(
```
dentiny: same here, maybe directly unwrap?

dentiny (commented):
curious do we need to return error from rpc server?

dentiny:
> curious do we need to return error from rpc server?

I'm even thinking, do you think it makes sense or overkill to make a mock backend, so we could do unit test on error propagation as well?

src/moonlink_rpc/src/lib.rs
```diff
@@ -25,16 +25,16 @@ macro_rules! rpcs {
...
+    create_snapshot(database: String, table: String, lsn: u64) -> Result<()>;
+    create_table(database: String, table: String, src: String, src_uri: String, table_config: String) -> Result<()>;
+    drop_table(database: String, table: String) -> Result<()>;
+    get_parquet_metadata(data_file: String) -> Result<Vec<u8>>;
+    get_table_schema(database: String, table: String) -> Result<Vec<u8>>;
+    list_tables() -> Result<Vec<Table>>;
+    load_files(database: String, table: String, files: Vec<String>) -> Result<()>;
+    optimize_table(database: String, table: String, mode: String) -> Result<()>;
+    scan_table_begin(database: String, table: String, lsn: u64) -> Result<Vec<u8>>;
+    scan_table_end(database: String, table: String) -> Result<()>;
```
dpxcc: Since we are going to return `Result<T>` for every RPC API. It would be better to hide `Result` part in macro, instead of repeating it for every RPC API

The changes would be something like from
```
read(stream).await
```
to
```
read(stream).await?
```
This also avoids all callees to manually`??`

fweilun:
> > curious do we need to return error from rpc server?
> 
> I'm even thinking, do you think it makes sense or overkill to make a mock backend, so we could do unit test on error propagation as well?

I'll do it this way.

src/moonlink_datafusion/src/table_provider.rs
```diff
@@ -41,7 +41,7 @@ impl MooncakeTableProvider {
     pub async fn try_new(uri: &str, schema: String, table: String, lsn: u64) -> Result<Self> {
         let mut stream = UnixStream::connect(uri).await?;
         let table_schema = get_table_schema(&mut stream, schema.clone(), table.clone()).await?;
-        let table_schema = StreamReader::try_new(table_schema.as_slice(), None)?.schema();
+        let table_schema = StreamReader::try_new(table_schema?.as_slice(), None)?.schema();
```
dentiny: curious why do we prefer `table_schema?` at L44 over `table_schema??` at L43?

src/moonlink_service/src/rpc_server.rs
```diff
@@ -170,8 +178,9 @@ where
             } => {
                 let state = backend
                     .scan_table(database.to_string(), table.to_string(), Some(lsn))
-                    .await?;
-                write(&mut stream, &state.data).await?;
+                    .await;
+
+                write(&mut stream, &state.as_ref().map(|s| s.data.clone())).await?;
```
dentiny: curious why do we need copy here?

dentiny:
LG, just some rust questions which I'm not familiar with :sweat_smile:

dentiny:
@dpxcc for final review, since you requested change

src/moonlink_service/src/rpc_server.rs
fweilun: The assertion on line 184 requires access to the state value. Could you advise if there’s a better approach?

src/moonlink_datafusion/src/table_provider.rs
fweilun: I thought using ?? might be confusing for some people, but I'll use it.

src/moonlink_rpc/src/lib.rs
```diff
@@ -16,7 +16,7 @@ macro_rules! rpcs {
                 },)*
             }
 
-            $(pub async fn $func<S: AsyncRead + AsyncWrite + Unpin>(stream: &mut S, $($name: $type),*) -> Result<$res> {
+            $(pub async fn $func<S: AsyncRead + AsyncWrite + Unpin>(stream: &mut S, $($name: $type),*) -> Result<Result<$res>> {
```
dpxcc: https://github.com/Mooncake-Labs/moonlink/pull/1746#discussion_r2304636978
Instead of changing from `Result<$res>` to `Result<Result<$res>>`, just change from `read(stream).await` to `read(stream).await?` above
This avoids all `??` and `unwrap().unwrap()`

src/moonlink_service/src/rpc_server.rs
```diff
@@ -123,45 +124,52 @@ where
...
             }
             Request::GetTableSchema { database, table } => {
-                let database = backend.get_table_schema(database, table).await?;
-                let writer = StreamWriter::try_new(vec![], &database)?;
-                let data = writer.into_inner()?;
-                write(&mut stream, &data).await?;
+                let resp = async {
+                    let schema = backend.get_table_schema(database, table).await?;
+                    let writer = StreamWriter::try_new(vec![], &schema)?;
+                    writer.into_inner().map_err(Error::from)
```
dpxcc: nit:
```
Ok(writer.into_inner()?)
```

src/moonlink_service/src/rpc_server.rs
```diff
@@ -103,8 +104,8 @@ where
                 table,
                 lsn,
             } => {
-                backend.create_snapshot(database, table, lsn).await?;
-                write(&mut stream, &()).await?;
+                let resp = backend.create_snapshot(database, table, lsn).await;
```
dpxcc: nit: I would prefer `res` (short for `result`) over `resp`

src/moonlink_service/src/rpc_server.rs
dpxcc: One way to avoid copy is to wrap `Vec<u8>` inside `Cow`: https://doc.rust-lang.org/std/borrow/enum.Cow.html

src/moonlink_service/src/rpc_server.rs
```diff
@@ -123,45 +124,52 @@ where
...
+                            database: table.database,
+                            table: table.table,
+                            cardinality: table.cardinality,
+                            commit_lsn: table.commit_lsn,
+                            flush_lsn: table.flush_lsn,
+                            iceberg_warehouse_location: table.iceberg_warehouse_location,
+                        })
+                        .collect()
+                });
+                write(&mut stream, &tables_resp).await?;
```
dpxcc: This is also suspicious
You are sending `std::result::Result<Vec<Table>, moonlink_backend::Error>` but receiver expects `std::result::Result<Vec<Table>, moonlink_rpc::Error>` which doesn't seem to match
Have you tested with any client and verified your change works?

dpxcc: Same applies to other RPCs

src/moonlink_rpc/src/error.rs
```diff
@@ -18,10 +20,36 @@ pub enum Error {
...
+        ))
+    }
+}
+
+impl From<BackendError> for Error {
+    #[track_caller]
+    fn from(_source: BackendError) -> Self {
+        Error::Backend(ErrorStruct::new(
+            "Backend error".to_string(),
+            ErrorStatus::Permanent,
```
fweilun: Hi @dpxcc  — I’m running into a decode error whenever source is set.
DEBUG-1 is the value I write, DEBUG-2 is what I read back.
Could you advise how I should handle this issue? Thanks!

`[DEBUG-1] Err(Backend(ErrorStruct { message: "Backend error", status: Permanent, source: Some(Moonlink connector error (permanent) at /Users/user/Projects/moonlink/src/moonlink_backend/src/lib.rs:143:26, source: Table nonexistent-db.nonexistent-table not found (permanent) at src/moonlink_connectors/src/replication_manager.rs:293:28), location: Some("src/moonlink_service/src/rpc_server.rs:108:107") }))`

`[DEBUG-2] Err(Decode(ErrorStruct { message: "Decode error", status: Permanent, source: Some(UnexpectedVariant { type_name: "Option<T>", allowed: Range { min: 0, max: 1 }, found: 239 }), location: Some("/Users/user/Projects/moonlink/src/moonlink_rpc/src/lib.rs:56:8") }))`

src/moonlink_rpc/Cargo.toml
```diff
@@ -5,7 +5,9 @@ edition = { workspace = true }
 license = "MIT"
 
 [dependencies]
+arrow = { workspace = true }
 bincode = { workspace = true }
+moonlink_backend = { workspace = true }
```
dpxcc: moonlink_rpc is a thin crate shared by client and server, and should not include moonlink_backend

src/moonlink_rpc/src/error.rs
dpxcc: It just means that you are writing one type, but trying to read another type
Take `ScanTableBegin` as an example
```
write(&mut stream, &state.as_ref().map(|s| Cow::Borrowed(&s.data))).await?;
```
is writing `Result<Cow<Vec<u8>>>` but the reader expects `Result<Vec<u8>>`

src/moonlink_service/src/test_utils.rs
```diff
@@ -678,6 +678,7 @@ pub(crate) async fn drop_table(client: &reqwest::Client, database: &str, table:
 }
 
 /// Util function to optimize table via REST API.
+#[allow(dead_code)]
```
dentiny: curious why this annotation?

src/moonlink_service/src/test.rs
```diff
@@ -103,6 +103,29 @@ async fn test_schema() {
     assert_eq!(response.lsn, 1);
 }
 
+#[tokio::test]
+#[serial]
+async fn test_rpc_error_propagation_nonexistent_table() {
```
dentiny: checkout the end of the file, we have a bunch of failure test cases

src/moonlink_rpc/Cargo.toml
```diff
@@ -5,6 +5,7 @@ edition = { workspace = true }
 license = "MIT"
 
 [dependencies]
+arrow = { workspace = true }
```
dentiny: is it used anywhere?

dentiny (commented):
CC @dpxcc

src/moonlink_service/src/test.rs
fweilun: Should I implement all the test using rpc propagation?
All the failure test cases are propagate via rest server.

fweilun: I'm also curious whether all the possible failures are included.

Cargo.lock
```diff
@@ -4716,6 +4716,7 @@ dependencies = [
 name = "moonlink_rpc"
 version = "0.0.1"
 dependencies = [
+ "arrow",
```
dpxcc: arrow is not used, remove

src/moonlink_service/src/rpc_server.rs
```diff
@@ -122,68 +130,104 @@ where
...
+                match backend
                     .scan_table(database.to_string(), table.to_string(), Some(lsn))
-                    .await?;
-                write(&mut stream, &state.data).await?;
-                assert!(map.insert((database, table), state).is_none());
+                    .await
+                    .map_err(into_backend_error)
+                {
+                    Ok(state) => {
+                        let res: RpcResult<Vec<u8>> = Ok(state.data.clone());
```
dpxcc, line 203: We can use `Cow` to avoid clone here
But we need to make sure we also change moonlink_rpc to be consistent
Not required for this diff

src/moonlink_service/src/rpc_server.rs
```diff
@@ -122,68 +130,104 @@ where
...
+                    .await
+                    .map_err(into_backend_error);
+                write(&mut stream, &metadata_res).await?;
             }
             Request::GetTableSchema { database, table } => {
-                let database = backend.get_table_schema(database, table).await?;
-                let writer = StreamWriter::try_new(vec![], &database)?;
-                let data = writer.into_inner()?;
-                write(&mut stream, &data).await?;
+                let res: RpcResult<Vec<u8>> = (async {
```
dpxcc: `handle_stream` could use some cleanup
I haven’t thought much about how yet, but the code is noticeably harder to read after this diff

fweilun: Thanks for the review. I did some clean-up, and I think it looks better now.

src/moonlink_service/src/test.rs
dentiny: I mean, we should put error cases at the end of the file (i already separate them by adding intentional newliners), or feel free to create another test file.

fweilun: ohh, ic.

dpxcc (approved):
thx for working on this!

dentiny:
@fweilun do you mind rebasing main?

Issue #1748, opened 2025-08-27
BUG: stream batch id assignment crash


dentiny:
## Summary

```rs
#[tokio::test]
async fn test_streaming_batch_id_assignment() -> Result<()> {
    let context = TestContext::new("streaming_batch_id_assignment");
    let mut table = test_table(&context, "lsn_table", IdentityProp::FullRow).await;
    let (event_completion_tx, mut event_completion_rx) = mpsc::channel(100);
    table.register_table_notify(event_completion_tx).await;

    let xact_id_1 = 1;
    let xact_id_2 = 2;

    // Start streaming transactions
    table.append_in_stream_batch(test_row(1, "A", 20), xact_id_1)?;
    table.append_in_stream_batch(test_row(2, "B", 21), xact_id_2)?;

    // Flush streaming transactions with different LSNs (can be out of order)
    let disk_slice_1 =
        flush_stream_and_sync_no_apply(&mut table, &mut event_completion_rx, xact_id_1, Some(50))
            .await
            .expect("Disk slice 1 should be present");
    table.commit_transaction_stream_impl(xact_id_1, /*lsn=*/ 50).unwrap();

    let disk_slice_2 =
        flush_stream_and_sync_no_apply(&mut table, &mut event_completion_rx, xact_id_2, Some(100))
            .await
            .expect("Disk slice 2 should be present");
    table.commit_transaction_stream_impl(xact_id_2, /*lsn=*/ 100).unwrap();

    // Verify both streaming LSNs are tracked
    assert!(table.ongoing_flush_lsns.contains(&100));
    assert!(table.ongoing_flush_lsns.contains(&50));
    assert_eq!(table.get_min_ongoing_flush_lsn(), 50);

    // Complete streaming flushes
    table.apply_stream_flush_result(
        xact_id_1,
        disk_slice_1,
        uuid::Uuid::new_v4(), /*placeholder*/
    );
    table.apply_stream_flush_result(
        xact_id_2,
        disk_slice_2,
        uuid::Uuid::new_v4(), /*placeholder*/
    );

    // Verify all pending flushes are cleared
    assert!(table.ongoing_flush_lsns.is_empty());
    assert_eq!(table.get_min_ongoing_flush_lsn(), u64::MAX);

    Ok(())
}
```
gets error message
```sh
---- storage::mooncake_table::tests::test_streaming_batch_id_assignment stdout ----

[...]

PR #1749, opened 2025-08-27
[storage] Fix batch id assignment

dentiny:
## Summary

We could have multiple streaming transaction ongoing, current implementation doesn't guarantee each batch gets their unique id. See linked issue for details.

## Related Issues

Closes https://github.com/Mooncake-Labs/moonlink/issues/1748

## Checklist

- [x] Code builds correctly

### DIFF
--- a/src/moonlink/src/storage/mooncake_table/batch_id_counter.rs
+++ b/src/moonlink/src/storage/mooncake_table/batch_id_counter.rs
@@ -34,6 +34,8 @@ impl BatchIdCounter {
         self.counter.load(Ordering::Relaxed)
     }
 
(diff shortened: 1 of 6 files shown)


src/moonlink/src/storage/mooncake_table/batch_id_counter.rs
```diff
@@ -34,6 +34,8 @@ impl BatchIdCounter {
         self.counter.load(Ordering::Relaxed)
     }
 
+    // Increment the id by 1, and return the id before change.
```
dentiny, line 37: This should be clearly documented, I spent quite a while to debug.

Issue #1622, opened 2025-08-20
BUG: append only table unit test fail


dentiny:
## Summary

```sh
        SLOW [> 60.000s] moonlink table_handler::tests::test_append_only_table_high_volume
        SLOW [>120.000s] moonlink table_handler::tests::test_append_only_table_high_volume
        SLOW [>180.000s] moonlink table_handler::tests::test_append_only_table_high_volume
        SLOW [>240.000s] moonlink table_handler::tests::test_append_only_table_high_volume
        SLOW [>300.000s] moonlink table_handler::tests::test_append_only_table_high_volume
        SLOW [>360.000s] moonlink table_handler::tests::test_append_only_table_high_volume
        SLOW [>420.000s] moonlink table_handler::tests::test_append_only_table_high_volume
```

PR #1751, opened 2025-08-27
[storage] Handle multiple ongoing flushes for streaming transaction

dentiny:
## Summary

Currently we use a set data structure to manage ongoing flush operations: https://github.com/Mooncake-Labs/moonlink/blob/c178bfb7e7d80127eed7fddcb8fd1113e8512b54/src/moonlink/src/storage/mooncake_table.rs#L464-L465
but for streaming transaction we could have multiple flushes in the background.

A detailed example,
- LSN = 13, streaming txn, no commit LSN, issue flush
- LSN = 16, non-streaming txn, issue flush
- LSN = 13, streaming txn, has commit LSN, issue flush

After that we have two ongoing flush LSNs, one for 13, another for 16.
The completion order is as below:
- LSN = 13
- LSN = 16

When the first flush finish, the lowest LSN is 16, which means we set flush LSN to 16 and will dump iceberg with LSN 16.
This PR fixes the issue by use a <LSN, counter> map to record all ongoing flush operations.

Another thing confuses me a lot is the unit tests, which seems to test on cases which will never happen in production.
In production codepath, we will have only two stream flush:
- on append, when threshold reaches, we trigger a flush but no commit, with no commit LSN
- on commit, we force a flush and commit with LSN

To simplify the commit / flush / LSN, this PR rewrote some of the unit tests to conform production behavior.

On flush operation, the new behavior looks like
- we only record ongoing flush LSNs to ongoing LSN map on commit; otherwise it's completely invisible to other transactions
- on commit / flush, we record all ongoing flush count to the map, which will be decremented when each of them finishes

Notice: I didn't add any new unit tests in the PR, since a simply flush LSN progression assertion already fails a few existing ones...

## Checklist

- [x] Code builds correctly

### DIFF
--- a/src/moonlink/src/storage/mooncake_table.rs
+++ b/src/moonlink/src/storage/mooncake_table.rs
@@ -840,6 +840,10 @@ impl MooncakeTable {
     }
 
     /// Flushes the disk slice for the transaction.
+    ///
+    /// # Arguments
+    ///
+    /// * ongoing_flush_count: used to increment ongoing flush count for the given LSN.
     fn flush_disk_slice(
         &mut self,
         disk_slice: &mut DiskSliceWriter,
--- a/src/moonlink/src/storage/mooncake_table.rs
+++ b/src/moonlink/src/storage/mooncake_table.rs
@@ -65,7 +65,7 @@ use delete_vector::BatchDeletionVector;
 pub(crate) use disk_slice::DiskSliceWriter;
 use mem_slice::MemSlice;
 pub(crate) use snapshot::SnapshotTableState;
-use std::collections::{BTreeSet, HashMap, HashSet};
+use std::collections::{BTreeMap, HashMap, HashSet};
 use std::path::PathBuf;
 use std::sync::Arc;
 use table_snapshot::{IcebergSnapshotImportResult, IcebergSnapshotIndexMergeResult};
@@ -462,7 +462,8 @@ pub struct MooncakeTable {
     wal_manager: WalManager,
 
     /// LSN of ongoing flushes.
-    pub ongoing_flush_lsns: BTreeSet<u64>,
+    /// Maps from LSN to its count.
+    pub ongoing_flush_lsns: BTreeMap<u64, u32>,
 
     /// Table replay sender.
     event_replay_tx: Option<mpsc::UnboundedSender<MooncakeTableEvent>>,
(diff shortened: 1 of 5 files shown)


src/moonlink/src/storage/mooncake_table/tests.rs
```diff
@@ -737,10 +737,14 @@ async fn test_streaming_begin_flush_commit_end_flush() {
 
     // Begin the flush
     // This will drain the mem slice and add its relevant state to the stream state
-    let disk_slice =
```
dentiny, line 740: why do we have flush LSN > commit LSN?

src/moonlink/src/storage/mooncake_table/tests.rs
```diff
@@ -808,25 +812,25 @@ async fn test_streaming_begin_flush_commit_end_flush_multiple() {
 
     // Begin the flush
     // This will drain the mem slice and add its relevant state to the stream state
-    let disk_slice1 =
-        flush_stream_and_sync_no_apply(&mut table, &mut event_completion_rx, xact_id, Some(lsn))
```
dentiny, line 812: in production codepath, flush stream only has commit LSN on commit

src/moonlink/src/storage/mooncake_table/transaction_stream.rs
```diff
@@ -417,12 +417,6 @@ impl MooncakeTable {
                 .unwrap();
         }
 
-        // Perform table flush completion notification.
-        if let Some(lsn) = disk_slice.lsn() {
```
dentiny, line 421: This is sth confused me a lot, for some disk slices, we call `try_set_next_flush_lsn` for twice
- L423
- L449

After this PR refactor, flush LSN is only set at most once when transaction already committed.

src/moonlink/src/storage/mooncake_table.rs
```diff
@@ -915,31 +922,41 @@ impl MooncakeTable {
     fn try_set_next_flush_lsn(&mut self, lsn: u64) {
         let min_pending_lsn = self.get_min_ongoing_flush_lsn();
         if lsn < min_pending_lsn {
+            if let Some(old_flush_lsn) = self.next_snapshot_task.new_flush_lsn {
```
dentiny, line 925: There's another bug (unrelated to this one): suppose multiple flush ongoing, if a flush operation with high LSN finishes, it cannot be reflected to `new_flush_lsn` since there're lower flushes ongoing, we should keep their information (LSN) in a separate data structure, otherwise it gets lost forever and leads to data inconsistency. I will fix it in the followup PR.

dentiny:
The failed unit test is known to be flaky, I already filed a ticket for it: https://github.com/Mooncake-Labs/moonlink/issues/1622

PR #1761, opened 2025-08-27
Reestablish pg connection

nbiscaro:
## Summary

We don't currently implement any reconnection logic in the event that we lose our connection to Postgres. This is true for the replication task as well as any control commands we send (ALTER TABLE/PUB, etc.)

In the case of control commands to PG, instead of holding a client across commands that may at some point disconnect, we should just create and drop ephemeral PG clients as we need. The command events are rare enough that this shouldn't be a performance concern and ensures the PG client remains fresh. It also prevents any lingering unused connections to PG. 

In the case the replication task goes down, for each control command we check the status of the replication task and reestablish if not currently active. 

## Related Issues

## Changes

- 
- 
- 

## Checklist

- [ ] Code builds correctly

### DIFF
--- a/src/moonlink_connectors/src/pg_replicate.rs
+++ b/src/moonlink_connectors/src/pg_replicate.rs
@@ -244,6 +244,15 @@ impl PostgresConnection {
         }
     }
 
+    /// Ensure the command channel exists (recreate if the receiver was taken by a finished task).
+    fn ensure_cmd_channel(&mut self) {
+        if self.cmd_rx.is_none() {
+            let (tx, rx) = mpsc::channel(8);
+            self.cmd_tx = tx;
+            self.cmd_rx = Some(rx);
+        }
+    }
+
     pub fn retry_drop(uri: &str, drop_query: &str) -> JoinHandle<Result<()>> {
         let uri = uri.to_string();
         let drop_query = drop_query.to_string();
@@ -309,7 +318,7 @@ impl PostgresConnection {
 
     /// Add table to PostgreSQL replication
     pub async fn add_table_to_replication(
-        &self,
+        &mut self,
         src_table_id: SrcTableId,
         schema: TableSchema,
         event_sender: mpsc::Sender<TableEvent>,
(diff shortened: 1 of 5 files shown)


src/moonlink_connectors/src/pg_replicate.rs
```diff
@@ -130,12 +128,47 @@ impl PostgresConnection {
...
+        Ok(())
+    }
+
     /// Include full row in cdc stream (not just primary keys).
     pub async fn alter_table_replica_identity(&self, table_name: &str) -> Result<()> {
-        self.postgres_client
-            .simple_query(&format!("ALTER TABLE {table_name} REPLICA IDENTITY FULL;"))
-            .await?;
-        Ok(())
+        let query = format!("ALTER TABLE {table_name} REPLICA IDENTITY FULL;");
```
dpxcc, line 170: What about other control commands, e.g. `ALTER PUBLICATION ADD/DROP TABLE` etc?

Also, it looks like you’re moving toward a model where each control command uses its own connection. That could be problematic, especially when users create a large number of tables (we’ve had requests for ~3k tables). You might run into similar issues as below:

> When pg_mooncake runs a query, it opens many connections to read remote files. Each connection takes a temporary port (in the range 0–65535). Even after the connection closes, the OS keeps that port busy for about 60 seconds to prevent delayed packets from being misread
>
> If too many connections are created in a short time, we use up all the available ports. Once that happens, even a single query can fail until ports are freed again

nbiscaro:
Closing in favor of https://github.com/Mooncake-Labs/moonlink/pull/1781

Track overall task progress https://github.com/Mooncake-Labs/moonlink/issues/1778

src/moonlink_connectors/src/pg_replicate.rs
nbiscaro: Agree we should take a different approach. See https://github.com/Mooncake-Labs/moonlink/pull/1761#issuecomment-3234581589

PR #1776, opened 2025-08-28
Implement Upsert/ Delete if exist for event API

zhousun:
## Summary
If the table have primary key,
regular INSERT/ UPDATE/ DELETE could error and event api is usually not designed to handle this kind of failure.

Introducing Upsert /Delete if not exists, which provide simpler semantics and will success in all cases.

Upsert is implemented as append + delete_if_not_exist.

## Related Issues

## Changes

- 
- 
- 

## Checklist

- [ ] Code builds correctly

### DIFF
--- a/src/moonlink/src/storage/mooncake_table/snapshot.rs
+++ b/src/moonlink/src/storage/mooncake_table/snapshot.rs
@@ -777,6 +777,7 @@ impl SnapshotTableState {
         index_lookup_result: Vec<RecordLocation>,
         file_id_to_lsn: &HashMap<FileId, u64>,
         batch_id_to_lsn: &HashMap<u64, u64>,
+        delete_if_exists: bool,
     ) -> Vec<ProcessedDeletionRecord> {
         let mut candidates: Vec<RecordLocation> = index_lookup_result
             .into_iter()
@@ -790,6 +791,30 @@ impl SnapshotTableState {
                     )
             })
             .collect();
+
+        if delete_if_exists {
+            let mut processed_deletions = Vec::new();
+            assert!(deletions.len() == 1);
+            let deletion = deletions.first().unwrap();
+            if deletion.row_identity.is_none() {
+                let candidate = candidates.pop();
+                if let Some(candidate) = candidate {
+                    processed_deletions.push(Self::build_processed_deletion(deletion, candidate));
+                }
+            } else {
+                for loc in &candidates {
+                    if self
+                        .matches_identity(loc, deletion.row_identity.as_ref().unwrap())
+                        .await
+                    {
+                        processed_deletions
+                            .push(Self::build_processed_deletion(deletion, loc.clone()));
+                        break;
+                    }
+                }
+            }
+            return processed_deletions;
+        }
         // This optimization is important when working with table without primary key.
         // Postgres never distinguish row with same value, so they will almost always be processed together.
         // Thus we can avoid full row identity comparison if we also process them together.
@@ -988,6 +1013,17 @@ impl SnapshotTableState {
         if new_deletions.is_empty() {
             return;
         }
+        // moonlink don't support mix delete_if_exists and not delete_if_exists for now.
+        //
+        let delete_if_exists = new_deletions
+            .iter()
+            .any(|deletion| deletion.delete_if_exists);
+        assert_eq!(
+            delete_if_exists,
+            new_deletions
+                .iter()
+                .all(|deletion| deletion.delete_if_exists)
+        );
         let mut index_lookup_result = self
             .current_snapshot
             .indices
@@ -1006,7 +1042,9 @@ impl SnapshotTableState {
             }
             let deletions = &new_deletions[start_i..i];
             let mut lookup_result = Vec::new();
-            while index_lookup_result[j].0 != new_deletions[start_i].lookup_key {
+            while j < index_lookup_result.len()
+                && index_lookup_result[j].0 != new_deletions[start_i].lookup_key
+            {
                 j += 1;
             }
             let mut j_end = j;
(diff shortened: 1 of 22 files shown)


src/moonlink/src/storage/mooncake_table/snapshot.rs
```diff
@@ -790,6 +791,30 @@ impl SnapshotTableState {
                     )
             })
             .collect();
+
+        if delete_if_exists {
+            let mut processed_deletions = Vec::new();
+            assert!(deletions.len() == 1);
```
dentiny:
```suggestion
            assert_eq!(deletions.len(), 1);
```

src/moonlink/src/table_handler/chaos_test.rs
```diff
@@ -635,6 +635,7 @@ impl ChaosState {
                 row: self.get_random_row_to_delete(),
                 xact_id: self.get_cur_xact_id(),
                 lsn: self.get_and_update_cur_lsn(),
+                delete_if_exists: false,
```
dentiny: I will followup with new chaos test on delete if exists semantics

src/moonlink_service/src/rest_api.rs
```diff
@@ -505,15 +505,15 @@ async fn ingest_data(
     // Parse operation
     let operation = match payload.operation.as_str() {
         "insert" => RowEventOperation::Insert,
-        "update" => RowEventOperation::Update,
+        "upsert" => RowEventOperation::Upsert,
```
dentiny, line 508: Any chance to add a unit test here: https://github.com/Mooncake-Labs/moonlink/blob/main/src/moonlink_service/src/test.rs?
You could get LSN back, block wait with snapshot created, and check the result.

src/moonlink/src/storage/mooncake_table/tests.rs
```diff
@@ -3129,3 +3129,66 @@ async fn test_out_of_order_flush_for_iceberg_snapshot() {
...
+
+    // upsert a row
+    let row3 = test_row(1, "C", 22);
+    table.delete_if_exists(row3.clone(), 103).await;
+    table.append(row3.clone()).unwrap();
+    table.commit(104);
+    flush_table_and_sync(&mut table, &mut event_completion_rx, 104)
+        .await
+        .unwrap();
+
```
dentiny, line 3192: Shouldn't we check the result?

src/moonlink/src/table_handler/chaos_test.rs
zhousun: ++ added.

src/moonlink_service/src/rest_api.rs
zhousun: will do in a separate diff that enforce identity of event table.

src/moonlink/src/storage/mooncake_table/table_creation_test_utils.rs
```diff
@@ -212,11 +216,13 @@ pub(crate) fn get_random_identity(random_seed: u64, append_only: bool) -> Identi
...
     let mut rng = rand::rngs::StdRng::seed_from_u64(random_seed);
 
-    let identities = [
+    let mut identities = vec![
         IdentityProp::SinglePrimitiveKey(0),
         IdentityProp::Keys(vec![0]),
-        IdentityProp::FullRow,
     ];
+    if !upsert_delete_if_exists {
+        identities.push(IdentityProp::FullRow);
```
dentiny, line 224: qq: curious why do we disallow full row for upsert?

src/moonlink/src/storage/mooncake_table/tests.rs
```diff
@@ -3129,3 +3129,90 @@ async fn test_out_of_order_flush_for_iceberg_snapshot() {
...
+    // create a snapshot and verify the row is upserted
+    {
+        create_mooncake_snapshot_for_test(&mut table, &mut event_completion_rx).await;
```
dentiny, line 3195: we should make a util function for check, I will followup

src/moonlink/src/storage/mooncake_table/table_creation_test_utils.rs
zhousun: upsert full row is not well defined (it actually don't make any sense :) ).

src/moonlink/src/table_handler/chaos_test.rs
```diff
@@ -1500,7 +1549,29 @@ async fn test_append_only_chaos_on_local_fs_with_data_compaction() {
...
+            atomic_write_dir: None,
+        },
+    };
+    let env = TestEnvironment::new(test_env_config).await;
+    chaos_test_impl(env).await;
+}
+
+#[named]
+#[tokio::test(flavor = "multi_thread", worker_threads = 2)]
+async fn test_upsert_chaos_on_local_fs_with_data_compaction() {
```
dentiny, line 1567: You didn't enabled in the CI I think, I will followup

src/moonlink/src/table_handler/chaos_test.rs
```diff
@@ -417,6 +432,13 @@ impl ChaosState {
 
     /// Get a random row to delete.
     fn get_random_row_to_delete(&mut self) -> MoonlinkRow {
+        if self.is_upsert_table && self.rng.random_range(0..100) < 50 {
```
dentiny, line 435: The row returned here could be used in both streaming transaction or non-streaming transaction, but you only implement "delete if exists" semantics for non-streaming txn, I think it's breaking chaos test

src/moonlink/src/table_handler/chaos_test.rs
```diff
@@ -176,6 +176,8 @@ struct ChaosState {
     rng: StdRng,
     /// Whether to enable delete operations.
     append_only: bool,
+    /// Whether to test upsert/ delete if exists.
+    is_upsert_table: bool,
```
dentiny, line 180: The table property should be persisted, otherwise replay system has no way to replay with the same operation (delete vs delete_if_exists)

Issue #1793, opened 2025-08-29
BUG: background compaction / iceberg snapshot / mooncake snapshot


dentiny:
## Summary

```rs
#[tokio::test]
async fn test_debug() {
    // three disk files
    // two of them in iceberg already. one of them in mooncake snapshot

    // Local filesystem for iceberg.
    let iceberg_temp_dir = tempdir().unwrap();
    let iceberg_table_config = get_iceberg_table_config(&iceberg_temp_dir);

    // Local filesystem to store write-through cache.
    let table_temp_dir = tempdir().unwrap();
    let local_table_directory = table_temp_dir.path().to_str().unwrap().to_string();

    // Create mooncake metadata.
    let file_index_config = FileIndexMergeConfig::disabled();
    let data_compaction_config = DataCompactionConfig {
        min_data_file_to_compact: 2,
        max_data_file_to_compact: u32::MAX,
        data_file_final_size: u64::MAX,
        data_file_deletion_percentage: 0,
    };
    let iceberg_persistence_config = IcebergPersistenceConfig {
        new_data_file_count: 1,
        new_committed_deletion_log: 1,
        new_compacted_data_file_count: 1,
        old_compacted_data_file_count: 1,
        old_merged_file_indices_count: usize::MAX,
    };
    let mut config = MooncakeTableConfig::new(table_temp_dir.path().to_str().unwrap().to_string());
    config.file_index_config = file_index_config;
    config.data_compaction_config = data_compaction_config;
    config.persistence_config = iceberg_persistence_config;
    let mooncake_table_metadata = create_test_table_metadata_with_config(
        local_table_directory,
        config,
    );

    // Local filesystem to store read-through cache.
    let cache_temp_dir = tempdir().unwrap();
    let object_storage_cache = create_test_object_storage_cache(&cache_temp_dir);

    let (mut table, mut notify_rx) = create_mooncake_table_and_notify(
        mooncake_table_metadata.clone(),
        iceberg_table_config.clone(),
        object_storage_cache.clone(),
    )
    .await;
    let row = test_row_1();

    // disk file 1: 
    // one row, in mooncake and iceberg
    // deleted, in batch deletion vector and puffin
    //
    // disk file 2: 
    // one row, in mooncake and iceberg
    // deleted, in batch deletion vector but not puffin
    //
    // disk file 3: 
    // one row, not in iceberg
    // no deletion
    //
    // Initial append.
    table.append(row.clone()).unwrap();
[...]

PR #1795, opened 2025-08-29
[storage] Fix deletion record remapping

dentiny:
## Summary

A committed deletion record could appear in two places: in-memory committed deletion logs at snapshot, or persisted deletion vector.
After compaction we need to do remapping on both.

## Related Issues

Closes https://github.com/Mooncake-Labs/moonlink/issues/1793

## Checklist

- [ ] Code builds correctly

### DIFF
--- a/src/moonlink/src/storage/iceberg/iceberg_table_syncer.rs
+++ b/src/moonlink/src/storage/iceberg/iceberg_table_syncer.rs
@@ -1,3 +1,4 @@
+use crate::storage::compaction::table_compaction::RemappedRecordLocation;
 use crate::storage::iceberg::deletion_vector::DeletionVector;
 use crate::storage::iceberg::deletion_vector::{
     DELETION_VECTOR_CADINALITY, DELETION_VECTOR_REFERENCED_DATA_FILE,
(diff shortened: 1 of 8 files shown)


src/moonlink/src/storage/iceberg/iceberg_table_syncer.rs
```diff
@@ -460,79 +504,6 @@ impl IcebergTableManager {
         Ok(remote_file_indices)
     }
 
-    /// Get current manifest entries, used for sanity check purpose.
-    async fn get_expected_entry_count_after_sync(
```
dentiny, line 464: Before considering deletion log remapping, getting the expected manifest entry count is very straightforward;
but calculating the puffin blob count for remapping is essentially the same as rewriting remapping logic, don't think worth it.

zhousun:
am I understanding it correctly, this is the order:
1. iceberg compaction
2. iceberg persistence 
3. iceberg persistence finish, deletion log truncated
4. compaction finishes

dentiny:
> am I understanding it correctly, this is the order:
> 
> 1. iceberg compaction
> 2. iceberg persistence
> 3. iceberg persistence finish, deletion log truncated
> 4. compaction finishes

yes

Issue #1448, opened 2025-08-14
chaos: create unit test for replay program


dentiny:
## Summary

The plan is to manually create a serialized and persisted event records, load and replay, and verify everything works properly.

dentiny:
replay function here: https://github.com/Mooncake-Labs/moonlink/blob/main/src/moonlink/src/storage/mooncake_table/replay/replay_events.rs

peterxcli:
I would like to work on this ~

dentiny:
> I would like to work on this ~

Thanks a ton!

dentiny:
BTW, as the initial version, a very very simple json should be good enough.
For example, you could generate a short json with event count 30 with any chaos test.

PR #1796, opened 2025-08-29
[chaos] Add regression test

dentiny:
## Summary

Add chaos test and replay event based regression test, also add to CI.

Closes https://github.com/Mooncake-Labs/moonlink/issues/1448

## Checklist

- [x] Code builds correctly


src/moonlink/src/table_handler/regression.rs
```diff
@@ -0,0 +1,3 @@
+#[cfg(test)]
+#[cfg(feature = "chaos-test")]
+mod issue_1793;
```
dentiny, line 3: Likely I will rename the mod later, not a big issue for now.

PR #1801, opened 2025-08-29
[datafusion] Connection pooling 

### LINKED ISSUE #1374: [datafusion] Connection pooling
https://github.com/Mooncake-Labs/moonlink/pull/1367 creates a connection to moonlink_service for every table in every query. But if a query uses the same table multiple times, they will share the same connection and table metadata, which is good that the read is consistent within each query

CheyuWu:
## Summary

- Create a connection pool in `src/moonlink_datafusion/src/connection_pool.rs`
- Add a test to test the connection pool

## Related Issues

Closes #1374 

## Changes

- Create `src/moonlink_datafusion/src/connection_pool.rs`
- Use connection pool in `src/moonlink_datafusion/src/table_provider.rs`

## Checklist

- [ ] Code builds correctly

---

> [!NOTE]
> Introduce a global UnixStream connection pool and use it in the table provider; add tests and wiring.
> 
> - **Core**:
>   - Implement global connection pool in `src/moonlink_datafusion/src/connection_pool.rs` with `Pool` and `PooledStream` (idle timeout, max entries per-URI, reuse on drop).
> - **Integration**:
>   - Replace direct `UnixStream` usage in `src/moonlink_datafusion/src/table_provider.rs` with pooled connections (`Pool::get_stream`, pass `PooledStream`, update drop to return to pool).
>   - Expose module in `src/moonlink_datafusion/src/lib.rs`.
> - **Tests**:
>   - Add async tests for basic reuse, multi-URI concurrency, and max-capacity behavior in `connection_pool.rs`.
> - **Deps**:
>   - Add `tempfile` as `dev-dependencies` in `src/moonlink_datafusion/Cargo.toml`.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit e0c543ac7c64d8aebc06c7c294364b135af212d1. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/src/moonlink_datafusion/src/catalog_provider.rs
+++ b/src/moonlink_datafusion/src/catalog_provider.rs
@@ -1,9 +1,9 @@
+use crate::connection_pool::Pool;
 use crate::error::Result;
 use crate::schema_provider::MooncakeSchemaProvider;
 use datafusion::catalog::{CatalogProvider, SchemaProvider};
 use std::any::Any;
 use std::sync::Arc;
-use tokio::net::UnixStream;
 
 #[derive(Debug)]
 pub struct MooncakeCatalogProvider {
@@ -12,7 +12,8 @@ pub struct MooncakeCatalogProvider {
 
 impl MooncakeCatalogProvider {
     pub async fn try_new(uri: String) -> Result<Self> {
-        UnixStream::connect(&uri).await?;
+        let _ = Pool::get_stream(&uri).await?;
+
         Ok(Self { uri })
     }
 }


src/moonlink_datafusion/src/connection_pool.rs
```diff
@@ -0,0 +1,44 @@
+use std::sync::{LazyLock, Mutex};
+
+use tokio::net::UnixStream;
+
+static CONN_POOL: LazyLock<Mutex<Vec<UnixStream>>> = LazyLock::new(|| Mutex::new(Vec::new()));
+
+pub(crate) async fn get_stream(uri: &str) -> std::io::Result<UnixStream> {
+    if let Some(stream) = CONN_POOL.lock().unwrap().pop() {
```
dpxcc: You are putting connections of different uris in the same pool, so there is no guarantee that you will get a correct connection later

dpxcc (changes requested):

src/moonlink_datafusion/src/connection_pool.rs
```diff
@@ -0,0 +1,44 @@
+use std::sync::{LazyLock, Mutex};
+
+use tokio::net::UnixStream;
+
+static CONN_POOL: LazyLock<Mutex<Vec<UnixStream>>> = LazyLock::new(|| Mutex::new(Vec::new()));
+
+pub(crate) async fn get_stream(uri: &str) -> std::io::Result<UnixStream> {
```
dentiny: Should we use `crate::Result` here? otherwise we loss error source location

src/moonlink_datafusion/src/table_provider.rs
```diff
@@ -39,7 +40,7 @@ pub struct MooncakeTableProvider {
 
 impl MooncakeTableProvider {
     pub async fn try_new(uri: &str, schema: String, table: String, lsn: u64) -> Result<Self> {
-        let mut stream = UnixStream::connect(uri).await?;
+        let mut stream = get_stream(uri).await?;
```
dpxcc: This works. But for a query joining many tables, each table will get one connection. I think we can improve this so that we use one connection per query per catalog

src/moonlink_datafusion/Cargo.toml
```diff
@@ -18,5 +18,6 @@ moonlink_table_metadata = { path = "../moonlink_table_metadata" }
 object_store = { workspace = true }
 parquet = { workspace = true }
 roaring = { workspace = true }
+tempfile = { workspace = true }
```
dpxcc: This is test only

src/moonlink_datafusion/src/connection_pool.rs
```diff
@@ -0,0 +1,44 @@
+use std::sync::{LazyLock, Mutex};
+
```
dpxcc, line 7: nit: remove new line

src/moonlink_datafusion/src/connection_pool.rs
```diff
@@ -0,0 +1,44 @@
...
+use tokio::net::UnixStream;
+
+static CONN_POOL: LazyLock<Mutex<Vec<UnixStream>>> = LazyLock::new(|| Mutex::new(Vec::new()));
```
dpxcc: nit: let's just name it POOL

src/moonlink_datafusion/src/table_provider.rs
```diff
@@ -202,7 +203,7 @@ impl ParquetFileReaderFactory for MooncakeParquetFileReaderFactory {
 }
 
 #[derive(Debug)]
-struct MooncakeTableScan {
+pub(crate) struct MooncakeTableScan {
```
dpxcc: no need to `pub(crate)`

CheyuWu:
Hi @dentiny @dpxcc 
- I have moved [tempfile](https://github.com/Mooncake-Labs/moonlink/pull/1801/files#diff-2b71c503e045161b9f985586a701c893abdd08f617674d56475f9da522565ae7R23-R25) to dev-dependency
- Address the nit problem
- Create a [struct](https://github.com/Mooncake-Labs/moonlink/pull/1801/files#diff-cc4ac496ce89ed13b4dbf693e301b4178be09673c87291d08420147606a0940fR5-R14) to store uri
- Use HashMap to address the different URIs in [here](https://github.com/Mooncake-Labs/moonlink/pull/1801/files#diff-cc4ac496ce89ed13b4dbf693e301b4178be09673c87291d08420147606a0940fR16-R17)
- Add concurrent stream test in [here](https://github.com/Mooncake-Labs/moonlink/pull/1801/files#diff-cc4ac496ce89ed13b4dbf693e301b4178be09673c87291d08420147606a0940fR68-R119)

PTAL

src/moonlink_datafusion/src/connection_pool.rs
```diff
@@ -0,0 +1,120 @@
...
+
+pub(crate) async fn get_stream(uri: &str) -> crate::Result<PooledStream> {
+    let mut pool = POOL.lock().await;
+    if let Some(mutex_vec) = pool.get_mut(uri) {
+        let mut vec = mutex_vec.lock().await;
+        if let Some(stream) = vec.pop() {
+            return Ok(PooledStream::new(uri.to_string(), stream));
+        }
+    }
+    let stream = UnixStream::connect(uri).await?;
```
dentiny: Please don't put IO operation inside of critical section

src/moonlink_datafusion/src/connection_pool.rs
```diff
@@ -0,0 +1,120 @@
...
+impl PooledStream {
+    pub fn new(uri: String, stream: UnixStream) -> Self {
+        PooledStream { uri, stream }
+    }
+}
+
+static POOL: LazyLock<Mutex<HashMap<String, Mutex<Vec<UnixStream>>>>> =
+    LazyLock::new(|| Mutex::new(HashMap::new()));
+
+pub(crate) async fn get_stream(uri: &str) -> crate::Result<PooledStream> {
```
dentiny: Do you think it possible to provide a RAII wrapper, which gets or create a connection on construction, and returns back on destruction? That way we never forget return.

src/moonlink_datafusion/src/connection_pool.rs
dentiny: Some other considerations on connection pool:
- Do you want to set timeout to connection entries? If you do, do we need background maintainance thread?
- Do you want to set a max cap for connection pool? I think it's always good to have resource capped, for here it's fd / port number and memory.

src/moonlink_datafusion/src/table_provider.rs
```diff
@@ -39,10 +39,11 @@ pub struct MooncakeTableProvider {
 
 impl MooncakeTableProvider {
     pub async fn try_new(uri: &str, schema: String, table: String, lsn: u64) -> Result<Self> {
-        let mut stream = UnixStream::connect(uri).await?;
-        let table_schema = get_table_schema(&mut stream, schema.clone(), table.clone()).await?;
+        let mut pooled_stream = get_stream(uri).await?;
```
dentiny: If the returned cached connection is broken, how do we handle retry?

src/moonlink_datafusion/src/connection_pool.rs
```diff
@@ -0,0 +1,120 @@
+use std::{collections::HashMap, sync::LazyLock};
+use tokio::{net::UnixStream, sync::Mutex};
+
+#[derive(Debug)]
+pub struct PooledStream {
+    pub uri: String,
```
dentiny: Not sure why do we need uri here?

src/moonlink_datafusion/src/table_provider.rs
```diff
@@ -230,14 +237,15 @@ impl MooncakeTableScan {
...
     fn drop(&mut self) {
-        let stream = self.stream.take();
+        let pooled_stream = self.pooled_stream.take();
         let schema = std::mem::take(&mut self.schema);
         let table = std::mem::take(&mut self.table);
         tokio::spawn(async move {
-            let mut stream = stream.expect("stream should be set by try_new");
-            if let Err(e) = scan_table_end(&mut stream, schema, table).await {
+            let mut pooled_stream = pooled_stream.expect("stream should be set by try_new");
+            if let Err(e) = scan_table_end(&mut pooled_stream.stream, schema, table).await {
```
dentiny: If the rpc invocation fails, do we want to return connection back to pool still?

src/moonlink_datafusion/src/connection_pool.rs
CheyuWu: OK, no problem

src/moonlink_datafusion/src/connection_pool.rs
CheyuWu: I think I can do this, let me try it

src/moonlink_datafusion/src/connection_pool.rs
CheyuWu: I wanted to use the stream to search the HashMap when returning it.

src/moonlink_datafusion/src/table_provider.rs
CheyuWu: I think we can just drop it

src/moonlink_datafusion/src/table_provider.rs
CheyuWu: if the connection is broken, we retry twice

src/moonlink_datafusion/src/table_provider.rs
CheyuWu: We can always return the connection to the pool, regardless of whether the RPC invocation succeeds or fails.

And check the connection status in the `get_stream` function

src/moonlink_datafusion/src/connection_pool.rs
CheyuWu: Thanks for the suggestions!
- For timeout, I will add a background maintainance thread
- For the connection pool, I will set a maximum cap per URI (10 -50  pooled connections per URI)

CheyuWu:
Hi @dentiny PTAL
In the latest PR, I have
- Introduced a RAII wrapper for PooledStream, so we no longer need to explicitly call return_stream at the end of the lifecycle.
- Moved I/O operations out of critical sections to avoid blocking the pool lock.
- Added a background maintenance thread to handle idle timeouts.
- Enforced a maximum capacity for the connection pool.
- Improved connection health check in get_stream by removing broken connections.
- Added a retry mechanism (up to two attempts) in MooncakeTableProvider::try_new to increase robustness.

src/moonlink_datafusion/src/connection_pool.rs
```diff
@@ -0,0 +1,361 @@
...
+    time::{Duration, Instant},
+};
+use tokio::{net::UnixStream, sync::Mutex, time::sleep};
+
+struct PooledEntry {
+    stream: UnixStream,
+    last_used: Instant,
+}
+
+static POOL: LazyLock<Mutex<HashMap<String, Mutex<Vec<PooledEntry>>>>> =
```
dentiny: nit: add a documentation, key is URI, value is a pool of entries.

src/moonlink_datafusion/src/connection_pool.rs
```diff
@@ -0,0 +1,361 @@
...
+    LazyLock::new(|| Mutex::new(HashMap::new()));
+
+const MAX_PER_URI: usize = 30;
+
+static IDLE_TIMEOUT_MS: AtomicU64 = AtomicU64::new(30_000);
```
dentiny: I only see it updated in testing, which we should inject via connection pool configs, instead of global variables.

src/moonlink_datafusion/src/connection_pool.rs
```diff
@@ -0,0 +1,361 @@
...
+
+impl Drop for PooledStream {
+    fn drop(&mut self) {
+        if let Some(stream) = self.stream.take() {
+            let uri = self.uri.clone();
+            tokio::spawn(async move {
+                let mut pool = POOL.lock().await;
+                let mut pool_vec = pool.entry(uri).or_default().lock().await;
+
+                if pool_vec.len() < MAX_PER_URI {
```
dentiny: so every single time you push to the end and pop from the end of the vector;
on insufficient connection stream space, should we pop from the front? since it's least recently accessed, otherwise it's easy to get them timed-out anyway

src/moonlink_datafusion/src/connection_pool.rs
```diff
@@ -0,0 +1,361 @@
...
+const MAX_PER_URI: usize = 30;
+
+static IDLE_TIMEOUT_MS: AtomicU64 = AtomicU64::new(30_000);
+
+pub fn idle_timeout() -> Duration {
+    Duration::from_millis(IDLE_TIMEOUT_MS.load(Ordering::Relaxed))
+}
+
+#[derive(Debug)]
+pub struct PooledStream {
```
dentiny: curious why do you call it `PooledStream`?

src/moonlink_datafusion/src/connection_pool.rs
dentiny: we already have a global lock for `POOL`, curious what's the purpose for mutex inside of each entry?

src/moonlink_datafusion/src/connection_pool.rs
```diff
@@ -0,0 +1,361 @@
...
+    pub uri: String,
+    pub stream: Option<UnixStream>,
```
dentiny, line 97: nit: add a comment that option is made for ownership transfer at drop

src/moonlink_datafusion/src/connection_pool.rs
```diff
@@ -0,0 +1,361 @@
...
+        atomic::{AtomicU64, Ordering},
+        LazyLock,
+    },
+    time::{Duration, Instant},
+};
+use tokio::{net::UnixStream, sync::Mutex, time::sleep};
+
+struct PooledEntry {
+    stream: UnixStream,
+    last_used: Instant,
```
dentiny: Incorrect variable naming I think, it's the timestamp where the entry is inserted into the pool, not the timestamp it was last used.

src/moonlink_datafusion/src/connection_pool.rs
```diff
@@ -0,0 +1,361 @@
...
+pub(crate) async fn get_stream(uri: &str) -> crate::Result<PooledStream> {
+    {
+        let mut pool = POOL.lock().await;
+        if let Some(mutex_vec) = pool.get_mut(uri) {
+            let mut vec = mutex_vec.lock().await;
+            // Remove expired streams
+            vec.retain(|entry| entry.last_used.elapsed() <= idle_timeout());
+
+            while let Some(entry) = vec.pop() {
+                if is_connection_alive(&entry.stream).await {
```
dentiny: please don't put IO operations inside of critical section
for implementation easiness, we could return the stream directly

src/moonlink_datafusion/src/connection_pool.rs
```diff
@@ -0,0 +1,361 @@
...
+    }
+}
+
+pub(crate) async fn get_stream(uri: &str) -> crate::Result<PooledStream> {
+    {
+        let mut pool = POOL.lock().await;
+        if let Some(mutex_vec) = pool.get_mut(uri) {
+            let mut vec = mutex_vec.lock().await;
+            // Remove expired streams
+            vec.retain(|entry| entry.last_used.elapsed() <= idle_timeout());
```
dentiny: let's use deque for perf

src/moonlink_datafusion/src/connection_pool.rs
```diff
@@ -0,0 +1,361 @@
...
+        Err(_) => false,
+    }
+}
+
+pub async fn start_maintenance_task() {
+    loop {
+        sleep(Duration::from_secs(5)).await;
+
+        let mut pool = POOL.lock().await;
+        for (_uri, vec_mutex) in pool.iter_mut() {
```
dentiny: nit
```suggestion
        for (_, vec_mutex) in pool.iter_mut() {
```

src/moonlink_datafusion/src/connection_pool.rs
```diff
@@ -0,0 +1,361 @@
...
+
+async fn is_connection_alive(stream: &UnixStream) -> bool {
+    match stream.try_read(&mut []) {
+        Ok(_) => true,
+        Err(e) if e.kind() == std::io::ErrorKind::WouldBlock => true,
+        Err(_) => false,
+    }
+}
+
+pub async fn start_maintenance_task() {
```
dentiny: I'm fine with a background maintenance task, but can we scope it to pool?

src/moonlink_datafusion/src/connection_pool.rs
dentiny: Sorry for the confusion, I'm actually thinking about sth else on maintenance thread.

So for now, we only cleanup expire entries in the background thread (in an eager fashion), which is fine. But lazy mode (only evict on access) is also ok.

I'm actually thinking about reconnection, which should be part of the config;
it's super useful when you have large number of connection expiring at similarly the same time.
I won't treat it as a merge blocker in this PR, so for the simpliest impl we just remove the thread.

Hopefully it makes sense to you, happy to chat offline. :)

src/moonlink_datafusion/src/connection_pool.rs
CheyuWu: Great idea, I like that, thx

src/moonlink_datafusion/src/connection_pool.rs
CheyuWu: Well, this is named by ChatGPT. I am poor at naming the function or variable.
If there is a better name, I am willing to change it

src/moonlink_datafusion/src/connection_pool.rs
CheyuWu: My idea is to first acquire the outer lock, then get the array corresponding to the URI. At this point, we can release the outer lock, and afterwards lock only the inner array. This way, we don’t need to lock the entire HashMap. However, my current implementation seems to have some issues, so I’ll work on fixing it.

src/moonlink_datafusion/src/connection_pool.rs
CheyuWu: I am sorry about the same mistake. This won't happen again

src/moonlink_datafusion/src/connection_pool.rs
CheyuWu: Got it, thanks for clarifying! I’ll go with the simpler approach and remove the background thread for now. 
We can discuss reconnection and more advanced maintenance later if needed.

dentiny:
ping me when ready, thanks!

CheyuWu:
Hi @dentiny, I have addressed the problem that you have mentioned, PTAL.

src/moonlink_datafusion/src/connection_pool.rs
```diff
@@ -0,0 +1,406 @@
...
+    collections::{HashMap, VecDeque},
+    sync::{
+        atomic::{AtomicU64, Ordering},
+        Arc, LazyLock,
+    },
+    time::{Duration, Instant},
+};
+use tokio::{net::UnixStream, sync::Mutex, time::sleep};
+
+pub static POOL: LazyLock<Arc<Pool>> = LazyLock::new(|| Arc::new(Pool::new(30, 30_000)));
```
dentiny: make these two params constant.
also we should be able to create a constant duration value?

src/moonlink_datafusion/src/connection_pool.rs
```diff
@@ -0,0 +1,406 @@
...
+
+#[derive(Debug)]
+pub struct PooledEntry {
+    stream: UnixStream,
+    inserted_at: Instant,
+}
+
+#[derive(Debug)]
+pub struct Pool {
+    /// Global connection pool.
```
dentiny: nit: this is struct-wise comment?

src/moonlink_datafusion/src/connection_pool.rs
dentiny: Your implementation works on the assumption that emplacing an entry is expensive, which I don't really think so; one global lock should be good enough.

src/moonlink_datafusion/src/connection_pool.rs
```diff
@@ -0,0 +1,406 @@
...
+    ///
+    /// Key: URI (String)
+    /// Value: Mutex-protected VecDeque of PooledEntry (the pool for that URI)
+    pub inner: Mutex<HashMap<String, Arc<Mutex<VecDeque<PooledEntry>>>>>,
+    pub max_per_uri: usize,
+    pub idle_timeout_ms: AtomicU64,
```
dentiny: Why do we use atomic variable here? It shouldn't be changed during runtime.

dentiny: My point is, timeout and max per uri should both be configs, which keeps constant at runtime.

src/moonlink_datafusion/src/connection_pool.rs
```diff
@@ -0,0 +1,406 @@
...
+}
+
+#[derive(Debug)]
+pub struct Pool {
+    /// Global connection pool.
+    ///
+    /// Key: URI (String)
+    /// Value: Mutex-protected VecDeque of PooledEntry (the pool for that URI)
+    pub inner: Mutex<HashMap<String, Arc<Mutex<VecDeque<PooledEntry>>>>>,
+    pub max_per_uri: usize,
```
dentiny: nit
```suggestion
    pub max_entries_per_uri: usize,
```

src/moonlink_datafusion/src/connection_pool.rs
```diff
@@ -0,0 +1,406 @@
...
+    pub maintenance_interval: Duration,
+}
+
+impl Pool {
+    pub fn new(max_per_uri: usize, idle_timeout_ms: u64) -> Self {
+        Self {
+            inner: Mutex::new(HashMap::new()),
+            max_per_uri,
+            idle_timeout_ms: AtomicU64::new(idle_timeout_ms),
+            maintenance_interval: Duration::from_secs(5),
```
dentiny: let's make the duration a constant.

src/moonlink_datafusion/src/lib.rs
```diff
@@ -1,8 +1,10 @@
 mod catalog_provider;
+mod connection_pool;
 mod error;
 mod schema_provider;
 mod table_provider;
 
 pub use catalog_provider::MooncakeCatalogProvider;
+pub use connection_pool::{start_global_maintenance_pool_task, POOL};
```
dentiny: Why do we export it?

src/moonlink_datafusion/src/connection_pool.rs
CheyuWu: Ok no problem

src/moonlink_datafusion/src/lib.rs
CheyuWu: Sorry for the confusion earlier—I misunderstood Rust’s export rules.
You’re right, we don’t need to export this function since it’s only used internally.
I’ve updated the code to keep it internal. Thanks for pointing it out!

src/moonlink_datafusion/src/connection_pool.rs
CheyuWu: OK no problem

CheyuWu:
Hi @dentiny, I have addressed the issues PTAL
- Change param to constant
- Move struct-wise comment
- Change to global lock in connection pool
- Remove atomic variable
- Fix the wrong export
- Remove retry and background clean idle timeout thread

src/moonlink_datafusion/src/connection_pool.rs
```diff
@@ -0,0 +1,276 @@
+use std::{
+    collections::{HashMap, VecDeque},
+    sync::{Arc, LazyLock},
+    time::{Duration, Instant},
+};
+use tokio::{net::UnixStream, sync::Mutex};
+
+const DEFAULT_MAX_ENTRIES_PER_URI: usize = 30;
+const DEFAULT_IDLE_TIMEOUT_MS: u64 = 30_000;
```
dentiny: nit: can we make it a duration constant?

src/moonlink_datafusion/src/connection_pool.rs
```diff
@@ -0,0 +1,276 @@
...
+
+pub(crate) static POOL: LazyLock<Arc<Pool>> = LazyLock::new(|| {
```
dentiny: All connection pool related variables, classes and functions seems to be scoped in the file, which means they could be `private`?

src/moonlink_datafusion/src/connection_pool.rs
```diff
@@ -0,0 +1,276 @@
...
+pub(crate) async fn get_stream_with_pool(
+    uri: &str,
+    pool: &Arc<Pool>,
+) -> crate::Result<PooledStream> {
+    {
+        let mut pool_inner = pool.inner.lock().await;
+
+        if let Some(vec) = pool_inner.get_mut(uri) {
+            // Remove expired streams
+            vec.retain(|entry| entry.inserted_at.elapsed() <= pool.idle_timeout());
```
dentiny: The entry is sorted in insertion time, so we don't need to iterate through all items.
could do sth like
```rs
// get the first item which is larger than timeout cutoff
// remove all items before
```

src/moonlink_datafusion/src/connection_pool.rs
dentiny: no pressure :)

src/moonlink_datafusion/src/connection_pool.rs
```diff
@@ -0,0 +1,276 @@
...
+                pool_vec.push_back(PooledEntry {
+                    stream,
+                    inserted_at: Instant::now(),
+                });
+            });
+        }
+    }
+}
+
+pub(crate) async fn get_stream_with_pool(
```
dentiny: (no request to change, just a question)
Do you think it better to make it a member function for `Pool`?

CheyuWu: You are right !! 
Making it a member function of `Pool` would make the API more idiomatic and easier to use.
Should I refactor it into a method in this PR or in the following PR?

src/moonlink_datafusion/src/connection_pool.rs
CheyuWu: Should we change Vecque and use binary search to remove the timeout stream? Or is it better to use linear to delete the timeout stream?

CheyuWu:
Hi @dentiny, I have addressed the problem PTAL
- Make `DEFAULT_IDLE_TIMEOUT_MS` a duration constant
- All connection pool-related variables, classes, and functions seem to be scoped in the file
- optimize the removal of  `idle_timeout` stream
> Use linear to delete the timeout stream, if it s better to change to binary search, plz let me know

src/moonlink_datafusion/src/connection_pool.rs
dentiny: Let's clean it up in this pr, thanks!

CheyuWu:
Hi @dentiny, I have moved the get stream function into Pool PTAL

src/moonlink_datafusion/src/connection_pool.rs
```diff
@@ -0,0 +1,280 @@
...
+                let mut pool_inner = pool.inner.lock().await;
+                let pool_vec = pool_inner.entry(uri).or_default();
+
+                while pool_vec.len() >= pool.max_entries_per_uri {
+                    pool_vec.pop_front();
+                }
+
+                pool_vec.push_back(PooledEntry {
+                    stream,
+                    inserted_at: Instant::now(),
```
dentiny, line 132: I'm not 100% sure if it's good to always restart a new insertion timestamp, since the connection itself has lasted long might have timeout.
But no strong opinion here.

src/moonlink_datafusion/src/connection_pool.rs
```diff
@@ -0,0 +1,280 @@
...
+
+        let addr1 = stream1.stream_mut().peer_addr().unwrap();
+        let addr2 = stream2.stream_mut().peer_addr().unwrap();
+
+        drop(stream1);
+        drop(stream2);
+
+        // Retrieve streams for both URIs again; should reuse connections from the pool
+        let (stream1b, stream2b) = tokio::join!(
+            test_pool.get_stream_with_pool(uri1),
```
dentiny, line 227: Since we want to check returned stream is the recycled one, do you think it's better to use an invalid IP/port so it won't even connect?

src/moonlink_datafusion/src/connection_pool.rs
```diff
@@ -0,0 +1,280 @@
...
+            .await
+            .expect("should connect");
+        drop(stream1);
+
+        let mut stream2 = test_pool
+            .get_stream_with_pool(&uri_str)
+            .await
+            .expect("should reuse from pool");
+        let unix_stream = stream2.stream.take().unwrap();
+        assert!(unix_stream.into_std().is_ok());
```
dentiny:
```suggestion
        let _ = unix_stream.into_std().unwrap();
```
so we know the error

src/moonlink_datafusion/src/connection_pool.rs
CheyuWu: We are using Unix domain sockets, so there is no IP/port. If we intentionally use an invalid path, the first connect fails, and we cannot obtain an initial stream to later verify reuse.
 
 I switched the test to capture the original stream’s raw file descriptor and assert the descriptor is identical when the stream is fetched again. If the pool reuses the connection, the fd matches; if it creates a new one, it differs.

src/moonlink_datafusion/src/connection_pool.rs
CheyuWu: Maybe we can have a follow-up PR to introduce a background thread that maintains the timeout?

CheyuWu:
Hi @dentiny, I have addressed the issue. PTAL

src/moonlink_datafusion/src/connection_pool.rs
dentiny: > If we intentionally use an invalid path, the first connect fails, and we cannot obtain an initial stream to later verify reuse.

I only asked to use invalid URI here, not everywhere.

CheyuWu: Sorry, I misunderstood your point earlier. Thanks for clarifying.

CheyuWu:
Hi @dentiny, I have addressed the problem you have mentioned in https://github.com/Mooncake-Labs/moonlink/pull/1801#discussion_r2383734157 PTAL 

I removed the file descriptor on my side to produce an invalid URI.
I’m not entirely sure if this matches what you had in mind.

src/moonlink_datafusion/src/connection_pool.rs
CheyuWu: I removed the file descriptor on my side to produce an invalid URI.
I’m not entirely sure if this matches what you had in mind.

dpxcc:
I’m more inclined toward catalog-level connection pooling rather than a global one, but I’m fine with the current implementation.
Let's also update https://github.com/Mooncake-Labs/moonlink/blob/9e4ae4f83bf0dff2adff697d48de2af1286bd631/src/moonlink_datafusion/src/catalog_provider.rs#L15 to use the connection pool

dentiny:
@CheyuWu ping me if you want me to pick it up, thanks.

CheyuWu:
> @CheyuWu ping me if you want me to pick it up, thanks.
Hi @dentiny Thanks for offering! I’ll take this one, but I’ll reach out if I need any help.

CheyuWu:
Hi @dpxcc, I have used the connection pool in `catalog_provider.rs` PTAL

dpxcc:
lgtm

Issue #1802, opened 2025-08-29
[BUG] Snapshot collision on initial copy finish


nbiscaro:
## Description

```
thread 'tokio-runtime-worker' panicked at moonlink/src/moonlink/src/storage/mooncake_table.rs:990:9:
assertion failed: !self.mooncake_snapshot_ongoing
stack backtrace:
```

## Steps to Reproduce

1. 
2. 
3. 

## Version

- moonlink version (if applicable):
- pg_mooncake version (if applicable):

## Expected Behavior

What did you expect to happen?

## Actual Behavior

What actually happened?

## Additional Context

Any other information (logs, screenshots, configs) that might help.

PR #1803, opened 2025-08-29
Coordinate initial copy finish with snapshot

nbiscaro:
## Summary

When initial copy finishes we create a snapshot unconditionally. We should first check if there is an ongoing snapshot and only then issue the snapshot and set the flag. If there is a snapshot currently ongoing, we skip and let the next scheduled snapshot pick up the data from initial copy. 

## Related Issues

Closes https://github.com/Mooncake-Labs/moonlink/issues/1802

## Changes

- Guard against ongoing snapshot
- Add regression test
- 

## Checklist

- [ ] Code builds correctly

### DIFF
--- a/src/moonlink/src/table_handler/tests.rs
+++ b/src/moonlink/src/table_handler/tests.rs
@@ -2460,3 +2460,56 @@ async fn test_initial_copy_iceberg_ack_without_wal() {
 
     env.shutdown().await;
 }
+
+#[tokio::test]
+async fn test_finish_initial_copy_overlaps_with_ongoing_snapshot() {
+    // Setup environment
+    let mut env = TestEnvironment::default().await;
+    let sender = env.handler.get_event_sender();
+
+    // Start initial copy mode
+    sender
+        .send(TableEvent::StartInitialCopy)
+        .await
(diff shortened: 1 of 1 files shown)


src/moonlink/src/table_handler.rs
```diff
@@ -349,17 +349,18 @@ impl TableHandler {
...
-                            dump_snapshot: false,
-                            iceberg_snapshot_option: IcebergSnapshotOption::BestEffort(
-                                uuid::Uuid::new_v4()
-                            ),
-                            index_merge_option: MaintenanceOption::Skip,
-                            data_compaction_option: MaintenanceOption::Skip,
-                        }));
-                        table_handler_state.mooncake_snapshot_ongoing = true;
+                        if !table_handler_state.mooncake_snapshot_ongoing {
+                            let _ = table.create_snapshot(SnapshotOption {
```
dentiny: curious why ignore the return value?

nbiscaro: oops added back

dentiny:
I will take a look at the failed chaos test.

Issue #1818, opened 2025-08-30
BUG: CI connectivity failure should fail the pipeline


dentiny:
## Summary

https://github.com/Mooncake-Labs/moonlink/blob/main/.github/workflows/ci.yml
currently it's only issuing warnings until timeout

PR #1789, opened 2025-08-29
[REST-API]: expose endpoints to create snapshot

Standing-Man:
## Summary

expose endpoints to create snapshot

## Related Issues

- Closes Part of #1646.

## Changes

- 
- 
- 

## Checklist

- [x] Code builds correctly
- [x] Tests have been added or updated
- [ ] Documentation updated if necessary
- [x] I have reviewed my own changes

PR #1828, opened 2025-08-31
[CI]: fix the Moonlink Service Integration Tests

Standing-Man:
## Summary

fix the Moonlink Service Integration Tests on **CI**

## Related Issues

- Closes Part of #1818.

## Changes

- 
- 
- 

## Checklist

- [x] Code builds correctly

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -662,8 +662,6 @@ jobs:
             sleep 2
           done
           echo "ERROR: nginx failed to start."
-          docker ps -a
-          docker logs nginx || true
           exit 1
 
       - name: Run moonlink_service integration tests
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -631,6 +631,7 @@ jobs:
         run: |
           mkdir -p ${{ github.workspace }}/.shared-nginx
           chmod 777 ${{ github.workspace }}/.shared-nginx
+          echo "nginx ok" > ${{ github.workspace }}/.shared-nginx/index.html
 
       # ─────────────────────── Start nginx manually ──────────────────────────
       - name: Start nginx
@@ -651,14 +652,19 @@ jobs:
       - name: Wait for nginx
         timeout-minutes: 2
         run: |
+          set -euo pipefail
           for i in {1..10}; do
             if curl -sf http://localhost:80; then
               echo "Nginx is ready!"
-              break
+              exit 0
             fi
             echo "Waiting for nginx..."
             sleep 2
           done
+          echo "ERROR: nginx failed to start."
+          docker ps -a
+          docker logs nginx || true
+          exit 1
 
       - name: Run moonlink_service integration tests
         timeout-minutes: 10
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -664,5 +664,6 @@ jobs:
         timeout-minutes: 10
         env:
           MOONLINK_BACKEND_DIR: ${{ github.workspace }}/.shared-nginx
+          NGINX_ADDR: http://nginx.local:80
         run: |
           RUST_BACKTRACE=1 cargo test --package moonlink_service --lib --features standalone-test -- --nocapture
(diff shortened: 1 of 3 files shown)


Standing-Man:
~~:construction: I’m still trying to resolve the CI issues. :joy:~~
`NGINX_ADDR` is currently hard-coded. Should we try reading it from the environment later?

dentiny:
> NGINX_ADDR is currently hard-coded. Should we try reading it from the environment later?

Sure

Standing-Man:
Hi @dentiny, I found that the nginx server startup issue was due to the mounted `.shared-nginx` directory being empty, while the nginx server expects an `index.html`. My current workaround is to create an `index.html`.

If we want the `.shared-nginx` directory to remain empty, I’ll look for another way to fix this issue.

~~:construction: Please don’t merge this PR yet, as it still contains some debugging information and typos in #1789.~~ Thanks.

.github/workflows/ci.yml
```diff
@@ -631,6 +631,7 @@ jobs:
         run: |
           mkdir -p ${{ github.workspace }}/.shared-nginx
           chmod 777 ${{ github.workspace }}/.shared-nginx
+          echo "nginx ok" > ${{ github.workspace }}/.shared-nginx/index.html
```
Standing-Man, line 634: Currently, there is an `index.html` stored in the `.shared-nginx` directory. Perhaps we could relax the condition so that `curl -sf` returning either **200** or **403** (when `.shared-nginx` is empty) is considered a success?

dentiny:
> Hi @dentiny, I found that the nginx server startup issue was due to the mounted `.shared-nginx` directory being empty, while the nginx server expects an `index.html`. My current workaround is to create an `index.html`.
> 
> If we want the `.shared-nginx` directory to remain empty, I’ll look for another way to fix this issue.

I think it's fine, I'm just curious why such requirement doesn't prevent nginx to start locally

.github/workflows/ci.yml
dentiny: I'm so stupid; yes making curl request to nginx does require to provide a URI, and index.html is the default.
I'm thinking, to check connectivity, do you think we could use `nc -vz`?

dentiny:
Ping me when ready, thank you so much for your generous and prompt help!

Standing-Man:
> I think it's fine, I'm just curious why such requirement doesn't prevent nginx to start locally

Since the local setup doesn’t use `curl` to request the nginx server, nginx starts successfully in CI, but the `.shared-nginx` directory is empty, which causes errors when accessing the nginx server. (Startup success does not equal access success.)

src/moonlink_service/src/rest_api.rs
```diff
@@ -517,8 +517,8 @@ async fn create_snapshot(
     Json(payload): Json<CreateSnapShotRequest>,
 ) -> Result<Json<CreateSnapShotResponse>, (StatusCode, Json<ErrorResponse>)> {
     debug!(
-        "Received create snapshot request for '{}': {:?}",
-        table, payload
+        "Received create snapshot request for '{}': {}",
+        &payload.database, table
```
Standing-Man, line 521: Is this error message more appropriate? If so, once the **CI** completes, this PR will be ready to merge. PTLA @dentiny, thank you.

.github/workflows/ci.yml
Standing-Man: As far as I know, `nc` is mainly employed to test whether a port is active. Do we need to look for a better way to check if the Nginx server is running properly?

dentiny: Yes, I'm thinking of `nc` because 
- CI is a constrained and isolated environment, port accessibility _should_ indicate nginx accessibility
- We don't need to prepare extra file

But I definitely agree curl nginx is more straightforward

Standing-Man: Yes, that's correct:
- `nc` can quickly check whether the Nginx process is running and the port is open. 
- `curl` is more suitable for confirming that Nginx is fully ready and able to handle requests.

Standing-Man:
Thanks for the review!

Issue #1671, opened 2025-08-22
REST: Add unit test for two table operations


dentiny:
## Summary

Now REST API supports replication LSN and commit LSN as we do in pg ingestion, we're able to test on multiple tables.
Test code should be placed under: https://github.com/Mooncake-Labs/moonlink/blob/main/src/moonlink_service/src/test.rs

francescoaristei:
Hi @dentiny, can I work on this issue? thank you

dentiny:
> Hi [@dentiny](https://github.com/dentiny), can I work on this issue? thank you

Sure thanks!

dentiny:
Hi @francescoaristei, any progress or blocker? I'm happy to help, thanks!

francescoaristei:
Hi @dentiny, I am new to moonlink so it is taking a bit, hope that’s not a problem..
If I understood correctly the goal is to test operations like upload/insert in more than 1 table as commit LSN is per table while replication LSN is per database. It would be helpful if you could explain me better what’s the end goal of the tests. Thank you!

dentiny:
> Hi [@dentiny](https://github.com/dentiny), I am new to moonlink so it is taking a bit, hope that’s not a problem.. If I understood correctly the goal is to test operations like upload/insert in more than 1 table as commit LSN is per table while replication LSN is per database. It would be helpful if you could explain me better what’s the end goal of the tests. Thank you!

No worries, ping me any time if you need help. :)

So the test linked in the issue description, if you take a look, only contains test cases which operates on the single table.
In detail, it always create a table, perform a CRUD operations upon the table, read table status, etc.

I want to add new test cases which operations on multiple tables, since we used to have related bugs in the past.
I imagine the simpliest way to add test is:
- create two tables
- insert some rows into table A
- insert some rows into table B
- scan table A and B respectively, and check their content
- mainly I want to make sure no operation gets blocked, and the content returned is correct

francescoaristei:
Oh okay, now is more clear. I got a bit confused with the distinction between commit LSN and replication LSN and I wasn’t sure what the tests were about. And thank you for the help :)

PR #1852, opened 2025-09-01
[moonlink_service][test] split testing utils into seperate file

kenchung285:
## Summary

Briefly explain what this PR does.

## Related Issues

Part of #1325 
Follow up

## Changes

- 
- 
- 

## Checklist

- [x] Code builds correctly
- [x] Tests have been added or updated
- [ ] Documentation updated if necessary
- [x] I have reviewed my own changes

PR #1853, opened 2025-09-01
Added tests for two table operations.

francescoaristei:
## Summary

Based on the tests for one table:

- Insert of json payload.
- Upload of parquet file.
- Insert of parquet file.

Extended same operations on two table for same database.

## Related Issues

Closes [#1671](https://github.com/Mooncake-Labs/moonlink/issues/1671)

## Checklist

- [x] Code builds correctly
- [x ] Tests have been added or updated

### DIFF
--- a/src/moonlink_service/src/test.rs
+++ b/src/moonlink_service/src/test.rs
@@ -706,7 +706,7 @@ async fn test_moonlink_standalone_file_insert() {
         &mut moonlink_stream,
         DATABASE.to_string(),
         TABLE.to_string(),
-        // Four events generated: one append and one commit, with commit LSN 2.
+        // Two events generated: one append and one commit, with commit LSN 2.
         /*lsn=*/
         2,
     )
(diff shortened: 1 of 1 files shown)


src/moonlink_service/src/test.rs
```diff
@@ -732,6 +732,368 @@ async fn test_moonlink_standalone_file_insert() {
...
+    .await
+    .unwrap();
+
+    // Scan table B and get data file and puffin files back.
+    let mut moonlink_stream = TcpStream::connect(MOONLINK_ADDR).await.unwrap();
+    let bytes = scan_table_begin(
+        &mut moonlink_stream,
+        DATABASE.to_string(),
+        table_b.to_string(),
+        /*lsn=*/ 1,
```
dentiny: I think LSN is an implementation detail and subject to change, so it's better to get LSN from response.
Example for reference: https://github.com/Mooncake-Labs/moonlink/blob/00ef02f13438be9a18f1200656a39d6169416692/src/moonlink_service/src/test.rs#L482

dentiny (commented):
Another high-level comment, since we have unit tests for both single table and multi-tables, is it possible to share some code and reduce code duplication?

src/moonlink_service/src/test.rs
francescoaristei: Ok, let me change it.

francescoaristei:
> Another high-level comment, since we have unit tests for both single table and multi-tables, is it possible to share some code and reduce code duplication?

Sure, let me work on it.

dentiny:
CC @kenchung285 as well

dentiny:
ping me when ready, thanks!

francescoaristei:
> ping me when ready, thanks!

@dentiny  Pushed now, thank you!

src/moonlink_service/src/test_utils.rs
```diff
@@ -82,6 +85,156 @@ pub(crate) fn create_test_arrow_schema() -> Arc<ArrowSchema> {
...
+        .post(format!("{REST_ADDR}/ingest/{table_name}"))
+        .header("content-type", "application/json")
+        .json(payload)
+        .send()
+        .await
+        .unwrap()
+}
+
+/// Test util function to send upload request.
+pub(crate) async fn send_test_upload(
```
dentiny: should we put the wait for response also in the util function?

src/moonlink_service/src/test.rs
```diff
@@ -351,6 +351,37 @@ async fn run_optimize_table_test(mode: &str) {
     .unwrap();
 }
 
+/// Util function to check data file and puffin
+async fn assert_data_and_puffin(table: &str, lsn: u64) {
```
kenchung285: We are moving these non-test functions into `test_utils.rs` to make `test.rs` with only testing scripts in the PR: https://github.com/Mooncake-Labs/moonlink/pull/1852
Would you mind move this function into `test_utils.rs`?

francescoaristei: Sure, let me do it

src/moonlink_service/src/test_utils.rs
francescoaristei: Meaning we want to block for the client to return the response?

dentiny: Yes, lets put it in util function

dentiny: Just like what we do for other test util functions

src/moonlink_service/src/test_utils.rs
```diff
@@ -512,3 +707,34 @@ pub(crate) async fn create_snapshot(
         "Response status is {response:?}"
     );
 }
+
+/// Util function to check data file and puffin
```
dentiny: would be nice if we could record how data gets generated, otherwise people get confused why there's only one data file.

src/moonlink_service/src/test_utils.rs
```diff
@@ -91,6 +95,197 @@ pub(crate) fn create_test_arrow_schema() -> Arc<ArrowSchema> {
...
+    payload: &serde_json::Value,
+) {
+    let response = client
+        .post(format!("{REST_ADDR}/tables/{table_name}"))
+        .header("content-type", "application/json")
+        .json(&payload)
+        .send()
+        .await
+        .unwrap();
+    assert!(!response.status().is_success());
```
dentiny: let's log response as well, otherwise we cannot debug

src/moonlink_service/src/test_utils.rs
francescoaristei: I added a small comment, did I understand correctly?

src/moonlink_service/src/test.rs
```diff
@@ -117,60 +117,18 @@ async fn run_optimize_table_test(mode: &str) {
...
-        .header("content-type", "application/json")
-        .json(&insert_payload)
-        .send()
-        .await
-        .unwrap();
-    assert!(
-        response.status().is_success(),
-        "Response status is {response:?}"
-    );
+    let response: IngestResponse =
```
dentiny, line 123: `send` is not a good naming, since you block wait its completion also

francescoaristei: Makes sense.

Issue #1864, opened 2025-09-02
[ISSUE] Propagate directory creation error


nbiscaro:
## Summary

In the event that the directory creation fails in the case of initial load we should propagate this error back up to the caller. 

## Context

Why is this issue important? Link any relevant discussions, designs, or documents.

## Tasks

- [ ] 

## Notes

Any extra notes or details that might be useful.

Issue #1863, opened 2025-09-02
[ISSUE] Use `base_path` for initial load


nbiscaro:
## Summary

It doesn't work with standalone deployment, which requires all files to be placed under backend `base_path`.

## Context

Why is this issue important? Link any relevant discussions, designs, or documents.

## Tasks

- [ ] 

## Notes

Any extra notes or details that might be useful.

PR #1865, opened 2025-09-02
Use `base_path` for initial load

nbiscaro:
## Summary

We need to place files under `base_path` during initial copy in order to be compatible with Moonlink standalone. Also propagate any errors to caller on dir creation. 

## Related Issues

Closes https://github.com/Mooncake-Labs/moonlink/issues/1863
Closes https://github.com/Mooncake-Labs/moonlink/issues/1864

## Changes

- Pass `table_base_path` to `perform_initial_copy`
- Add unit test for standalone
- Propagate directory creation error
- Panic at caller for initial copy
- Add error case test

## Checklist

- [ ] Code builds correctly

### DIFF
--- a/src/moonlink_connectors/src/pg_replicate.rs
+++ b/src/moonlink_connectors/src/pg_replicate.rs
@@ -264,7 +264,10 @@ impl PostgresConnection {
                 None,
             )
             .await
-            .expect("failed to copy table");
+            .expect(&format!(
+                "failed to copy table for src_table_id: {}",
+                src_table_id
+            ));
 
             // Commit the transaction
             copy_source


src/moonlink_connectors/src/pg_replicate/initial_copy.rs
```diff
@@ -34,6 +34,7 @@ pub async fn copy_table_stream<S>(
     mut stream: S,
     event_sender: &Sender<TableEvent>,
     start_lsn: u64,
+    base_path: &str,
```
dentiny: nit: do you mind renaming it to `table_base_path`?
- base_path is what we use for moonlink backend [here](https://github.com/Mooncake-Labs/moonlink/blob/40f9eb9af36f3fbca3ecd898c2930ebd4990fdd5/src/moonlink_backend/src/lib.rs#L56)
- table_base_path is made for mooncake table write through cache directory

Having consistent naming helps code readability. :)

dentiny:
> test table_handler::chaos_test::test_upsert_chaos_with_data_compaction ... ok
test table_handler::chaos_test::test_upsert_chaos_with_no_background_maintenance ... ok
test table_handler::chaos_test::test_upsert_chaos_with_index_merge ... ok

Weird, all tests look passing

src/moonlink_connectors/src/pg_replicate.rs
```diff
@@ -254,18 +255,16 @@ impl PostgresConnection {
...
+                table_base_path,
                 None,
             )
-            .await;
-
-            if let Err(e) = res {
-                error!(error = ?e, table_id = src_table_id, "failed to copy table");
-            }
+            .await
+            .expect("failed to copy table");
```
dentiny: The advantage for `expect` over `unwrap` is it allows more context, I would suggest include table name (or table base path) here.

Issue #1846, opened 2025-09-01
BUG: flaky reconnection test


dentiny:
## Summary

```sh
thread 'tests::test_replication_connection_cleanup' panicked at src/moonlink_backend/tests/test_backend.rs:226:9:
assertion `left == right` failed
  left: {}
 right: {1}

thread 'tokio-runtime-worker' panicked at src/moonlink_connectors/src/replication_state.rs:37:31:
called `Result::unwrap()` on an `Err` value: SendError { .. }
2025-08-30T04:46:15.993005Z  WARN moonlink_connectors::replication_connection: task join error during shutdown error=JoinError::Panic(Id(54667), "called `Result::unwrap()` on an `Err` value: SendError { .. }", ...)
```
link: https://github.com/Mooncake-Labs/moonlink/actions/runs/17339491395/job/49231368789

Feel free to re-assign.

PR #1868, opened 2025-09-02
No panic on mark

nbiscaro:
## Summary

During startup or shutdown `mark` may be called (on keepalive) causing a panic if the subscriber is dropped. Here we change mark to best-effort. 

## Related Issues

Closes https://github.com/Mooncake-Labs/moonlink/issues/1846

## Changes

- Change mark to best effort
- Add regression test
- 

## Checklist

- [ ] Code builds correctly

### DIFF
--- a/src/moonlink_connectors/src/replication_state.rs
+++ b/src/moonlink_connectors/src/replication_state.rs
@@ -34,7 +34,8 @@ impl ReplicationState {
     pub fn mark(&self, lsn: u64) {
         if lsn > self.current.load(Ordering::SeqCst) {
             self.current.store(lsn, Ordering::SeqCst);
-            self.tx.send(lsn).unwrap();
+            // Ignore send error if there are no subscribers (e.g., during shutdown)
+            let _ = self.tx.send(lsn);
         }
     }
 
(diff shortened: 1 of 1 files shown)


src/moonlink_connectors/src/replication_state.rs
```diff
@@ -34,7 +34,8 @@ impl ReplicationState {
     pub fn mark(&self, lsn: u64) {
         if lsn > self.current.load(Ordering::SeqCst) {
             self.current.store(lsn, Ordering::SeqCst);
-            self.tx.send(lsn).unwrap();
+            // Ignore send error if there are no subscribers (e.g., during shutdown)
+            let _ = self.tx.send(lsn);
```
dentiny: I would suggest add a logging here.

dentiny: Two reasons:
- The failure should only happen in rare cases, so logging shouldn't floods the whole screen
- Replication LSN is one of the core concept for moonlink; assume we have a bug somewhere, people won't even be able tell replication LSN is incorrect without logging
- preferred, but not sure how hard it is: we better do proper synchronization

PR #1861, opened 2025-09-02
integrate rest catalog into iceberg e2e tests

peterxcli:
## Summary

integrate rest catalog into iceberg e2e tests

## Changes

- 
- 
- 

## Checklist

- [ ] Code builds correctly
- [ ] Tests have been added or updated
- [ ] Documentation updated if necessary
- [ ] I have reviewed my own changes

PR #1875, opened 2025-09-03
[rest-catalog]: implement namespace functions

### LINKED ISSUE #1616: iceberg: implement namespace functions
## Summary

we could use iceberg-rust's rest catalog implementation and make a wrapper
this ticket is supposed to achieve a few namespace functionalities:
- create: https://github.com/apache/iceberg-rust/blob/fa5f2ed8dcd60be7438c5a3bfe098490ecd03e3a/crates/iceberg/src/catalog/mod.rs#L58
- get: https://github.com/apache/iceberg-rust/blob/fa5f2ed8dcd60be7438c5a3bfe098490ecd03e3a/crates/iceberg/src/catalog/mod.rs#L65
- existence: https://github.com/apache/iceberg-rust/blob/fa5f2ed8dcd60be7438c5a3bfe098490ecd03e3a/crates/iceberg/src/catalog/mod.rs#L68
- drop: https://github.com/apache/iceberg-rust/blob/fa5f2ed8dcd60be7438c5a3bfe098490ecd03e3a/crates/iceberg/src/catalog/mod.rs#L81

Some simple tests are enough

Standing-Man:
## Summary

implement namespace functions for `rest_catalog`

## Related Issues

- Closes #1616

## Changes

- implement namespace functions
- add more test cases for `namespace` function
- 🚧 refactor the RAII-Style Test Guard on `RestCatalog`.

## Checklist

- [x] Code builds correctly


dentiny:
curious how did you test it on local? Does it work for devcontainer?

Standing-Man:
> curious how did you test it on local? Does it work for devcontainer?

The tests failed locally, so I created this PR to verify them in the CI environment. 😂  I’m curious if there’s a way for us to test this locally.

peterxcli:
> > curious how did you test it on local? Does it work for devcontainer?
> 
> The tests failed locally, so I created this PR to verify them in the CI environment. 😂 I’m curious if there’s a way for us to test this locally.

@Standing-Man would you like to try https://github.com/Mooncake-Labs/moonlink/pull/1861/commits/755ff48691433f7b72cc7aed89b6b925844f05c8 to see if it helps?

Standing-Man:
> @Standing-Man would you like to try [755ff48](https://github.com/Mooncake-Labs/moonlink/commit/755ff48691433f7b72cc7aed89b6b925844f05c8) to see if it helps?

Awesome PR — really nice job! Hopefully it’ll be merged soon. I’ll keep an eye on it and try to help out whenever I can.

src/moonlink/src/storage/iceberg/rest_catalog.rs
```diff
@@ -68,10 +69,12 @@ impl Catalog for RestCatalog {
 
     async fn update_namespace(
         &self,
-        _namespace_ident: &NamespaceIdent,
-        _properties: HashMap<String, String>,
+        namespace_ident: &NamespaceIdent,
+        properties: HashMap<String, String>,
     ) -> IcebergResult<()> {
-        todo!("Update namespace is not supported");
+        self.catalog
```
dentiny: FYI: moonlink doesn't need for now, file catalog also didn't implement it

src/moonlink/src/storage/iceberg/rest_catalog_test_utils.rs
```diff
@@ -25,6 +28,29 @@ pub(crate) fn default_rest_catalog_config() -> RestCatalogConfig {
     }
 }
 
+/// Util function to transaction vec into set
+pub(crate) fn to_set<T: Eq + Hash>(vec: Vec<T>) -> HashSet<T> {
```
dentiny: not sure how useful it is, if you really want it, better to put it into a more centralized place so others could also use it.

src/moonlink/src/storage/iceberg/rest_catalog_test_utils.rs
```diff
@@ -25,6 +28,29 @@ pub(crate) fn default_rest_catalog_config() -> RestCatalogConfig {
...
+}
+
+/// Util function to test multiple_namespaces
+pub(crate) async fn create_namespaces(
+    catalog: &RestCatalog,
+    namespace_idents: &Vec<&NamespaceIdent>,
+) {
+    for namespace_ident in namespace_idents {
+        let _ = catalog
+            .create_namespace(namespace_ident, HashMap::new())
```
dentiny: nit:
```suggestion
            .create_namespace(namespace_ident, /*properties=*/HashMap::new())
```

dentiny (commented):
only nitpicks

dentiny:
Also need to address CI test failures :)

src/moonlink/src/storage/iceberg/rest_catalog_test_utils.rs
Standing-Man: Got it, I’ve fixed these minor issues.

Standing-Man:
The test failures in the current CI run are caused by namespaces not being cleaned up. I observed that prior table operations relied on `rest_catalog_guard` to manage both namespaces and tables, and the drop method requires recreating the rest_catalog. Could we explore a more effective way to handle these namespaces and tables?

dentiny:
unit tests still fail, please fix them first, thanks!

dentiny:
> The test failures in the current CI run are caused by namespaces not being cleaned up.

It's hard, since all unit tests run concurrently.
I think all unit tests could create their own namespace, and only do CRUD + list operation on namespaces inside of the parent namespace.

dentiny:
Ping me when ready, thanks!

src/moonlink/src/storage/iceberg/rest_catalog_test_guard.rs
```diff
@@ -10,13 +10,18 @@ pub(crate) struct RestCatalogTestGuard {
...
-    pub(crate) async fn new(namespace: String, table: Option<String>) -> Result<Self> {
+    pub(crate) async fn new(
+        namespace: Vec<String>,
+        table: Option<String>,
+        props: Option<HashMap<String, String>>,
+    ) -> Result<Self> {
         let config = default_rest_catalog_config();
         let catalog = RestCatalog::new(config)
             .await
             .expect("Catalog creation fail");
```
Standing-Man: Hi @dentiny, we set a Guard to protect the created catalog, but the test case still creates another catalog to call the CRUD API. In this case, what is the intended purpose of the Guard?

dentiny: I think the correct way is to create isolated catalog and namespace for each unit test, I suspect current implementation suffers data race.

Standing-Man: Right, I think the Guard is created to protect the catalog and avoid data races.

dentiny: Feel free to make changes you think reasonable, also welcome separate PRs to fix issues.
Thanks for the help!

Standing-Man:
Hi @dentiny, I don’t think **CI** errors comes from data race. In `test_list_namespaces_returns_empty_vector`, we don’t create any namespace, but `list_namespace` still shows multiple ones.

dentiny:
>  In test_list_namespaces_returns_empty_vector, we don’t create any namespace, but list_namespace still shows multiple ones.

- Are all rest catalog logs running in sequantially?
- Is there anything inside before test (i.e., with empty test case alone)?

src/moonlink/src/storage/iceberg/rest_catalog_test_guard.rs
```diff
@@ -1,55 +1,92 @@
...
 use crate::storage::iceberg::rest_catalog_test_utils::*;
-use iceberg::{Catalog, NamespaceIdent, Result, TableIdent};
-use std::collections::HashMap;
+use iceberg::{Catalog, NamespaceIdent, TableIdent};
+use tokio::sync::RwLock;
 
 pub(crate) struct RestCatalogTestGuard {
-    pub(crate) namespace: NamespaceIdent,
-    pub(crate) table: Option<TableIdent>,
+    pub(crate) catalog: Arc<RwLock<RestCatalog>>,
```
dentiny: curious why lock for the data member?

Standing-Man: Use `RwLock` only to ensure catalog is not accessed by multiple threads (can be removed later)

dentiny: sorry not quite sure about that, isn't one test guard used in one test case?

Standing-Man: Yes, you're right — usually one test guard is used per test case.

src/moonlink/src/storage/iceberg/rest_catalog_test_utils.rs
dentiny: Comment unaddressed?

dentiny:
Do you mind resolving the conflicts? Thanks!

peterxcli:
@Standing-Man would you like to put those test refactor into other PR?

Standing-Man:
> @Standing-Man would you like to put those test refactor into other PR?

I’d like to merge the namespace methods first, and handle the TestGuard refactor in a separate branch later. Is that okay?@dentiny.

dentiny:
> I’d like to merge the namespace methods first,

Let's do namespace change first, thank you!

dentiny (commented):
The code LG, it's delegating the invocation to internal rest catalog, but not sure why you delete all unit tests?

Standing-Man:
> The code LG, it's delegating the invocation to internal rest catalog, but not sure why you delete all unit tests?

Since those unit tests could not pass the CI checks, I removed them.

dentiny:
> Since those unit tests could not pass the CI checks, I removed them.

TBH I don't think that's acceptable, what's the issue?

src/moonlink/src/storage/iceberg/rest_catalog_test.rs
```diff
@@ -1,108 +1,154 @@
...
+        .expect("error: fail to check if the namespace exist"));
+
+    assert_eq!(
+        writer
+            .catalog
+            .list_namespaces(None)
+            .await
+            .expect("error: fail to list the namespaces"),
+        vec![]
+    );
```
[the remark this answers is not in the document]
Standing-Man: Got it!, i will update on next commit, thanks.

Standing-Man:
> TBH I don't think that's acceptable, what's the issue?

Currently, the table and namespace tests are merged into two cases. I plan to address the `RestCatalogTestGuard` refactor in a follow-up PR.

dentiny (commented):
```
thread 'storage::iceberg::rest_catalog_test::test_namespace_operation' panicked at src/moonlink/src/storage/iceberg/rest_catalog_test.rs:120:5:
assertion `left == right` failed
  left: [NamespaceIdent(["7AGXXSRGmF"]), NamespaceIdent(["Un1iYYNGh7"]), NamespaceIdent(["ul0DMIc12e"])]
 right: [NamespaceIdent(["ul0DMIc12e"])]
```
CI fails

src/moonlink/src/storage/iceberg/rest_catalog_test.rs
```diff
@@ -1,108 +1,157 @@
...
+    assert!(writer
+        .catalog
+        .namespace_exists(&ns_ident)
+        .await
+        .expect("error: fail to check if the namespace exist"));
+
+    assert_eq!(
+        writer
+            .catalog
```
dentiny, line 122: You are listing namespaces in **root namespace**, rather than your own unique **namespace**.

src/moonlink/src/storage/iceberg/rest_catalog_test.rs
```diff
@@ -1,108 +1,141 @@
...
-        create_test_table_schema().unwrap(),
-    )
-    .await
-    .unwrap();
-    let namespace = NamespaceIdent::new(namespace);
-    let table_creation = default_table_creation(table.clone());
-    let table_name = table_creation.name.clone();
-    catalog
-        .create_table(&namespace, table_creation)
+        .expect("error: fail to create testguard");
```
dentiny: just curious why do you switch back to expect here?

src/moonlink/src/storage/iceberg/rest_catalog_test_guard.rs
```diff
@@ -3,15 +3,17 @@ use crate::storage::iceberg::rest_catalog::RestCatalog;
...
 use iceberg::{Catalog, NamespaceIdent, Result, TableIdent};
-use std::collections::HashMap;
+use std::sync::Arc;
+use tokio::sync::RwLock;
 
 pub(crate) struct RestCatalogTestGuard {
-    pub(crate) namespace: NamespaceIdent,
-    pub(crate) table: Option<TableIdent>,
+    pub(crate) catalog: Arc<RwLock<RestCatalog>>,
+    pub(crate) namespace_idents: Option<Vec<NamespaceIdent>>,
```
dentiny, line 9: why do we use arc here?

Standing-Man: Would it be correct to use only RestCatalog here?

src/moonlink/src/storage/iceberg/rest_catalog_test.rs
Standing-Man: In the next commit, I will handle the error upload in the new function.

dentiny: > https://github.com/Mooncake-Labs/moonlink/issues/2005

Could you please elaborate what do you plan to do?

My thought: 
- In unit test, unwrap / panic / assert is enough, no need for proper error propagation
- The advantage for expect is more context; say, you need extra information when error happens, feel free to use expect to fill in more details, but don't think it's the case here.

src/moonlink/src/storage/iceberg/rest_catalog_test_guard.rs
dentiny: I think the default option is to use value directly (perf wise, code complexity wise);
you choose to use Arc and rwlock must be some reasons, right? :sweat_smile:

src/moonlink/src/storage/iceberg/rest_catalog_test.rs
Standing-Man: Using `expect` helps surface errors when creating the catalog.

dentiny: What's the advantage for expect over unwrap here

Standing-Man: that’s the intended usage.
```
pub(crate) async fn new() -> Result<Self> {
        let rest_catalog_config = default_rest_catalog_config();
        let accessor_config = default_accessor_config();
        let catalog = RestCatalog::new(
            rest_catalog_config,
            accessor_config,
            create_test_table_schema().unwrap(),
        )
        .await?;
        Ok(Self {
            catalog: Arc::new(RwLock::new(catalog)),
            namespace_idents: None,
            tables_idents: None,
        })
    }

------
let mut guard = RestCatalogTestGuard::new()
        .await
        .expect("error: fail to create tes tguard");
```
If we use `unwrap`, any error passed through the new function won’t provide a clear error message.

src/moonlink/src/storage/iceberg/rest_catalog_test_guard.rs
Standing-Man: Using `Arc` helps prevent extensive data duplication.

src/moonlink/src/storage/iceberg/rest_catalog_test.rs
dentiny: > that’s the intended usage.

My point: 

> In unit test, unwrap / panic / assert is enough, no need for proper error propagation

---

> The advantage for expect is more context; say, you need extra information when error happens, feel free to use expect to fill in more details, but don't think it's the case here.

I don't see how expect provides more context here?

src/moonlink/src/storage/iceberg/rest_catalog_test_guard.rs
dentiny: Could you please elaborate? say, which data gets duplicated if we don't use arc?

Standing-Man:
```
pub struct RestCatalog {
    pub(crate) catalog: IcebergRestCatalog,
    file_io: FileIO,
    warehouse_location: String,
    iceberg_schema: Option<IcebergSchema>,
    deletion_vector_blobs_to_add: HashMap<String, Vec<PuffinBlobMetadataProxy>>,
    file_index_blobs_to_add: HashMap<String, Vec<PuffinBlobMetadataProxy>>,
    puffin_blobs_to_remove: HashSet<String>,
    data_files_to_remove: HashSet<String>,
}
```
Some HashMaps and HashSets may hold large data. Since strings, vectors, and hashmaps are **heap-allocated**, frequent cloning can hurt performance.

src/moonlink/src/storage/iceberg/rest_catalog_test.rs
Standing-Man: Got it, I’ll just use unwrap to handle the error.

src/moonlink/src/storage/iceberg/rest_catalog_test_guard.rs
dentiny: these data structures are heap allocated instead of stack

Standing-Man: I’m wondering, when we clone these heap-allocated types, does it duplicate the underlying data or just the structure?

dentiny: It's deep copy, if youre not using shared pointer

Standing-Man: Exactly, that’s why we use Arc—to avoid deep copies.

dentiny: Why do we need to copy test guard

Standing-Man: In the future, when consolidating the test cases, I’m not sure if references would work—what do you think?


https://github.com/Mooncake-Labs/moonlink/blob/main/src/moonlink/src/storage/iceberg/rest_catalog_test.rs#L174

Standing-Man: I’d appreciate any advice — I can address it in my next commit.

dentiny: - reference means exclusive usage (in rust)
- shared pointer means immutable ownership share

For rest catalog test, I imagine 
- one unit test should have one guard to construct and cleanup resource
- one rest catalog to perform catalog operation

The ownership seems very clear to me. If you think rest catalog could be shared, do you mind giving me an example?

Standing-Man: The ownership seems not very clear to me. 😂

dentiny: could you please tell me, in your mind, what's the relationship between (1) unit test; (2) test guard; (3) rest catalog?

dentiny: When I say "relationship", I mean one-to-one, one-to-many, many-to-many

Standing-Man: I only know one unit test have one test guard.

dentiny: That's not good answer
- Currently all our unit tests have relationship that: one unit test has one rest catalog, one test guard maps one rest catalog
- If you want to propose something else, you should provide justification why it's necessary

src/moonlink/src/storage/iceberg/rest_catalog_test.rs
```diff
@@ -1,108 +1,96 @@
...
-async fn test_drop_table() {
-    let namespace = get_random_string();
-    let table = get_random_string();
-    let mut guard = RestCatalogTestGuard::new(namespace.clone(), Some(table.clone()))
+    // create a namespace
+    let ns_name = get_random_string();
+    let ns_ident = NamespaceIdent::new(ns_name);
+    guard
+        .catalog
+        .create_namespace(&ns_ident, HashMap::new())
```
dentiny: nit:
```suggestion
        .create_namespace(&ns_ident, /*properties=*/HashMap::new())
```

src/moonlink/src/storage/iceberg/rest_catalog_test.rs
```diff
@@ -1,108 +1,96 @@
...
+    assert!(guard.catalog.table_exists(&table_ident).await.unwrap());
+    // check the list table method
+    assert_eq!(
+        guard.catalog.list_tables(&ns_ident).await.unwrap(),
+        vec![table_ident.clone()]
+    );
 
-    // Load table and check schema.
-    let table = catalog.load_table(&table_ident).await.unwrap();
-    let actual_schema = table.metadata().current_schema();
```
dentiny, line 82: Why do you drop the schema test?

src/moonlink/src/storage/iceberg/rest_catalog_test.rs
```diff
@@ -1,108 +1,96 @@
...
-    assert_is_same_schema(actual_schema.as_ref().clone(), iceberg_schema);
+    // drop the table
+    guard.catalog.drop_table(&table_ident).await.unwrap();
+
+    guard.remove_table(&table_ident);
+
+    assert!(!guard.catalog.table_exists(&table_ident).await.unwrap());
+    assert_eq!(guard.catalog.list_tables(&ns_ident).await.unwrap(), vec![]);
+
+    guard.cleanup().await;
```
dentiny: why do we need to cleanup manually? I think the whole purpose for test guard is dtor

src/moonlink/src/storage/iceberg/rest_catalog_test_guard.rs
```diff
@@ -3,15 +3,15 @@ use crate::storage::iceberg::rest_catalog::RestCatalog;
...
-    pub(crate) namespace: NamespaceIdent,
-    pub(crate) table: Option<TableIdent>,
+    pub(crate) catalog: RestCatalog,
+    pub(crate) namespace_idents: Option<Vec<NamespaceIdent>>,
+    pub(crate) tables_idents: Option<Vec<TableIdent>>,
 }
 
 impl RestCatalogTestGuard {
-    pub(crate) async fn new(namespace: String, table: Option<String>) -> Result<Self> {
+    pub(crate) async fn new() -> Result<Self> {
```
dentiny: nit: no need for result

src/moonlink/src/storage/iceberg/rest_catalog_test_guard.rs
```diff
@@ -21,44 +21,51 @@ impl RestCatalogTestGuard {
...
+        }
+    }
+
+    pub(crate) async fn cleanup(&mut self) {
+        let tables = self.tables_idents.take();
+        let namespaces = self.namespace_idents.take();
+
+        if let Some(tables) = tables {
+            for t in tables {
+                let _ = self.catalog.drop_table(&t).await;
```
dentiny: why ignore the error?

src/moonlink/src/storage/iceberg/rest_catalog_test_guard.rs
```diff
@@ -21,44 +21,51 @@ impl RestCatalogTestGuard {
...
+            }
+        }
+
+        if let Some(namespaces) = namespaces {
+            for ns in namespaces {
+                let _ = self.catalog.drop_namespace(&ns).await;
```
dentiny: same question here

src/moonlink/src/storage/iceberg/rest_catalog_test.rs
Standing-Man: Since the `Drop` function cannot directly perform asynchronous cleanup the `table` and `namespace`, the way to run async code there is to spawn a background task that handles the cleanup inside `drop`.

dentiny: You can do https://github.com/Mooncake-Labs/moonlink/blob/c4a7d6cb28c5a7818d820cd6e0676064bcf6a618/src/moonlink/src/storage/filesystem/s3/test_guard.rs#L21

Standing-Man: Might this lead to a deadlock?

dentiny: No

src/moonlink/src/storage/iceberg/rest_catalog_test.rs
```diff
@@ -7,102 +9,128 @@ use crate::storage::iceberg::schema_utils::assert_is_same_schema;
...
-async fn test_drop_table() {
-    let namespace = get_random_string();
-    let table = get_random_string();
-    let mut guard = RestCatalogTestGuard::new(namespace.clone(), Some(table.clone()))
+    guard.record_namespace(ns_ident.clone());
+
+    // create a table
+    let table_name = get_random_string();
+    let creation = default_table_creation(table_name.clone());
+    guard
```
dentiny, line 29: nit: you don't need to create `TableIdent` yourself, simply get it from returned `Table`

src/moonlink/src/storage/iceberg/rest_catalog_test_guard.rs
```diff
@@ -21,43 +21,56 @@ impl RestCatalogTestGuard {
...
-            table: table_ident,
-        })
+        Self {
+            catalog,
+            namespace_idents: None,
+            tables_idents: None,
+        }
+    }
+
+    pub(crate) fn record_table(&mut self, table_ident: TableIdent) {
```
dentiny, line 31: On record, should we check whether the table hasn't been recorded before?
On removal, should we check whether it does exist?

src/moonlink/src/storage/iceberg/rest_catalog_test_guard.rs
```diff
@@ -2,16 +2,16 @@ use crate::storage::iceberg::catalog_test_utils::create_test_table_schema;
...
 /// A RAII-style test guard, which creates namespace ident, table ident at construction, and deletes at destruction.
 use crate::storage::iceberg::rest_catalog_test_utils::*;
-use iceberg::{Catalog, NamespaceIdent, Result, TableIdent};
-use std::collections::HashMap;
+use iceberg::{Catalog, NamespaceIdent, TableIdent};
 
 pub(crate) struct RestCatalogTestGuard {
-    pub(crate) namespace: NamespaceIdent,
-    pub(crate) table: Option<TableIdent>,
+    pub(crate) catalog: RestCatalog,
```
dentiny, line 8: I think the purpose for test guard is to automatically delete namespaces and tables, in case we forget or an error happens. If all (un)registration needs to do manually, I feel like we kinda lose the purpose.

dentiny: I'm wondering why do we change the previous implementation (i.e., create namespace and optional table at ctor).

Standing-Man: > I think the purpose for test guard is to automatically delete namespaces and tables, in case we forget or an error > happens. If all (un)registration needs to do manually, I feel like we kinda lose the purpose.

Why have a test guard if it doesn’t automatically clean up namespaces and tables when we forget or when an error occurs?

> I'm wondering why do we change the previous implementation (i.e., create namespace and optional table at ctor).

Why do we need to create the namespace and table in the constructor? What are the benefits?

src/moonlink/src/storage/iceberg/rest_catalog_test_guard.rs
Standing-Man: Why don’t we validate that a table isn’t already recorded on insert, and that it exists on removal? do we need it?

src/moonlink/src/storage/iceberg/rest_catalog_test_guard.rs
dentiny: > Why have a test guard if it doesn’t automatically clean up namespaces and tables when we forget or when an error occurs?

Sorry I don't understand, could you please elaborate?

For example, I think current implementation in main branch does cleanup at dtor.

https://github.com/Mooncake-Labs/moonlink/blob/893f6a0df06d93343af80ea750ffeb45698d0ede/src/moonlink/src/storage/iceberg/rest_catalog_test_guard.rs#L57-L60

dentiny: > Why do we need to create the namespace and table in the constructor? What are the benefits?

That's [RAII](https://en.wikipedia.org/wiki/Resource_acquisition_is_initialization).

src/moonlink/src/storage/iceberg/rest_catalog_test_guard.rs
dentiny: > Why don’t we validate that a table isn’t already recorded on insert

That's my point.

I hope you could validate the precondition for insertion and removal.

src/moonlink/src/storage/iceberg/rest_catalog_test_guard.rs
Standing-Man: It’s possible that the RestCatalogTestGuard I implemented doesn’t fully follow a pure RAII mechanism.

Standing-Man:
Since this PR does not provide significant improvements to the `RestCatalogTestGuard`, we’ll be closing it. If anyone is interested in this issue, please feel free to pick it up.

PR #1877, opened 2025-09-03
composite types for REST api

### LINKED ISSUE #1876: [kafka] nested schema in REST API
## Summary

A common pattern in kafka logs is nested structs - we want to be able to support this in our REST API which the kafka connector will send events to.

lohpaul9:
## Summary

Allows for nested structs / composite type schema in the REST API. This allows us to support nested structs using the Kafka connector too, which is currently planned to connect to moonlink through the REST API. 

Closes #1876 

## Checklist

- [ ] Code builds correctly

### DIFF
--- a/src/moonlink_connectors/src/error.rs
+++ b/src/moonlink_connectors/src/error.rs
@@ -269,7 +269,7 @@ impl From<schema_builder::SchemaBuildError> for Error {
     fn from(source: schema_builder::SchemaBuildError) -> Self {
         Error::SchemaBuildError(ErrorStruct {
             message: "Schema building error".to_string(),
-            status: ErrorStatus::Temporary,
+            status: ErrorStatus::Permanent,
             source: Some(Arc::new(source.into())),
             location: Some(Location::caller().to_string()),
         })


src/moonlink_service/src/rest_api.rs
```diff
@@ -253,56 +259,144 @@ async fn create_table(
     // Convert field schemas to Arrow schema with proper field IDs (like PostgreSQL)
     use arrow_schema::{DataType, Field, Schema};
 
-    let mut field_id = 0;
+    fn parse_decimal(data_type_str: &str) -> Result<(u8, i8), String> {
```
dentiny: any chance to keep the comments? not sure why we delete them

src/moonlink_service/src/rest_api.rs
```diff
@@ -253,56 +259,144 @@ async fn create_table(
...
+            let scale: i8 = parts[1]
+                .parse()
+                .map_err(|_| format!("Invalid decimal scale in: {data_type_str}"))?;
+            Ok((precision, scale))
+        } else {
+            Err(format!("Invalid decimal type: {data_type_str}"))
+        }
+    }
+
+    fn build_field_from_schema(
```
dentiny: Document return value please

src/moonlink_service/src/rest_api.rs
```diff
@@ -253,56 +259,144 @@ async fn create_table(
...
+            }
+            // Struct type: { data_type: "struct", fields: [...] }
+            "struct" => {
+                let child_schemas = field_schema
+                    .fields
+                    .as_ref()
+                    .ok_or_else(|| format!("Missing 'fields' for struct '{}", field_schema.name))?;
+                let children: Result<Vec<Field>, String> = child_schemas
+                    .iter()
+                    .map(|child| build_field_from_schema(child, None, None, field_id))
```
dentiny:
```suggestion
                    .map(|child| build_field_from_schema(child, /*override_name=*/None, /*override_nullable=*/None, field_id))
```

src/moonlink_service/src/rest_api.rs
```diff
@@ -253,56 +259,144 @@ async fn create_table(
...
-            let mut metadata = HashMap::new();
-            metadata.insert("PARQUET:field_id".to_string(), field_id.to_string());
-            field_id += 1;
-
-            let field_with_metadata =
-                Field::new(&field.name, data_type, field.nullable).with_metadata(metadata);
-
-            Ok(field_with_metadata)
-        })
+        .map(|fs| build_field_from_schema(fs, None, None, &mut field_id))
```
dentiny:
```suggestion
        .map(|fs| build_field_from_schema(fs,  /*override_name=*/None, /*override_nullable=*/None, &mut field_id))
```

src/moonlink_connectors/src/rest_ingest/json_converter.rs
```diff
@@ -181,6 +181,49 @@ impl JsonToMoonlinkRowConverter {
...
             }
+            DataType::Struct(child_fields) => {
+                if let Some(obj) = value.as_object() {
+                    let mut values = Vec::with_capacity(child_fields.len());
+                    for child in child_fields {
+                        let child_name = child.name();
+                        let child_value = match obj.get(child_name) {
+                            Some(v) => v,
+                            None => {
+                                if child.is_nullable() {
```
dentiny: Need unit test in json converter for 
- struct parse succeed
- no such field in object but nullable
- no such field and required

src/moonlink_connectors/src/rest_ingest/json_converter.rs
```diff
@@ -181,6 +181,49 @@ impl JsonToMoonlinkRowConverter {
...
+                                    values.push(RowValue::Null);
+                                    continue;
+                                } else {
```
dentiny: No need for `else`
Ref: https://medium.com/swlh/return-early-pattern-3d18a41bba8

src/moonlink_connectors/src/rest_ingest/json_converter.rs
```diff
@@ -181,6 +181,49 @@ impl JsonToMoonlinkRowConverter {
                     Err(JsonToMoonlinkRowError::TypeMismatch(field.name().clone()))
                 }
             }
+            DataType::Struct(child_fields) => {
+                if let Some(obj) = value.as_object() {
+                    let mut values = Vec::with_capacity(child_fields.len());
+                    for child in child_fields {
```
dentiny: nit: up to you
```suggestion
                    for child_field_type in child_fields {
```

src/moonlink_connectors/src/rest_ingest/json_converter.rs
```diff
@@ -181,6 +181,49 @@ impl JsonToMoonlinkRowConverter {
                     Err(JsonToMoonlinkRowError::TypeMismatch(field.name().clone()))
                 }
             }
+            DataType::Struct(child_fields) => {
+                if let Some(obj) = value.as_object() {
```
dentiny, line 185: need unit test in json converter to check cases where value cannot be converted into object
- case-1: nulllable and not exist
- case-2: mismatched type

src/moonlink_connectors/src/rest_ingest/json_converter.rs
```diff
@@ -181,6 +181,49 @@ impl JsonToMoonlinkRowConverter {
...
+                                        JsonToMoonlinkRowError::TypeMismatch(format!(
+                                            "{}.{}",
+                                            field.name(),
+                                            existing_path
+                                        ))
+                                    }
+                                    other => other,
+                                }
+                            })?;
+                        values.push(std::mem::take(&mut converted));
```
dentiny:
```suggestion
                        values.push(converted);
```
I tried to avoid `std::mem::take` since it takes the ownership and allow later reuse with no compiler warning

src/moonlink_service/src/rest_api.rs
```diff
@@ -65,11 +65,17 @@ pub struct CreateTableRequest {
 }
 
 /// Field schema definition
-#[derive(Debug, Serialize, Deserialize)]
+#[derive(Debug, Serialize, Deserialize, Default)]
 pub struct FieldSchema {
     pub name: String,
     pub data_type: String,
     pub nullable: bool,
+    /// For `struct` type: list of child fields
```
dentiny: Should we annotation "default" for other types as well?

src/moonlink_service/src/rest_api.rs
dentiny: I'm think if it's better to alias precision and scale here...

src/moonlink_service/src/rest_api.rs
```diff
@@ -253,56 +259,144 @@ async fn create_table(
...
+                *field_id += 1;
+                Ok(field.with_metadata(metadata))
+            }
+            // List type: { data_type: "list", item: { ... } }
+            "list" | "array" => {
+                let item_schema = field_schema
+                    .item
+                    .as_ref()
+                    .ok_or_else(|| format!("Missing 'item' for list '{}", field_schema.name))?;
+                if item_schema.data_type == "struct" {
```
dentiny: array of array is also not supported

src/moonlink_service/src/rest_api.rs
```diff
@@ -253,56 +259,144 @@ async fn create_table(
...
+                    return Err(format!(
+                        "Array of structs '{}' is not supported",
+                        field_schema.name
+                    ));
+                }
+                let item_field =
```
dentiny: nit (up to you): add a comment on override name and nullable

src/moonlink_service/src/test.rs
```diff
@@ -315,6 +315,361 @@ async fn test_schema() {
     assert_eq!(response.lsn, 1);
 }
 
+#[tokio::test]
+#[serial]
+async fn test_schema_with_struct_and_list() {
```
dentiny: I'm not sure why we put unit test for service instead of schema / field conversion; this file is intended to test service-oriented features.

src/moonlink_service/src/rest_api.rs
lohpaul9: Do you mean adding something like:
```
    struct DecimalType {
        precision: u8,
        scale: i8,
    }
```

I've added it for clarity and re-added the comments.

src/moonlink_service/src/rest_api.rs
lohpaul9: Added documentation

src/moonlink_service/src/rest_api.rs
lohpaul9: Added changes

src/moonlink_service/src/rest_api.rs
lohpaul9: Added change

src/moonlink_connectors/src/rest_ingest/json_converter.rs
lohpaul9: Changed to just take ownership

src/moonlink_service/src/rest_api.rs
lohpaul9: Hmm - I don't think so - the default here is None which should be the correct default for the primitive types (primitive types should not have children or fields)

Although I did remove the addition of the 'default' annotation for the overall `FieldSchema` struct

src/moonlink_service/src/rest_api.rs
lohpaul9: Added limitation

src/moonlink_connectors/src/rest_ingest/json_converter.rs
lohpaul9: Added unit tests

src/moonlink_connectors/src/rest_ingest/json_converter.rs
lohpaul9: Added unit tests as per https://github.com/Mooncake-Labs/moonlink/pull/1877#discussion_r2317658294 too

src/moonlink_service/src/test.rs
lohpaul9: Good point - I'm going to move the schema conversion logic from moonlink service to moonlink connector and shift tests there too.

src/moonlink_service/src/rest_api.rs
dentiny: I was actually thinking
```rs
pub type Precision = ...;
pub type Scale = ...;
```
not a big difference

src/moonlink_service/src/rest_api.rs
dentiny: It's for compatibility --- so for old code parsing new data, you need to assign them a default value otherwise deser fails.

dentiny: > Although I did remove the addition of the 'default' annotation for the overall FieldSchema struct

It's not a prod issue for now since we haven't released, but ALL fields for serde should have default value and rename alias as well.

src/moonlink_connectors/src/rest_ingest/schema_builder.rs
```diff
@@ -0,0 +1,356 @@
...
+        Ok(DecimalType { precision, scale })
+    } else {
+        Err(SchemaBuildError::InvalidDecimal(data_type_str.to_string()))
+    }
+}
+
+/// Build an Arrow `Field` from a `FieldSchema`.
+///
+/// Returns an Arrow field and mutates the `field_id` as a side effect (used to
+/// populate Parquet field-id metadata similar to PostgreSQL column IDs).
```
dentiny: nit: maybe we remove the pg part comment?
```suggestion
/// populate Parquet field-id metadata).
```
- I think it's postgres oid
- for arrow / parquet, nested field has its own field id, but doesn't seem to be the case for pg oid

src/moonlink_connectors/src/rest_ingest/schema_builder.rs
```diff
@@ -0,0 +1,356 @@
...
+/// Returns an error if the field schema is invalid.
+fn build_field_from_schema(
+    field_schema: &FieldSchema,
+    override_name: Option<&str>,
+    field_id: &mut i32,
+) -> Result<Field, SchemaBuildError> {
```
dentiny, line 68: I suggest adding the error into the crate-wise centralized error: https://github.com/Mooncake-Labs/moonlink/blob/main/src/moonlink_service/src/error.rs
The reason is, we implemented e[rror source](https://github.com/Mooncake-Labs/moonlink/blob/9ea0944bccf74d477e0eeed20938ff373f86cab1/src/moonlink_error/src/error.rs#L35), which automatically records the source location of the error. Your implementation didn't.

src/moonlink_service/src/test.rs
```diff
@@ -344,9 +380,9 @@ async fn run_optimize_table_test(mode: &str) {
     });
     test_readiness_probe().await;
 
-    // Create test table.
+    // Create test table
```
dentiny: why this change lol

src/moonlink_service/src/test.rs
```diff
@@ -1108,3 +1135,67 @@ async fn test_create_table_with_invalid_config() {
...
+    });
+
+    let response = client
+        .post(format!("{REST_ADDR}/tables/{crafted_src_table_name}"))
+        .header("content-type", "application/json")
+        .json(&payload)
+        .send()
+        .await
+        .unwrap();
+    assert!(!response.status().is_success());
```
dentiny: suggest to include the `status` in the assertion, so at least we know what happened on failure

src/moonlink_service/src/test_utils.rs
```diff
@@ -42,6 +43,170 @@ pub(crate) fn create_test_arrow_batch() -> RecordBatch {
...
+                "emails": ["alice@example.com", "alice2@example.com"],
+                "location": {"lat": 37.7749, "lon": -122.4194}
+            },
+            "events": [1712345678901_i64, 1712345678902_i64]
+        }
+    });
+    insert_payload
+}
+
+/// Util function to create a nested test arrow schema matching get_create_table_payload_nested
```
dentiny: nit:
```suggestion
/// Util function to create a nested test arrow schema matching [`get_create_table_payload_nested`]
```

src/moonlink_connectors/src/rest_ingest/schema_builder.rs
```diff
@@ -0,0 +1,356 @@
...
+    pub data_type: String,
+    pub nullable: bool,
+    #[serde(default)]
+    pub fields: Option<Vec<FieldSchema>>, // for struct
+    #[serde(default)]
+    pub item: Option<Box<FieldSchema>>, // for list/array
+}
+
+#[derive(thiserror::Error, Debug)]
+pub enum SchemaBuildError {
```
dentiny, line 17: All the builder file (which ends with `xx_builder.rs`) I've seen in rust follows builder pattern, instead of a few util functions.

Example for iceberg-rust: https://github.com/apache/iceberg-rust/blob/main/crates/iceberg/src/spec/table_metadata_builder.rs

src/moonlink_connectors/src/rest_ingest/schema_builder.rs
```diff
@@ -0,0 +1,356 @@
+use arrow_schema::{DataType, Field, Schema};
+use serde::{Deserialize, Serialize};
+use std::collections::HashMap;
+
+#[derive(Debug, Clone, Serialize, Deserialize, PartialEq, Eq)]
+pub struct FieldSchema {
+    pub name: String,
+    pub data_type: String,
```
dentiny: comment on case-sensetivity

src/moonlink_connectors/src/rest_ingest/schema_builder.rs
```diff
@@ -0,0 +1,356 @@
...
+        "list" | "array" => {
+            let item_schema = field_schema.item.as_ref().ok_or_else(|| {
+                SchemaBuildError::InvalidSchema(format!(
+                    "Missing 'item' for list '{}'",
+                    field_schema.name
+                ))
+            })?;
+
+            if matches!(item_schema.data_type.as_str(), "list" | "array" | "struct") {
+                return Err(SchemaBuildError::InvalidSchema(format!(
```
dentiny:
```suggestion
                return Err(SchemaBuildError::UnsupportedType(format!(
```

src/moonlink_connectors/src/rest_ingest/json_converter.rs
```diff
@@ -181,6 +181,48 @@ impl JsonToMoonlinkRowConverter {
...
+            DataType::Struct(child_fields) => {
+                if let Some(obj) = value.as_object() {
+                    let mut values = Vec::with_capacity(child_fields.len());
+                    for child_field in child_fields {
+                        let child_name = child_field.name();
+                        let child_value = match obj.get(child_name) {
+                            Some(v) => v,
+                            None => {
+                                if !child_field.is_nullable() {
+                                    return Err(JsonToMoonlinkRowError::MissingField(format!(
```
dentiny, line 193: nit (not merge blocker): reorder happy path and error path

src/moonlink_connectors/src/rest_ingest/schema_builder.rs
lohpaul9: Removed the PG comment

src/moonlink_connectors/src/rest_ingest/schema_builder.rs
lohpaul9: But this is meant to be an error that is bubbled up to the REST API and returned to the client. Do we still want to use this the same way as the centralized error?

src/moonlink_service/src/test.rs
lohpaul9: Reverted :)

src/moonlink_connectors/src/rest_ingest/schema_builder.rs
lohpaul9: yes this makes sense - we can refactor it to a cleaner version in a future change?

lohpaul9: #1902

src/moonlink_connectors/src/rest_ingest/schema_builder.rs
lohpaul9: Added comment

src/moonlink_connectors/src/rest_ingest/schema_builder.rs
lohpaul9: Changed - thanks.

src/moonlink_connectors/src/rest_ingest/schema_builder.rs
dentiny: https://github.com/Mooncake-Labs/moonlink/blob/f269d93a40e20f1128f600c788e03f55d1d4d22b/src/moonlink_service/src/rest_api.rs#L375
I think thata what we do now

src/moonlink_connectors/src/rest_ingest/json_converter.rs
lohpaul9: #1903

src/moonlink_connectors/src/rest_ingest/schema_builder.rs
lohpaul9: Okay - I've changed it to use the crate level error

src/moonlink_connectors/src/error.rs
```diff
@@ -259,3 +263,15 @@ impl From<parquet::errors::ParquetError> for Error {
         })
     }
 }
+
+impl From<schema_builder::SchemaBuildError> for Error {
+    #[track_caller]
+    fn from(source: schema_builder::SchemaBuildError) -> Self {
+        Error::SchemaBuildError(ErrorStruct {
+            message: "Schema building error".to_string(),
+            status: ErrorStatus::Temporary,
```
dentiny: why temporary? I don't think it's retryable

src/moonlink_connectors/src/rest_ingest/json_converter.rs
dentiny: I could followup

src/moonlink_connectors/src/error.rs
lohpaul9: you're right - user would have to change the schema. Changed to permanent.

Issue #1790, opened 2025-08-29
[ISSUE] Add deterministic ordering between cmd_rx and stream


nbiscaro:
## Summary

There is no coordination between when the schema is actually added to the CdcStream and when logical replication events will begin streaming

[add_table_to_replication](https://github.com/Mooncake-Labs/moonlink/blob/d14c3a8ae95626670e72465d991eb216b5b37c1e/src/moonlink_connectors/src/pg_replicate.rs#L301) (adds the  TableSchema to the CdcStream)

[add_table to_publication] https://github.com/Mooncake-Labs/moonlink/blob/d14c3a8ae95626670e72465d991eb216b5b37c1e/src/moonlink_connectors/src/pg_replicate.rs#L208

## Context

Why is this issue important? Link any relevant discussions, designs, or documents.

## Tasks

- [ ] 

## Notes

Any extra notes or details that might be useful.

PR #1891, opened 2025-09-03
Add table handshake

nbiscaro:
## Summary

Currently no coordination between when the schema is actually added to the `CdcStream` and when logical replication events will begin streaming. 

In this PR we add a `oneshot` channel to notify the caller that the table has been added to the stream and its safe to proceed. 

We also do a small refactor to lazily create and start the replication connection as part of add table. Note that we now spawn the replication task first, then add the table. 

## Related Issues

Closes https://github.com/Mooncake-Labs/moonlink/issues/1790

## Changes

- 
- 
- 

## Checklist

- [ ] Code builds correctly

### DIFF
--- a/src/moonlink_connectors/src/replication_manager.rs
+++ b/src/moonlink_connectors/src/replication_manager.rs
@@ -43,6 +43,30 @@ impl ReplicationManager {
         }
     }
 
+    pub async fn get_or_create_connection(
+        &mut self,
+        src_uri: &str,
+    ) -> Result<&mut ReplicationConnection> {
+        let replication_connection = match self.connections.entry(src_uri.to_string()) {
+            Entry::Occupied(entry) => entry.into_mut(),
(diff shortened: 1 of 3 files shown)


src/moonlink_connectors/src/replication_manager.rs
```diff
@@ -62,23 +86,16 @@ impl ReplicationManager {
...
-                    )
-                    .await?;
-                    (entry.insert(replication_connection), true)
-                }
-            };
+
+        // Error handling: don't allow duplicate mooncake table id be registered.
+        if self.table_info.contains_key(&mooncake_table_id) {
+            return Err(Error::repl_duplicate_table(mooncake_table_id.to_string()));
+        }
```
nbiscaro, line 93: Moving this to top of function, we only need `mooncake_table_id` to perform the check.

Issue #1872, opened 2025-09-02
[Kafka] REST API moonlink row protobuf


lohpaul9:
## Summary

As the kafka connector is written in Java, we need some way to efficiently pass data around between platforms. We add an option to have moonlink rows passed around using protobuf instead of json for the REST API.

PR #1892, opened 2025-09-03
proto uses vec<u8> instead of bytes

lohpaul9:
## Summary

We were previously converting moonlink row value's protobuf `bytes` and `fixed_len_bytes` into rust Byte type - under the hood moonlink row value actually uses `vec<u8>`, so we directly let the proto type be `vec<u8>` instead of converting into the intermediate `Byte`.

## Related Issues

Follow on from #1872


dentiny (approved):
Thanks!

Issue #1895, opened 2025-09-04
[ISSUE] Expose payload len on XLogData


nbiscaro:
## Summary

Only holds generic data type `D` so can't currently get size in bytes. 

## Context

Why is this issue important? Link any relevant discussions, designs, or documents.

## Tasks

- [ ] 

## Notes

Any extra notes or details that might be useful.

PR #1896, opened 2025-09-04
Bump rust-postgres

nbiscaro:
## Summary

Pull in changes for: 
https://github.com/Mooncake-Labs/rust-postgres/pull/13

This allows us to compute the `end_lsn` per `XLogData` block, providing more granular level LSN data (e.g no longer have to wait for stream commit for lsn data). 

This will be used in an upcoming change related to respawning the replication task, however I also think there are applications for our WAL writer in that this allows us to store partial transactions/not have to start from the beginning of a streamed transaction in PG during recovery(?) 

## Related Issues

Closes http://github.com/Mooncake-Labs/moonlink/issues/1895

## Changes

- 
- 
- 

## Checklist

- [ ] Code builds correctly

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4053,9 +4053,9 @@ dependencies = [
 
 [[package]]
 name = "log"
-version = "0.4.27"
+version = "0.4.28"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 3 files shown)


dentiny:
image
would appreciate if we could use more descriptive commit message at merge :)

PR #1904, opened 2025-09-04
bump ubuntu build version

lohpaul9:
## Summary

We need to do ubuntu versioning changes as we currently build the moonlink binaries using the latest rust devcontainer and then port it onto a slim ubuntu image. Recently, we encountered a GLIBC_2.38 requirement and therefore bump up the ubuntu image to `24.04`.


dentiny (approved):
Thank you!

PR #1915, opened 2025-09-05
refactor]make schema builder follow builder pattern

Standing-Man:
## Summary

## Related Issues

## Changes

- 
- 
- 

## Checklist

- [ ] Code builds correctly


Standing-Man:
> curious why don't we do the simple thing to rename the file?

That makes sense. We could rename it to `field_scheme.rs` to make the naming more consistent. What do you think?

dentiny:
> That makes sense. We could rename it to `field_scheme.rs` to make the naming more consistent. What do you think?

sure, it's fine; I would rather call it schema util or sth alike, up to you.

Standing-Man:
> sure, it's fine; I would rather call it schema util or sth alike, up to you.

Thanks! I’ve updated the filename to `schema_util`.

src/moonlink_connectors/src/rest_ingest/schema_util.rs
```diff
@@ -0,0 +1,378 @@
...
+    UnsupportedType(String),
+}
+
+struct DecimalType {
+    precision: u8,
+    scale: i8,
+}
+
+#[derive(Debug, Clone, Serialize, Deserialize, PartialEq, Eq)]
+pub struct FieldSchemaBuilder {
```
dentiny: I'm wondering why do we still need the builder struct?

Standing-Man: I checked, and the `PartialEq` and `Eq` macros are indeed not needed. thanks

dentiny: Oh sorry for the confusion, actually I mean the `struct` itself

Standing-Man: Are you suggesting that we consolidate everything into one SchemaBuilder and have it generate recommendations?

dentiny: So some context
- we name the file as builder but didn't actually followed the builder pattern
- the issue is made to clean this up
- we could either write a real builder, should be kinda standard, with a series of setter apis and a build, maybe also annotate with builder attribute
- or rename the whole file and simply treat it as a util file, no code change needed

Standing-Man: Add more methods in FileSchemaBuilder to set and access related variables?

dentiny: If you want to follow (1), [this](https://github.com/apache/iceberg-rust/blob/main/crates/iceberg/src/spec/table_metadata_builder.rs) might be a good reference

Standing-Man: I’ve implemented a somewhat unconventional schema builder. Since the `name`, `data_type`, and `null_bool` variables are required from the beginning, I created a `default` method to initialize them. However, I’m not sure if this is the best approach. Do you have any suggestions for improvement?

Standing-Man: I’m curious why `data_type` is using `String` instead of `enum`. Wouldn’t it be better to use `enum` type?

src/moonlink_connectors/src/rest_ingest/schema_util.rs
```diff
@@ -0,0 +1,361 @@
...
+    pub nullable: bool,
+    #[serde(default)]
+    pub fields: Option<Vec<FieldSchemaBuilder>>, // for struct
+    #[serde(default)]
+    pub item: Option<Box<FieldSchemaBuilder>>, // for list/array
+}
+
+impl FieldSchemaBuilder {
+    #[must_use]
+    pub fn default(name: String, data_type: String, nullable: bool) -> Self {
```
dentiny: `default` has its own attribute, usually doesn't take any arguments

Standing-Man: So, should I be using the `new` method for this?

dentiny: yes

Standing-Man: Got it.

src/moonlink_connectors/src/rest_ingest/schema_util.rs
dentiny: Because it comes from http request payload: https://github.com/Mooncake-Labs/moonlink/blob/9e3af999848ea0d987e60f41b0aa99fa84a0919a/src/moonlink_service/src/rest_api.rs#L83

dentiny: > Wouldn’t it be better to use enum type?

Curious what's your proposal of using enum in the payload?

src/moonlink_connectors/src/rest_ingest/schema_util.rs
```diff
@@ -0,0 +1,359 @@
...
+    pub name: String,
+    pub data_type: String, // case insensitive
+    pub nullable: bool,
+    #[serde(default)]
+    pub fields: Option<Vec<FieldSchemaBuilder>>, // for struct
+    #[serde(default)]
+    pub item: Option<Box<FieldSchemaBuilder>>, // for list/array
+}
+
+impl FieldSchemaBuilder {
```
dentiny: Two comments:
- Builder usually use `set_xxx` or `update_xxx` to set field's value
  + Ref: https://github.com/apache/iceberg-rust/blob/main/crates/iceberg/src/spec/table_metadata_builder.rs
  + `fn name` sounds like a getter
- As I comment [here](https://github.com/Mooncake-Labs/moonlink/pull/1915#discussion_r2325027478), there're two ways, one is to implement a real builder struct, or we rename to util filename. not sure why do we still rename the file?

src/moonlink_connectors/src/rest_ingest/schema_util.rs
```diff
@@ -0,0 +1,359 @@
...
+
+    /// Build an Arrow `Field` from a `FieldSchema`.
+    ///
+    /// Returns an Arrow field and mutates the `field_id` as a side effect
+    /// Returns an error if the field schema is invalid.
+    fn build_field_from_schema(
+        &self,
+        override_name: Option<&str>,
+        field_id: &mut i32,
+    ) -> Result<Field, SchemaBuildError> {
```
dentiny: You might also need to refactor the error, essentially all errors should be a centralized error type in `crate::Error`, current implementation loses track of error source location

Feel free to implement in a separate PR.

src/moonlink_connectors/src/rest_ingest/schema_util.rs
```diff
@@ -0,0 +1,359 @@
...
+                metadata.insert("PARQUET:field_id".to_string(), field_id.to_string());
+                *field_id += 1;
+                Ok(Field::new(&name, list_type, nullable).with_metadata(metadata))
+            }
+            other => Err(SchemaBuildError::UnsupportedType(other.to_string())),
+        }
+    }
+}
+
+pub fn build_arrow_schema_impl(fields: &[FieldSchemaBuilder]) -> Result<Schema, SchemaBuildError> {
```
dentiny: Curious what's the benefit of using builder, instead of util function here?

src/moonlink_connectors/src/rest_ingest/schema_util.rs
Standing-Man: I feel that I may have made the feedback on this PR a bit confusing, so I’d like to take some time to organize my thoughts. 😵‍💫

src/moonlink_connectors/src/rest_ingest/schema_util.rs
Standing-Man: Got it, I’ll handle the error refactor after this PR is merged.

src/moonlink_connectors/src/rest_ingest/schema_util.rs
Standing-Man: Not 100% sure, but maybe `enum` would help manage the payload and help avoid typos in `data_type`.

src/moonlink_connectors/src/rest_ingest/schema_util.rs
Standing-Man: Since these functions just rely on `FieldSchemaBuilder` to create `Field`, I feel `FieldSchemaBuilder` or even `FieldSchema` itself would work better as util function.

Issue #1780, opened 2025-08-28
[ISSUE] Replication task reconnection


nbiscaro:
## Summary

Currently, in the event that PG connection drops our replication task will die and will not recover. We should implement a retry mechanism and recovery for the replication stream, and ensure that we hold on to our recover any state held in the replication task itself (schemas, stream, etc.)

## Context

Why is this issue important? Link any relevant discussions, designs, or documents.

## Tasks

- [ ] 

## Notes

Any extra notes or details that might be useful.

PR #1921, opened 2025-09-05
Replication task supervisor

nbiscaro:
## Summary

Allows respawning of the replication task in the event we lost our connection stream to PG. Retires the connection with backoff. 

We do this by wrapping our replication task in an outer supervisor loop. In the event the connection breaks we copy out any state we need from the existing stream (schemas) and then reinitialize with a new stream and rehydrate with the saved schemas.

When reconnecting the stream we will start from `last_confirmed_flush_lsn`. In this case dedup is tricky since there is in memory state we are holding in addition to the state that is already written. This means we will have dups between the in memory data and the initial data sent in the stream on restart. 

We dedup by tracking the last LSN we pulled off the stream for each frame (using https://github.com/Mooncake-Labs/rust-postgres/pull/13) and skipping anything less than our high watermark. 

## Related Issues

Closes https://github.com/Mooncake-Labs/moonlink/issues/1780

## Changes

- 
- 
- 

## Checklist

- [ ] Code builds correctly

### DIFF
--- a/src/moonlink_backend/tests/test_backend.rs
+++ b/src/moonlink_backend/tests/test_backend.rs
@@ -2,12 +2,12 @@ mod common;
 
 #[cfg(test)]
 mod tests {
-    use crate::common::{ids_from_state, nonunique_ids_from_state};
+    use crate::common::ids_from_state;
 
     use super::common::{
         assert_scan_ids_eq, crash_and_recover_backend_with_guard, create_backend_from_base_path,
         current_wal_lsn, get_database_uri, get_serialized_table_config, smoke_create_and_insert,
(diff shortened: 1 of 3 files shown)


src/moonlink_backend/tests/test_backend.rs
```diff
@@ -959,4 +978,315 @@ mod tests {
             .await;
         assert!(res.is_err());
     }
+
+    /// Reconnect resumes replication (single table)
+    #[cfg(feature = "test-utils")]
```
dentiny, line 983: Are you sure we want to guard it with "test-util" attribute? Is it actually tested in CI?

Issue #1928, opened 2025-09-05
[ISSUE] Refactor test_backend


nbiscaro:
## Summary

Currently most integration tests sit in a single file -- we should separate these out logically

## Context

Targeted integration testing, cleaner logs, etc. 

## Tasks

- [ ] 

## Notes

Any extra notes or details that might be useful.

Issue #1926, opened 2025-09-05
[ISSUE] Remove guard for replication recovery tests


nbiscaro:
## Summary

remove test-utils guard from `test_backend`

## Context

Why is this issue important? Link any relevant discussions, designs, or documents.

## Tasks

- [ ] 

## Notes

Any extra notes or details that might be useful.

PR #1930, opened 2025-09-05
Refactor integration tests

nbiscaro:
## Summary

We currently put most of our integration tests into `test_backend.rs`. We should logically separate this into separate files. This gives us more targeted testing options and makes logs cleaner when viewing issues with `--no-capture`.

We now have
`test_basic.rs`
`test_control_plane_recovery.rs`
`test_replication_recovery.rs`
`test_table_recovery.rs`
`test_wal_recovery.rs`
`test_initial_copy.rs`

Additionally, the pipeline currently only runs `test_backend.rs`, however there are integration tests outside of this which we are unintentionally(?) skipping. Change to use `*` to always run all current and future integration tests in this crate. 

## Related Issues

Closes https://github.com/Mooncake-Labs/moonlink/issues/1928
Closes https://github.com/Mooncake-Labs/moonlink/issues/1926

## Changes

- Refactor tests
- Update pipeline
- Remove `#[cfg(feature = "test-utils")]` for `replication_recovery.rs`

## Checklist

- [ ] Code builds correctly

### DIFF
--- a/src/moonlink_backend/tests/test_replication_recovery.rs
+++ b/src/moonlink_backend/tests/test_replication_recovery.rs
@@ -11,8 +11,6 @@ mod tests {
 
     use crate::common::nonunique_ids_from_state;
 
-    use crate::common::SRC_URI;
-
     async fn terminate_replication_new_conn() {
         let uri = get_database_uri();
         let (client, connection) = connect_to_postgres(&uri).await;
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -609,7 +609,7 @@ jobs:
         env:
           DATABASE_URL: postgresql://postgres:postgres@localhost:5432/postgres
         run: |
-          cargo test --package moonlink_backend --test test_backend
+          cargo test --package moonlink_backend --test '*'
```
dentiny, line 612: Thanks for catching!

PR #1932, opened 2025-09-06
Add replication LSN assertion

nbiscaro:
## Summary

We currently operate under the assumption that when we receive a `KeepAlive` message from postgres with LSN x, all CDC events with LSN < x have already been received. This is what is used to gate in [try_read](https://github.com/Mooncake-Labs/moonlink/blob/b436884ba031822652d20b400bcaf320df96abed/src/moonlink/src/union_read/read_state_manager.rs#L161), for example. 

I'm a little skeptical of this... lets add an assertion and make sure this is always true. 

## Related Issues

## Changes

- 
- 
- 

## Checklist

- [ ] Code builds correctly

### DIFF
--- a/src/moonlink_connectors/src/pg_replicate/moonlink_sink.rs
+++ b/src/moonlink_connectors/src/pg_replicate/moonlink_sink.rs
@@ -45,6 +45,9 @@ pub struct Sink {
     /// Streaming hot-path cache of the last processed (xid, table_id, lsn).
     /// Skips streaming state lookup when the next row has the same xid and table.
     streaming_last_key: Option<(u32, SrcTableId, u64)>,
+    /// Tracks the maximum LSN observed from primary keepalive messages.
+    /// Used to assert that subsequent LSN-bearing CDC events are not older.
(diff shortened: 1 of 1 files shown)


src/moonlink_connectors/src/pg_replicate/moonlink_sink.rs
```diff
@@ -192,6 +196,12 @@ impl Sink {
         match event {
             CdcEvent::Begin(begin_body) => {
                 debug!(final_lsn = begin_body.final_lsn(), "begin transaction");
+                assert!(
```
dentiny, line 199: Why don't we use more assertions so you do tl need to print the value.....

src/moonlink_connectors/src/pg_replicate/moonlink_sink.rs
```diff
@@ -236,6 +252,12 @@ impl Sink {
                     end_lsn = stream_commit_body.end_lsn(),
                     "stream commit"
                 );
+                assert!(
```
dentiny, line 255: Same here

Issue #1933, opened 2025-09-06
otel: dump otel data into mooncake table


dentiny:
## Summary

As titled, once http server endpoint has been setup, we need to dump data into mooncake table created.

PR #1936, opened 2025-09-06
[otel] Implement end-to-end otel metrics ingestion server

dentiny:
## Summary

This PR implements the HTTP server side for otel ingestion, which is able to take otel requests in protobuf format and write into mooncake table, so in theory with otel sdk it's working end-to-end.

Current implementation is not production-ready yet, which requires a few more features:
- It only supports metrics, I plan to add support tracing
- It only supports most important metrics, gauge / counter / histogram, plan to add other types of metrics in the future
- Needs better error propagation handling

## Related Issues

Closes https://github.com/Mooncake-Labs/moonlink/issues/1933

## Checklist

- [ ] Code builds correctly

### DIFF
--- a/src/moonlink_service/src/otel/metrics_handler.rs
+++ b/src/moonlink_service/src/otel/metrics_handler.rs
@@ -1,0 +1,201 @@
+use std::collections::HashSet;
+use std::sync::Arc;
+
+/// Handler to process metrics ingestion.
+use crate::error::{Error, Result};
+use crate::otel::otel_schema::otlp_metrics_gsh_schema;
+use crate::otel::otel_to_moonlink_pb;
+use crate::rest_api::ListTablesResponse;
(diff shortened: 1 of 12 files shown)


src/moonlink_service/src/otel/metrics_handler.rs
```diff
@@ -0,0 +1,201 @@
...
+            error!("Failed to ingest otel data: {:?}", response.unwrap_err());
+        }
+    }
+
+    /// Handle request for the incoming metrics request.
+    pub(crate) async fn handle_request(
+        &mut self,
+        request: ExportMetricsServiceRequest,
+    ) -> Result<ExportMetricsServiceResponse> {
+        // TODO(hjiang): Currently only supports metrics from one single table.
```
dentiny, line 190: This is a strong requirement, I will followup in the next PR.

Issue #1867, opened 2025-09-02
BUG: index reader read bits failure


dentiny:
## Summary

```sh
thread 'tokio-runtime-worker' panicked at src/moonlink/src/storage/index/persisted_bucket_hash_map.rs:724:18:
called `Result::unwrap()` on an `Err` value: Error { kind: UnexpectedEof, message: "failed to fill whole buffer" }
stack backtrace:
```
Code reference:
```rs
self.current_bucket_entry_end = self
                .bucket_reader
                .read_unsigned_var::<u32>(self.metadata.bucket_bits)
                .unwrap();
```
Seems we're not writing index content correctly.

PR #1947, opened 2025-09-06
[storage] Fix index flush on alignment

dentiny:
## Summary

The bug is:
- When alignment reaches boundary and requests to flush, the last flush only writes the active buffer and switch buffer index
- Now we have a complete byte left in the active buffer unflushed, the fix here is to flush twice: one for active buffer, one for the left byte

It's hard to write a proper unit test, since it's hard to trigger such boundary case :(
I checked with the previously failed chaos test replay, and confirmed it works now.

Thanks @zhousun for saving my life :sob:

## Related Issues

Closes https://github.com/Mooncake-Labs/moonlink/issues/1867

## Checklist

- [x] Code builds correctly


src/moonlink/src/storage/async_bitwriter.rs
```diff
@@ -175,6 +175,7 @@ impl<W: AsyncWrite + Unpin + Send + Sync, E: Endianness> BitWriter<W, E> {
     /// Pads the current bit queue with zeros until aligned to the next byte boundary.
     /// This should be called before the final flush to ensure the output is byte-aligned.
     /// Returns `true` if a flush is required due to buffer spill.
+    #[must_use]
```
dentiny, line 178: A post-merge second thought, if alignment is only involved at close, it's better to handle the `align_and_flush_all` semantics inside of bit writer.

PR #1961, opened 2025-09-07
[Chore] Restructure workspace 

dandedotdev:
## Summary

Refactor project structure by extracting common dependencies to the workspace level and ensuring all workspace members are properly included.

I also tried to understand why so many duplicate checks need to be skipped in `deny.toml`. At first, I thought it was due to some unnecessary default features being declared. However, I later realized it was because we explicitly [enabled the all-features check](https://github.com/Mooncake-Labs/moonlink/blob/main/deny.toml#L2C1-L2C13). Therefore, I only suppressed these warnings.

## Related Issues

_None_

## Changes

- Extracted common dependencies to the workspace level.
- Improved imports of workspace members.
- Suppressed `cargo-deny` warnings.

I tested with the following commands:

```shell
cargo test --features storage-gcs
cargo test --features storage-s3
```

> I also tried to test with `cargo test --all-features`. However, the chaos tests take a long time to complete.

## Call for Help

This is my first contribution to `moonlink`.

I might make some mistakes. I hope I could be helped to improve this PR and make it up to the standards of this community.

## Checklist

- [x] Code builds correctly

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,14 +1,15 @@
 [workspace]
 members = [
-    "src/moonlink",
-    "src/moonlink_backend",
-    "src/moonlink_connectors",
-    "src/moonlink_datafusion",
-    "src/moonlink_metadata_store",
-    "src/moonlink_proto",
-    "src/moonlink_rpc",
-    "src/moonlink_service",
-    "src/moonlink_table_metadata",
+  "src/moonlink",
+  "src/moonlink_backend",
+  "src/moonlink_connectors",
+  "src/moonlink_datafusion",
+  "src/moonlink_error",
+  "src/moonlink_metadata_store",
+  "src/moonlink_proto",
+  "src/moonlink_rpc",
+  "src/moonlink_service",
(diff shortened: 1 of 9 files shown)


dandedotdev:
Thanks for the review!

I didn’t disable all-features in this PR, because I think it’s useful to proactively check whether all potential future features comply with the rules.

Of course, I could also remove the warning suppression I added, so the warnings would still be visible. However, given that we use so many crates, and many of them depend on the same underlying dependencies we also use, I believe these duplicates are quite difficult to avoid.

dentiny:
> Thanks for the review!
> 
> I didn’t disable all-features in this PR, because I think it’s useful to proactively check whether all potential future features comply with the rules.
> 
> Of course, I could also remove the warning suppression I added, so the warnings would still be visible. However, given that we use so many crates, and many of them depend on the same underlying dependencies we also use, I believe these duplicates are quite difficult to avoid.

gotcha, thank you! Let me have a try on taplo

PR #1966, opened 2025-09-07
refactor: catalog tests impl uses catalog test guard for resource management

peterxcli:
## Summary

## Related Issues

## Changes

- 
- 
- 

## Checklist

- [ ] Code builds correctly

### DIFF
--- a/src/moonlink/src/storage/iceberg/catalog_test_guard.rs
+++ b/src/moonlink/src/storage/iceberg/catalog_test_guard.rs
@@ -1,21 +1,20 @@
-use crate::storage::iceberg::rest_catalog::RestCatalog;
 /// A RAII-style test guard, which creates namespace ident, table ident at construction, and deletes at destruction.
 use crate::storage::iceberg::rest_catalog_test_utils::*;
 use iceberg::{Catalog, NamespaceIdent, Result, TableIdent};
 use std::collections::HashMap;
 
-pub(crate) struct RestCatalogTestGuard {
+pub(crate) struct CatalogTestGuard<'a> {
+    pub(crate) catalog: &'a dyn Catalog,
     pub(crate) namespace: NamespaceIdent,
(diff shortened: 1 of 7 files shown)


dentiny:
Cc @Standing-Man you might be working on sth similar or related

src/moonlink/src/storage/iceberg/catalog_test_guard.rs
```diff
@@ -1,21 +1,20 @@
-use crate::storage::iceberg::rest_catalog::RestCatalog;
 /// A RAII-style test guard, which creates namespace ident, table ident at construction, and deletes at destruction.
 use crate::storage::iceberg::rest_catalog_test_utils::*;
 use iceberg::{Catalog, NamespaceIdent, Result, TableIdent};
 use std::collections::HashMap;
 
-pub(crate) struct RestCatalogTestGuard {
+pub(crate) struct CatalogTestGuard<'a> {
+    pub(crate) catalog: &'a dyn Catalog,
```
fweilun, line 7: Have you checked using Arc<dyn Catalog> for solving the ownership issue?

Standing-Man: In PR #1875, I used `Arc` to implement the RAII mechanism for `TestGuard`, but I’m not sure if using `RwLock` to avoid data races is appropriate.

fweilun: I’m curious why we need to use an RwLock. IMO, each TestGuard should have its own separate Catalog, so there shouldn’t be any shared mutable state that requires locking.

Standing-Man: Use `RwLock` only to ensure `catalog` is not accessed by multiple threads (can be removed later)

PR #1986, opened 2025-09-10
[storage] Add basic unit test for iceberg rest

dentiny:
## Summary

This PR adds the basic unit test for iceberg rest catalog. 

To me, most of the rest catalog functionality should be fairly straightforward, since it's directly delegating to internal rest catalog without any additional logic; there're a few key things I'm concerned about:
- table creation
- schema setting (at table creation) and evolution
- table update / transaction commmit

These three items are the places we do our own hack. In this PR, I added least unit tests which cover these topics.

The ideal state for catalog test is to share the same testing logic with file catalog and maybe glue in the future, but there're certain blockers:
- file catalog unit tests use temporary directories, which are naturally isolated with each other; no extra test guard needed, no random table and namespace name needed
- but for rest catalog, all unit tests are accessing the devcontainer at the same time, which means they interfere with each other if we use the same namespace / table; tried to do some refactoring on random tables, but that requires huge refactor which I cannot afford at the moment

## Checklist

- [x] Code builds correctly

### DIFF
--- a/src/moonlink/src/storage/iceberg.rs
+++ b/src/moonlink/src/storage/iceberg.rs
@@ -81,3 +81,7 @@ mod snapshot_fetcher_test;
 
 #[cfg(test)]
 mod catalog_test_impl;
+
+#[cfg(feature = "catalog-rest")]
+#[cfg(test)]
+mod iceberg_rest_catalog_test;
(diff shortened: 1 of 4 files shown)


dentiny:
CC @peterxcli / @fweilun / @Standing-Man

Standing-Man:
Do you think it might be possible to isolate the `RestCatalog` unit tests?

PR #2007, opened 2025-09-11
Log error on replication eventloop failure

dentiny:
## Summary

We don't have dedicated thread block wait for the replication eventloop handler, which means when error happened, there's no logging or panic for the current implementation, which makes debugging impossible.

I think the general practice is:
- Don't exit on failure inside of eventloop
- For background task, consider error logging instead of general error propagation

## Checklist

- [x] Code builds correctly

### DIFF
--- a/src/moonlink_connectors/src/pg_replicate.rs
+++ b/src/moonlink_connectors/src/pg_replicate.rs
@@ -611,7 +611,14 @@ impl PostgresConnection {
         let cfg = self.source.get_cdc_stream_config().unwrap();
         let source = self.source.clone();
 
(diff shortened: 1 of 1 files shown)


nbiscaro:
Thanks, ooc which errors were not being logged within the loop?

dentiny:
> Thanks, ooc which errors were not being logged within the loop?

The error I met is outside of "loop" unfortunately: https://github.com/Mooncake-Labs/moonlink/blob/22500485019cf00f35ff6d78a4a100c7e5a54027/src/moonlink_connectors/src/pg_replicate.rs#L682

Issue #2023, opened 2025-09-12
[BUG] ErrorStruct deserialization Error When Using bincode::serde


fweilun:
## Related https://github.com/Mooncake-Labs/moonlink/pull/1746

I use this code to test serialization and deserialization for ErrorStruct, but the following error occurs.

```
let io_error = io::Error::new(io::ErrorKind::NotSeekable, "Test file not found");
let err = ErrorStruct::new("io error".to_string(), ErrorStatus::Permanent).with_source(io_error);
let config = bincode::config::standard();
let bytes = bincode::serde::encode_to_vec(&Err::<(), ErrorStruct>(err), config).unwrap();
let _: Result<(), ErrorStruct> = bincode::serde::decode_from_slice(&bytes, config).unwrap().0;
```

<img width="969" height="340" alt="Image" src="https://github.com/user-attachments/assets/d4421fed-6e87-4136-9aee-e04f754d243e" />

fweilun:
I'll fix it tomorrow.

dentiny:
Thank you so much for the help!

PR #2042, opened 2025-09-13
Fix ErrorStruct source serialization

fweilun:
Closes https://github.com/Mooncake-Labs/moonlink/issues/2023

## Changes

The bincode module uses a binary bit to indicate whether Option<String> is None or Some(value), which the original implementation did not support.

## Checklist

- [✓] Code builds correctly

### DIFF
--- a/src/moonlink_error/src/error.rs
+++ b/src/moonlink_error/src/error.rs
@@ -41,10 +41,8 @@ fn serialize_error_source<S>(
 where
     S: Serializer,
 {
-    match error {
-        Some(err) => serializer.serialize_str(&err.to_string()),
-        None => serializer.serialize_none(),
-    }
+    let s = error.as_ref().map(|e| e.to_string());
+    s.serialize(serializer)
 }
 
 fn deserialize_error_source<'de, D>(deserializer: D) -> Result<Option<Arc<anyhow::Error>>, D::Error>
@@ -118,6 +116,7 @@ impl ErrorStruct {
 #[cfg(test)]
 mod tests {
     use super::*;
+    use bincode;
     use regex::Regex;
     use std::io;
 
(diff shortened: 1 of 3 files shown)


fweilun:
cc @dentiny @machichima

src/moonlink_error/src/error.rs
```diff
@@ -199,4 +198,15 @@ mod tests {
...
     }
+
+    #[test]
+    fn test_error_struct_with_source_serialization_bincode() {
+        let io_error = io::Error::new(io::ErrorKind::NotFound, "Test file not found");
+        let error = ErrorStruct::new("IO operation failed".to_string(), ErrorStatus::Permanent)
+            .with_source(io_error);
+        let config = bincode::config::standard();
+        let bytes = bincode::serde::encode_to_vec(Err::<(), ErrorStruct>(error), config).unwrap();
+        let _: Result<(), ErrorStruct> =
```
dentiny, line 209: Several things we should do:
- We should check it's error, not ok
- Error message should be contained in the serialized form

dentiny (approved):
Thanks! I know you're busy this weekend, I will followup with the comment.

src/moonlink_error/src/error.rs
```diff
@@ -118,6 +116,7 @@ impl ErrorStruct {
 #[cfg(test)]
 mod tests {
     use super::*;
+    use bincode;
```
dentiny, line 119: why do we need bincode for testing?

dentiny: ic.. our rpc uses bincode for serde

src/moonlink_error/src/error.rs
fweilun: Thanks for reminding me.

Issue #2049, opened 2025-09-13
BUG: replay chaos test failure


dentiny:
## Summary

https://github.com/Mooncake-Labs/moonlink/actions/runs/17694831099/job/50293084858

dentiny:
another failure: https://github.com/Mooncake-Labs/moonlink/actions/runs/17695152928/job/50293798385?pr=2048

PR #2054, opened 2025-09-14
[storage] Fix async flush with streaming commit

dentiny:
## Summary

The bug is:
- Before streaming transaction commit, there're already completed async flush
- For the current implementation, it leaks for mooncake snapshot disk files
- But largest flush LSN is not updated, leading able to create iceberg snapshot
- Now the iceberg snapshot contains the completed flushed disk slice, which leads to data inconsistency for iceberg

## Related Issues

Closes https://github.com/Mooncake-Labs/moonlink/issues/2049

## Checklist

- [x] Code builds correctly

### DIFF
--- a/src/moonlink/src/storage/mooncake_table/transaction_stream.rs
+++ b/src/moonlink/src/storage/mooncake_table/transaction_stream.rs
@@ -648,6 +648,11 @@ impl MooncakeTable {
             deletion.lsn = lsn - 1;
         }
 
+        // Set largest flush LSN.
+        if !stream_state.flushed_files.is_empty() {
+            self.next_snapshot_task.try_set_largest_flush_lsn(lsn);
+        }
+
         let commit = TransactionStreamCommit {
             xact_id,
             commit_lsn: lsn,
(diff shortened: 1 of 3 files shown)


src/moonlink/src/storage/mooncake_table/transaction_stream.rs
```diff
@@ -648,6 +648,11 @@ impl MooncakeTable {
             deletion.lsn = lsn - 1;
         }
 
+        // Set largest flush LSN.
+        if !stream_state.flushed_files.is_empty() {
+            self.next_snapshot_task.try_set_largest_flush_lsn(lsn);
+        }
```
nbiscaro, line 654: Thanks for the fix! 

I don't fully understand the fix and why we update the flush LSN on commit? 

Also, can you explain what you mean by
> leaks for mooncake snapshot disk files

dentiny: - on txn commit, completed disk slices goes into snapshot disk slice, thats what I mean by leak
- current implementation (your flush lsn < min ongoing flush) is not enough to guard against creating iceberg snapshot with a partially flushed steaming txn

dentiny: If I were you, I will likely record a ongoing flush map in snapshot, instead of putting all disk slice in snapshot disk files

PR #2075, opened 2025-09-15
[storage] Avoid large manifest file

dentiny:
## Summary

When checking data files in the snapshot, I found we have giant manifest data files of ~30GiB, which is definitely bad for query. This PR limits the number of max manifest entries, so we could roughly cap the max file size.

## Checklist

- [ ] Code builds correctly


src/moonlink/src/storage/iceberg/tests.rs
```diff
@@ -317,6 +319,130 @@ async fn test_skip_iceberg_snapshot() {
...
+            new_deletion_vector: HashMap::new(),
+            file_indices: vec![],
+        },
+        index_merge_payload: IcebergSnapshotIndexMergePayload {
+            new_file_indices_to_import: vec![],
+            old_file_indices_to_remove: vec![],
+        },
+        data_compaction_payload: IcebergSnapshotDataCompactionPayload {
+            new_data_files_to_import: vec![],
+            old_data_files_to_remove: vec![data_files_to_import[0].clone()],
```
dentiny, line 421: to trigger a rewrite for manifest files

Issue #1940, opened 2025-09-06
object storage: add throughput and latency emulation config


dentiny:
## Summary

https://github.com/Mooncake-Labs/moonlink/blob/30bc3e8c1b54cee0cbe6cc113c75aeec80ea91d1/src/moonlink/src/storage/filesystem/accessor_config.rs#L118

so we could use local storage to emulator real usage

Andrewtangtang:
Hi @dentiny , can I work on this issue? thanks

dentiny:
> Hi [@dentiny](https://github.com/dentiny) , can I work on this issue? thanks

Sure, thanks!
Notice, I don't think it's a good first issue, ping me if needed, thanks!

PR #2078, opened 2025-09-16
storage: add throughput throttling with OpenDAL ThrottleLayer

Andrewtangtang:
## Summary

Implement throughput throttling layer in OpenDAL using `ThrottleLayer`, based on the **Generic Cell Rate Algorithm (GCRA)**.  
This layer enforces **per-accessor throughput control** across the storage backend and provides both bandwidth limiting and burst capacity.

Key characteristics (from OpenDAL official documentation https://docs.rs/opendal/latest/opendal/layers/struct.ThrottleLayer.html):

- **Bandwidth**: Maximum number of bytes allowed per second (using GCRA).
- **Burst**: Maximum number of bytes allowed to pass through at once.
- **Important**: The burst size **must be larger than any potential single operation size**, otherwise the request will always be rejected as `InsufficientCapacity`.

This implementation integrates the throttling layer at the filesystem accessor level so that **all write operations** can be rate-limited consistently. (Read operations currently pass through unthrottled.)

## Related Issues

- Related to #1940 (object storage: add throughput and latency emulation config)

## Changes

- Added `ThrottleConfig` with `bandwidth` (bytes/sec) and `burst` (bytes) parameters
- Integrated OpenDAL `ThrottleLayer` into filesystem accessor pipeline
- Exported `FsThrottleConfig` for public API usage
- Added tests covering:
  - Sequential writes under throttling
  - Parallel writes with shared limiter
  - Burst capacity limits (single request exceeding burst must fail)
- Updated `Cargo.toml` to include `layers-throttle` feature

## Additional Testing Considerations

The current test suite covers sequential writes, parallel writes, and burst capacity checks.  
If there are additional scenarios that reviewers consider important, please let me know and I will extend the tests accordingly.

## Checklist

- [x] Code builds correctly


src/moonlink/src/storage/filesystem/accessor_config.rs
```diff
@@ -109,7 +109,42 @@ impl Default for TimeoutConfig {
...
+    #[serde(default = "ThrottleConfig::default_bandwidth")]
+    pub bandwidth: u32,
 
+    /// Burst size in bytes
+    #[serde(default = "ThrottleConfig::default_burst")]
+    pub burst: u32,
+}
+
+impl ThrottleConfig {
+    const DEFAULT_BANDWIDTH: u32 = 10 * 1024 * 1024; // 10MB/s
```
dentiny: The value is too small to mimic object storage, 10x is much suitable

src/moonlink/src/storage/filesystem/accessor/throttle_test.rs
```diff
@@ -0,0 +1,204 @@
...
+        let baseline_accessor = create_throttle_accessor(&temp_dir, 1000.0, 1000.0);
+        let start_time = Instant::now();
+        for i in 0..num_files {
+            baseline_accessor
+                .write_object(&format!("baseline_{i}.dat"), test_data.clone())
+                .await
+                .unwrap();
+        }
+        let baseline_duration = start_time.elapsed();
+        println!("Sequential writes test:");
```
dentiny: let's remove the logging in testing; `cargo test` runs all unit test in parallel, logging is not that useful in this case.

src/moonlink/src/storage/filesystem/accessor/throttle_test.rs
```diff
@@ -0,0 +1,204 @@
...
+#[tokio::test]
+async fn test_throttle_sequential_writes() {
+    let temp_dir = tempdir().unwrap();
+    let test_future = async {
+        let file_size = 1024 * 1024; // 1 MB per file
+        let num_files = 6;
+        let test_data = vec![b'x'; file_size];
+
+        // Test with throttle configuration
+        let throttled_accessor = create_throttle_accessor(&temp_dir, 1.0, 2.0); // 1 MB/s, 2 MB burst
```
dentiny:
```suggestion
        let throttled_accessor = create_throttle_accessor(&temp_dir, /*bandwidth_mbps=*/1.0, /*burst_mb=*/2.0); // 1 MB/s, 2 MB burst
```

src/moonlink/src/storage/filesystem/accessor/throttle_test.rs
```diff
@@ -0,0 +1,204 @@
...
+        for i in 0..num_files {
+            throttled_accessor
+                .write_object(&format!("throttled_{i}.dat"), test_data.clone())
+                .await
+                .unwrap();
+        }
+        let throttled_duration = start_time.elapsed();
+
+        // Test without throttle (high limits)
+        let baseline_accessor = create_throttle_accessor(&temp_dir, 1000.0, 1000.0);
```
dentiny:
```suggestion
        let baseline_accessor = create_throttle_accessor(&temp_dir, /*bandwidth_mbps=*/1000.0, /*burst_mb=*/1000.0);
```

src/moonlink/src/storage/filesystem/accessor/throttle_test.rs
```diff
@@ -0,0 +1,204 @@
...
+        println!(
+            "  Baseline (1000MB/s): {:.3}s",
+            baseline_duration.as_secs_f64()
+        );
+        println!(
+            "  Slowdown ratio: {:.2}x",
+            throttled_duration.as_secs_f64() / baseline_duration.as_secs_f64()
+        );
+        // Throttled operations should take longer than baseline
+        assert!(
```
dentiny:
```suggestion
        ma::assert_lt!(
```
which prints out value on assertion failure

src/moonlink/src/storage/filesystem/accessor/throttle_test.rs
```diff
@@ -0,0 +1,204 @@
...
+                }
+            });
+            handles.push(handle);
+        }
+
+        for handle in handles {
+            handle.await.unwrap();
+        }
+        let baseline_duration = start_time.elapsed();
+        println!("Parallel writes test:");
```
dentiny: same for the following tests, remove logging and use `ma::assert`

dentiny (commented):
Thanks! Overall LG

Andrewtangtang:
Hi @dentiny 
Thanks for pointing that out! Right now the throttling is only per-accessor, so it doesn’t really enforce a global cap.

I’m thinking of switching to the [`governor`](https://docs.rs/governor/latest/governor/) crate and letting accessors with the same config share a single `Arc<RateLimiter>`.

The plan would be:

* Have a global manager that maps each throttle config to a shared limiter
* Add a small custom OpenDAL layer that calls `limiter.until_n_ready(n).await` before writes

That way, all accessors will respect the same global limit and we’ll get more realistic storage emulation. If this direction sounds good, I can go ahead and rework the implementation in that direction.

dentiny:
> Thanks for pointing that out! Right now the throttling is only per-accessor, so it doesn’t really enforce a global cap.

It's not bad --- the purpose of the ticket is to provide a way so everyone could conduct benchmark with no access to object storage, especially for open source contributors

src/moonlink/src/storage/filesystem/accessor/throttle_test.rs
```diff
@@ -0,0 +1,182 @@
...
+const TEST_TIMEOUT: Duration = Duration::from_secs(30);
+
+/// Create accessor with throttle configuration
+fn create_throttle_accessor(
+    temp_dir: &tempfile::TempDir,
+    bandwidth_mbps: f64,
+    burst_mb: f64,
+) -> std::sync::Arc<dyn BaseFileSystemAccess> {
+    let throttle_config = Some(ThrottleConfig {
+        bandwidth: (bandwidth_mbps * 1024.0 * 1024.0) as u32,
```
dentiny: not too big of an issue, but prefer to pass number of bytes directly

src/moonlink/src/storage/filesystem/accessor/throttle_test.rs
```diff
@@ -0,0 +1,182 @@
...
+        let baseline_duration = start_time.elapsed();
+        // Parallel throttled operations should take longer than baseline
+        ma::assert_gt!(
+            throttled_duration,
+            baseline_duration,
+            "Parallel throttled operations should be slower than baseline",
+        );
+    };
+
+    timeout(TEST_TIMEOUT, test_future).await.unwrap();
```
dentiny: I'm not sure why `timeout` is needed here, in theory all unit tests should be capped with an upper bound. If you think it's taking super long (which I don't think that's the case based on CI), we should move it to integration test maybe

Andrewtangtang:
I plan to open a new PR with the global limiter refactor and also refine the testing code. This PR is now a draft and will be closed once the new PR is ready.

dentiny:
> I plan to open a new PR with the global limiter refactor and also refine the testing code. This PR is now a draft and will be closed once the new PR is ready.

Do you mind telling me you plan first? I think current implementation is not bad.

src/moonlink/src/storage/filesystem/accessor/operator_utils.rs
```diff
@@ -1,9 +1,11 @@
 use crate::storage::filesystem::accessor::filesystem_accessor_chaos_wrapper::ChaosLayer;
 use crate::storage::filesystem::accessor_config::AccessorConfig;
 use crate::storage::filesystem::accessor_config::RetryConfig;
+use crate::storage::filesystem::accessor_config::ThrottleConfig;
 use crate::storage::filesystem::accessor_config::TimeoutConfig;
 use crate::storage::filesystem::storage_config::StorageConfig;
 use crate::Result;
+use opendal::layers::ThrottleLayer;
```
dentiny: the dependency break doesn't look right, place it along with third-party dependencies?

src/moonlink/src/storage/filesystem/accessor_config.rs
```diff
@@ -109,7 +109,42 @@ impl Default for TimeoutConfig {
...
+/// Throttle config
+/// ========================
+///
+#[derive(Clone, Debug, PartialEq, Deserialize, Serialize)]
+pub struct ThrottleConfig {
+    /// Bandwidth in bytes per second
+    #[serde(default = "ThrottleConfig::default_bandwidth")]
+    pub bandwidth: u32,
 
+    /// Burst size in bytes
```
dentiny: nit: add more details, for example, what if request size exceeds burst size

src/moonlink/src/storage/filesystem/accessor/throttle_test.rs
```diff
@@ -0,0 +1,165 @@
...
+        baseline_duration,
+        "Throttled operations should be slower than baseline",
+    );
+}
+
+#[tokio::test]
+async fn test_throttle_parallel_writes() {
+    let temp_dir = tempdir().unwrap();
+    let file_size = 1024 * 1024; // 1 MB per file
+    let num_parallel = 4;
```
dentiny: emm just curious why do we have two parallel parameters?

Andrewtangtang: The `num_parallel` here controls the number of concurrent write tasks, where each task writes two fixed 1MB files. This helps verify that the throttle layer correctly handles simultaneous requests.
I can rename it to something like `concurrent_tasks` to make it clearer, and I’ll also add an inline comment to explain this.

src/moonlink/src/storage/filesystem/accessor_config.rs
```diff
@@ -109,7 +109,45 @@ impl Default for TimeoutConfig {
...
     }
 }
+/// ========================
+/// Throttle config
+/// ========================
+///
+#[derive(Clone, Debug, PartialEq, Deserialize, Serialize)]
+pub struct ThrottleConfig {
+    /// Bandwidth in bytes per second
+    /// Maximum 4GB.
```
dentiny: Is 4GB from opendal?

Andrewtangtang: Yes, the 4GB limit comes from OpenDAL ThrottleLayer's implementation using u32 type for bandwidth and burst parameters.

dentiny: That's not really ideal... for GCE internet work it could reach 1Tib/s...

dentiny: but let's keep it for now, for such large load we don't really throttler at client side

dentiny: btw its `GiB`, not `GB`

Andrewtangtang: Thanks for the reminder.

dentiny (approved):
Thanks!

PR #2083, opened 2025-09-19
Avro schema JSON

lohpaul9:
## Summary

Create table with avro schema takes in JSON schema instead of stringified JSON schema.


zhousun:
any reason not to push?

Issue #1560, opened 2025-08-18
deployment: basic validation against standalone deployment


dentiny:
## Summary

Before every deployment, we should have some validation process against dev or staging instance.
The idea is similar to our unit test: https://github.com/Mooncake-Labs/moonlink/blob/main/src/moonlink_service/src/test.rs
Basically we create some tables, inserts some data, and query its content to check whether everything works fine.
- it could be hard-coded operations at the very beginning
- but at the end it should take a file, or generate random operations

At the end of the day, it should be part of the release process.
To test on remote standalone deployment, we could use kind for local testing and dev.

fweilun:
I'd like to address this issue, thanks!

dentiny:
> I'd like to address this issue, thanks!

Assigned thanks!

fweilun:
I wonder where I should put this test.
Does `src/moonlink_service/deploy/validation.rs` seem like a good location?

dentiny:
> Does src/moonlink_service/deploy/validation.rs seem like a good location?

Sounds good.

PR #2091, opened 2025-09-21
Basic validation for standalone deployment

fweilun:
## Summary

This PR provides some tests to verify that the deployment is working correctly.
1. create table
2. insert data
3. query data and check
4. drop table

Tested using the following standalone server config:

```
pub(crate) const BASE_PATH: &str = "/workspaces/moonlink/.shared-nginx";
pub(crate) const DATA_SERVER_URI: &str = "http://nginx.local:80";

let config = ServiceConfig {
    base_path: BASE_PATH.to_string(),
    data_server_uri: Some(DATA_SERVER_URI.to_string()),
    rest_api_port: Some(3030),
    tcp_port: Some(3031),
    otel_api_port: Some(3435),
    log_directory: None,
};
```

## Related Issues

Closes https://github.com/Mooncake-Labs/moonlink/issues/1560

## Checklist

- [✓] Code builds correctly

### DIFF
--- a/src/moonlink_service/deploy/validation.rs
+++ b/src/moonlink_service/deploy/validation.rs
@@ -121,7 +121,7 @@ pub(crate) async fn create_table(client: &reqwest::Client, database: &str, table
 
 pub(crate) async fn drop_table(client: &reqwest::Client, database: &str, table: &str) {
     let crafted_src_table_name = format!("{database}.{table}");
-    let payload = get_drop_table_payload(DATABASE, TABLE);
+    let payload = get_drop_table_payload(database, table);
     let response = client
         .delete(format!("{REST_ADDR}/tables/{crafted_src_table_name}"))
         .header("content-type", "application/json")


fweilun:
> Why iceberg related change?

Ohhh, I'm fixing this.

src/moonlink_service/deploy/validation.rs
```diff
@@ -0,0 +1,209 @@
+use arrow::datatypes::Schema as ArrowSchema;
```
dentiny, line 1: What's the difference with https://github.com/Mooncake-Labs/moonlink/blob/main/src/moonlink_service/src/test.rs?

fweilun: The goal is to separate deploy and test code. Once different deployments are up, I can use this minimal standalone test to quickly check if the service is healthy.

I'd like to know what should I do to improve this PR.

dentiny: The referenced file is used for CI, the issue you're working on is for CD.
There're a few difference on test code:
- CD could run much longer, for example, stress test, perf test, etc
- It's testing against dev or staging env
- It tests compatibility, including backward and forward compatibility

I would break down the issue to several parts:
- First have some basic testing, which I highly recommends to reuse existing CI to avoid code duplication
- Add the above tests one by one

fweilun: I plan to do the following.

In this PR: Create a new workflow file `cd.yml` that runs `moonlink_service/src/test.rs`.
In follow-up PRs: add the other CD-specific tests (stress, performance, compatibility, etc.) step by step.

Does that make sense to you?

dentiny: > In this PR: Create a new workflow file cd.yml that runs moonlink_service/src/test.rs.

What's the IP address do you plan to use?
Currently we don't even have a stable deployment, so I imagine them to be adhoc at first.

Issue #2105, opened 2025-09-24
[BUG] Initial Load fail for table with index [UNBREAK NOW]


zhousun:
## Description

`thread 'tokio-runtime-worker' panicked at moonlink/src/moonlink/src/row/moonlink_row.rs:339:31:
index out of bounds: the len is 2 but the index is 2
stack backtrace:
   0: __rustc::rust_begin_unwind
             at /rustc/6b00bc3880198600130e1cf62b8f8a93494488cc/library/std/src/panicking.rs:697:5
   1: core::panicking::panic_fmt
             at /rustc/6b00bc3880198600130e1cf62b8f8a93494488cc/library/core/src/panicking.rs:75:14
   2: core::panicking::panic_bounds_check
             at /rustc/6b00bc3880198600130e1cf62b8f8a93494488cc/library/core/src/panicking.rs:281:5
   3: <usize as core::slice::index::SliceIndex<[T]>>::index
             at /rustc/6b00bc3880198600130e1cf62b8f8a93494488cc/library/core/src/slice/index.rs:274:10
   4: core::slice::index::<impl core::ops::index::Index<I> for [T]>::index
             at /rustc/6b00bc3880198600130e1cf62b8f8a93494488cc/library/core/src/slice/index.rs:16:15
   5: <alloc::vec::Vec<T,A> as core::ops::index::Index<I>>::index
             at /rustc/6b00bc3880198600130e1cf62b8f8a93494488cc/library/alloc/src/vec/mod.rs:3376:9
   6: moonlink::row::moonlink_row::IdentityProp::get_lookup_key
             at /home/ubuntu/pg_mooncake/moonlink/src/moonlink/src/row/moonlink_row.rs:339:31
   7: moonlink::storage::mooncake_table::batch_ingestion::<impl moonlink::storage::mooncake_table::MooncakeTable>::build_index_for_files::{{closure}}
             at /home/ubuntu/pg_mooncake/moonlink/src/moonlink/src/storage/mooncake_table/batch_ingestion.rs:190:36
   8: moonlink::storage::mooncake_table::batch_ingestion::<impl moonlink::storage::mooncake_table::MooncakeTable>::batch_ingest::{{closure}}
             at /home/ubuntu/pg_mooncake/moonlink/src/moonlink/src/storage/mooncake_table/batch_ingestion.rs:147:88
`

fails if index is not the first column.

The problem is we only fetch index columns in parquet reader, 
[...]

PR #2107, opened 2025-09-24
Fix batch index build

nbiscaro:
## Summary

When reading the parquet for index build during batch ingestion we project only identity columns via `ProjectionMask`, but then compute the key using the generic `get_lookup_key` which assumes original column positions when indexing. This causes an out of bounds for any table that does not have the first column as a primary key. 

Added `get_lookup_key_from_identity_row` which assumes the row is already projected and contains only identity columns in identity order.

## Related Issues

Closes https://github.com/Mooncake-Labs/moonlink/issues/2105
## Changes

- Add `get_lookup_key_from_identity_row`
- Add regression test
- 

## Checklist

- [ ] Code builds correctly

### DIFF
--- a/src/moonlink/src/row/moonlink_row.rs
+++ b/src/moonlink/src/row/moonlink_row.rs
@@ -357,6 +357,7 @@ impl IdentityProp {
         match self {
             IdentityProp::SinglePrimitiveKey(_) => row.values[0].to_u64_key(),
             IdentityProp::Keys(keys) => {
+                assert_eq!(row.values.len(), keys.len());
                 let mut hasher = AHasher::default();
                 for i in 0..keys.len() {
                     row.values[i].hash(&mut hasher);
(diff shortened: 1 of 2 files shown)


src/moonlink/src/row/moonlink_row.rs
```diff
@@ -351,6 +351,29 @@ impl IdentityProp {
         }
     }
 
+    /// Compute lookup key assuming `row` contains only identity columns in identity order.
```
zhousun, line 354: assert the length match
nit: the three branches are actually the same

zhousun:
otherwise lgtm

dentiny:
>  generic get_lookup_key which assumes original column positions when indexing

Hmmm I will followup to rename `get_lookup_key` and avoid confusion, I didn't notice that at first glance as well :cry:

Issue #1356, opened 2025-08-11
observability: add stats for iceberg table recovery


dentiny:
## Summary

Code pointer:
- https://github.com/Mooncake-Labs/moonlink/blob/db080d596ffd11d073689d03ecdf2dc1af5ceab5/src/moonlink/src/storage/iceberg/iceberg_table_manager.rs#L210-L212

Open question:
- what kind of stats to collect?
- what metrics to collect?
- per-table or per-backend metrics?
- need finer-granularity stats?

fweilun:
HI @dentiny , Could you assign this to me? thanks!

dentiny:
> HI [@dentiny](https://github.com/dentiny) , Could you assign this to me? thanks!

Sure, be sure to let me know your answers to these question before implementing. Thanks!

fweilun:
HI @dentiny , I plan to do the following.

what kind of stats to collect?
    1. load data files
    2. load file indices and deletion vector.
    3. manifest_file number

what metrics to collect?
    Histogram

per-table or per-backend metrics?
    per-table

dentiny:
@fweilun sounds good to me

For manifest_file number, we might be also interested in manifest entries count.
We could do latency metrics collection first then manifest ones.

fweilun:
> [@fweilun](https://github.com/fweilun) sounds good to me
> 
> For manifest_file number, we might be also interested in manifest entries count. We could do latency metrics collection first then manifest ones.

Sure, I'll do the following for the PR.
1. load data files
2. load file indices and deletion vector.

PR #2117, opened 2025-09-24
Add stats for iceberg table recovery

fweilun:
## Summary

Add iceberg table recovery stats:
1. load_data_files_latency histogram
2. load_file_indices_deletion_vector_latency histogram

## Related Issues

Part of https://github.com/Mooncake-Labs/moonlink/issues/1356

### DIFF
--- a/src/moonlink/src/storage/iceberg/iceberg_table_manager.rs
+++ b/src/moonlink/src/storage/iceberg/iceberg_table_manager.rs
@@ -88,9 +88,6 @@ pub struct IcebergTableManager {
 
     /// Iceberg persistence stats for transaction commit.
     pub(crate) persistence_stats_transaction_commit: Arc<IcebergPersistenceStats>,
-
-    /// Iceberg table recovery stats.
-    pub(crate) recovery_stats: Arc<IcebergTableRecoveryStats>,
 }
 
 impl IcebergTableManager {
@@ -132,10 +129,9 @@ impl IcebergTableManager {
                 IcebergPersistenceStage::DeletionVectors,
             )),
             persistence_stats_transaction_commit: Arc::new(IcebergPersistenceStats::new(
-                mooncake_table_id.clone(),
+                mooncake_table_id,
                 IcebergPersistenceStage::TransactionCommit,
             )),
-            recovery_stats: Arc::new(IcebergTableRecoveryStats::new(mooncake_table_id)),
         })
     }
 
@@ -181,10 +177,9 @@ impl IcebergTableManager {
                 IcebergPersistenceStage::DeletionVectors,
             )),
             persistence_stats_transaction_commit: Arc::new(IcebergPersistenceStats::new(
-                mooncake_table_id.clone(),
+                mooncake_table_id,
                 IcebergPersistenceStage::TransactionCommit,
             )),
-            recovery_stats: Arc::new(IcebergTableRecoveryStats::new(mooncake_table_id)),
         })
     }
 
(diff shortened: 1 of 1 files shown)


fweilun:
I will rebase this after the previous PR is merged.

src/moonlink/src/storage/iceberg/iceberg_table_manager.rs
```diff
@@ -70,6 +71,9 @@ pub struct IcebergTableManager {
 
     /// Maps from remote data file path to its file id.
     pub(crate) remote_data_file_to_file_id: HashMap<String, FileId>,
+
+    /// Iceberg table recovery stats.
+    pub(crate) iceberg_recovery_stats: Arc<IcebergTableRecoveryStats>,
```
dentiny: we don't need `Arc` here, right?

src/moonlink/src/storage/iceberg/iceberg_table_loader.rs
```diff
@@ -331,6 +335,13 @@ impl IcebergTableManager {
                 }
             }
         }
+        let file_indices_deletion_vector_latency = time.elapsed().as_millis() as u64;
```
dentiny: Do you think it better if we first parallelize manifest entries load?

src/moonlink/src/storage/iceberg/iceberg_table_loader.rs
```diff
@@ -331,6 +335,13 @@ impl IcebergTableManager {
...
+        let mooncake_table_id = self.mooncake_table_metadata.mooncake_table_id.clone();
+        self.iceberg_recovery_stats.update(
```
dentiny: Or we could do the simplest thing first: record latency for overall recovery latency

src/moonlink/src/storage/iceberg/iceberg_table_manager.rs
fweilun: Ahh, my bad

src/moonlink/src/storage/iceberg/iceberg_table_loader.rs
fweilun: Yes, I think so.

src/moonlink/src/storage/iceberg/iceberg_table_loader.rs
fweilun: Do you mean using one update function to record data_files + file_indices + deletion_vector latency?

dentiny: oh sorry for the confusion, i mean we could record overall latency first

src/moonlink/src/storage/iceberg/iceberg_table_manager.rs
fweilun: We have to use Arc<T>. Since the latency guard borrows `&self`, we can't keep a`&mut self` at the same time..

src/moonlink/src/storage/iceberg/iceberg_table_manager.rs
```diff
@@ -259,6 +266,9 @@ impl TableManager for IcebergTableManager {
     }
 
     async fn load_snapshot_from_table(&mut self) -> Result<(u32, MooncakeSnapshot)> {
+        // Start recording iceberg recovery latency
+        let recovery_stats = self.recovery_stats.clone();
```
dentiny: We don't need to keep a separate variable for recovery, since it's called at most once

fweilun: I use this value to avoid borrow issue.

`self.load_snapshot_from_table_impl` takes mutable self, so we can't hold a `&self`for LatencyGuard.

dentiny: I mean, you can use local variable

fweilun: Do you mean to initialize `IcebergTableRecoveryStats` here?

```
let recovery_stats = IcebergTableRecoveryStats::new()
let _guard = recovery_stats.start()
```

dentiny: Yes.

PR #2122, opened 2025-09-25
Link with mold in devcontainer

dentiny:
## Summary

## Related Issues

## Changes

- 
- 
- 

## Checklist

- [ ] Code builds correctly

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -3,6 +3,7 @@ FROM mcr.microsoft.com/devcontainers/rust:latest
 RUN apt update \
  && apt install -y python3-pip \
  && apt install -y mold \
+ && sudo ln -fs /usr/local/bin/mold /usr/bin/ld \
  && rm -rf /var/lib/apt/lists/*
 
 USER vscode


.devcontainer/Dockerfile
```diff
@@ -3,6 +3,7 @@ FROM mcr.microsoft.com/devcontainers/rust:latest
 RUN apt update \
  && apt install -y python3-pip \
  && apt install -y mold \
+ && sudo ln -fs /usr/bin/mold /usr/bin/ld \
```
dpxcc, line 6: 1. rm `sudo`, you are running as root here
2. It's better to move `ln -sf /usr/bin/mold /usr/bin/ld` after `rm -rf /var/lib/apt/lists/*`, the latter is to clean apt cache, and therefore should be grouped with the apt commands above
3. also better to group those two `apt install`

See https://github.com/Mooncake-Labs/pg_mooncake/blob/main/.devcontainer/Dockerfile#L13

dentiny: sure, sounds good

PR #2128, opened 2025-09-26
otel: Implement RAII latency guard

dentiny:
## Summary

Implement a RAII guard which exports latency stats at destruction, so we don't need to record timestamp twice (before and after operation) and emits automatically.

## Checklist

- [ ] Code builds correctly

### DIFF
--- a/src/moonlink/src/observability/latency_exporter.rs
+++ b/src/moonlink/src/observability/latency_exporter.rs
@@ -1,0 +1,10 @@
+use crate::observability::latency_guard::LatencyGuard;
+
+/// An interface to export latency.
+pub(crate) trait BaseLatencyExporter: Send + Sync {
+    /// Start recording latency.
+    /// Returned latency guard will automatically records latency stats.
+    fn start<'a>(&'a self) -> LatencyGuard<'a>;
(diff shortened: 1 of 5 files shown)


dentiny:
CC @fweilun

fweilun:
I'm curious why should we use `LatencyGuard`.

I think this design is most suitable for measuring the latency of an entire function. In other cases, we would have to wrap the recorded code block in a closure.
2. Having multiple stats within one struct would be difficult to implement.

Personally, I feel a straightforward `start/finish` approach might be simpler.

dentiny:
The purpose is to provide RAII-like structure, so we don't forget to record on completion.

> I think this design is most suitable for measuring the latency of an entire function. In other cases, we would have to wrap the recorded code block in a closure. 

Yes, it's intentional. Either extracting into a separate function or quote with a code block.

> Having multiple stats within one struct would be difficult to implement.

I'm not sure how single or multiple metrics has something to do with this PR;
IMO you can and should record them separately?

> Personally, I feel a straightforward `start/finish` approach might be simpler.

I also keep the raw `record` function, so you can still use it if you think it simplier in certain cases.

PR #2141, opened 2025-09-30
fix opentelemetry unit test.

fweilun:
## Summary

Change the runtime configuration from single-threaded to multi-threaded.

## Checklist

- [✓] Code builds correctly

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -299,7 +299,21 @@ jobs:
     steps:
       - uses: actions/checkout@v5
 
+      # ────────────────────── Toolchain + Test Tools ─────────────────────────
       - uses: dtolnay/rust-toolchain@stable
+        id: toolchain
+        with:
+          components: llvm-tools-preview
+
+      - uses: swatinem/rust-cache@v2
+        with:
+          shared-key: ${{ steps.toolchain.outputs.cachekey }}
+
+      - uses: rui314/setup-mold@v1
+        with:
+          mold-version: 2.40.3
+          make-default: true
+
       - name: Opentelemetry feature test
         env:
             MOONLINK_BACKEND_DIR: ${{ github.workspace }}/.shared-nginx


dentiny (commented):
Do you mind adding it to CI? Thanks!

fweilun:
> Do you mind adding it to CI? Thanks!

Should I only add this test or the whole "otel-integration" feature

dentiny:
> Should I only add this test or the whole "otel-integration" feature

Right now there's only one test

dentiny:
> thread 'tokio-runtime-worker' panicked at src/moonlink_service/src/lib.rs:102:10:
called `Result::unwrap()` on an `Err` value: Io(ErrorStruct { message: "IO error", status: Permanent, source: Some(Failed to create directory "/workspaces/moonlink/.shared-nginx"

I guess you need to set backend directory via test env.

fweilun:
> running 1 test
> test otel::test::test_opentelemetry_export has been running for over 60 seconds
> Error: The operation was canceled.

HI @dentiny ,
Could you give me some guidance. It runs perfectly on my local env ( devcontainer ).

dentiny:
Check https://github.com/Mooncake-Labs/moonlink/blob/ce1d900515cbf2c122801e420016cf9fb56f8e48/.github/workflows/ci.yml#L721

fweilun:
It works, thanks.

.github/workflows/ci.yml
```diff
@@ -291,7 +291,22 @@ jobs:
...
+    name: Opentelemetry tests (feature)
+    runs-on: ubuntu-latest
+    timeout-minutes: 5
+    steps:
+      - uses: actions/checkout@v5
+      - uses: dtolnay/rust-toolchain@stable
+      - name: Opentelemetry feature test
+        env:
+            MOONLINK_BACKEND_DIR: ${{ github.workspace }}/.shared-nginx
+            NGINX_ADDR: http://nginx.local:80
```
dentiny: Do we need NGINX here?

.github/workflows/ci.yml
```diff
@@ -291,7 +291,22 @@ jobs:
...
   
-  # ──────────────────────── 5 · Rest catalog test (feature) ─────────────────
+  # ──────────────────────── 5 · Otel test (feature) ─────────────────
+  otel_test:
+    name: Opentelemetry tests (feature)
+    runs-on: ubuntu-latest
+    timeout-minutes: 5
+    steps:
+      - uses: actions/checkout@v5
+      - uses: dtolnay/rust-toolchain@stable
```
dentiny, line 303: Curious do you happen to know the advantage over
```
- uses: dtolnay/rust-toolchain@stable
        id: toolchain
        with:
          components: llvm-tools-preview
```
I would like to keep code consistency across CI workflow

.github/workflows/ci.yml
fweilun: I'll drop it. It is the same as the default settings.

.github/workflows/ci.yml
fweilun: It is used for specifying rust compiler, cargo, rust-std, ...

I think it's useful.

https://rust-lang.github.io/rustup/concepts/index.html?utm_source=chatgpt.com

dentiny: Please checkout other CI workflow configs

.github/workflows/ci.yml
dentiny: There's no "default" setting...

.github/workflows/ci.yml
```diff
@@ -291,7 +291,22 @@ jobs:
...
           ${{ matrix.cmd }}
   
-  # ──────────────────────── 5 · Rest catalog test (feature) ─────────────────
+  # ──────────────────────── 5 · Otel test (feature) ─────────────────
+  otel_test:
+    name: Opentelemetry tests (feature)
+    runs-on: ubuntu-latest
+    timeout-minutes: 5
+    steps:
+      - uses: actions/checkout@v5
```
dentiny, line 300: I mean, we should keep code consistency for all workflow setups.
https://github.com/Mooncake-Labs/moonlink/blob/9346a832d239110d10777e61e996477684b28d1f/.github/workflows/ci.yml#L351-L366
Take ^ as an example, what's the benefit for your implementation over existing ones?
As I mentioned [here](https://github.com/Mooncake-Labs/moonlink/pull/2141/files#r2390490607)

fweilun: ohh, ic.

PR #2154, opened 2025-10-08
fix(dev): devcontainer setting error

Standing-Man:
## Summary

In Debian 13 (Trixie), the APT repositories do not contain docker-ce and docker-ce-cli, making it impossible to locate and install the corresponding packages. Therefore, we have reverted to the stable Debian 12 (Bookworm) version

## Related Issues

## Changes

- 
- 
- 

## Checklist

- [ ] Code builds correctly

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -1,12 +1,11 @@
-FROM mcr.microsoft.com/devcontainers/rust:latest
+FROM mcr.microsoft.com/devcontainers/rust:bookworm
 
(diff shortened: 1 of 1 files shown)


Standing-Man:
> if we miss certain packages / dependencies, I think we should install it?

I noticed the download issue happens with 'docker-docker'. I'm not sure if docker-ce and docker-ce-cli are actually needed.

https://github.com/Mooncake-Labs/moonlink/blob/main/.devcontainer/devcontainer.json#L8

PR #2180, opened 2025-10-26
Fix/handle async lsn component receiver validation

thuongle2210:
## Summary

The PR resolve the asynchronous problems when event handling by using retry mechanism. The purpose of retries is to synchronize state from the next function wait_for_relevant_lsn_change, because the events are received from independent channels, which leads to asynchronous problems

## Related Issues

links to related issues: https://github.com/Mooncake-Labs/moonlink/issues/2181

## Changes

- Update mechanism to validate lsn ordering when reading state in ReadingStateManager
- Update Error type in moonlink package

## Checklist

- [x] Code builds correctly

### DIFF
--- a/src/moonlink/src/error.rs
+++ b/src/moonlink/src/error.rs
@@ -254,7 +254,7 @@ mod tests {
         if let Error::Io(ref inner) = io_error {
             let loc = inner.location.as_ref().unwrap();
             assert!(loc.contains("src/moonlink/src/error.rs"));
-            assert!(loc.contains("230"));
+            assert!(loc.contains("238"));
             assert!(loc.contains("9"));
         }
     }


src/moonlink/src/union_read/read_state_manager.rs
```diff
@@ -133,18 +137,50 @@ impl ReadStateManager {
...
+        if snapshot_lsn != NO_SNAPSHOT_LSN
+            && commit_lsn != NO_COMMIT_LSN
+            && snapshot_lsn > commit_lsn
+        {
+            return Err(Error::read_validation_error(format!(
+                "snapshot_lsn > commit_lsn: {} > {}",
+                snapshot_lsn, commit_lsn
+            )));
+        }
+        if commit_lsn > replication_lsn {
```
dentiny, line 156: What's the purpose of retry?

thuongle2210: The purpose of retries is to synchronize state from the next function `wait_for_relevant_lsn_change`, because the events are received from independent channels, which leads to asynchronous problems. You can reproduce the issue in my test case by setting const MAX_READ_SNAPSHOT_RETRIES: u8 = 0 to observe the problems.
In my case, if MAX_READ_SNAPSHOT_RETRIES: u8 = 0, I see 36 cases which exist the problem: replication_lsn<= commit_lsn. Causing a failure situation

dentiny: Not sure if it's a good fix or hiding the problem? It reads a synchronization problem to me, retry could also fail.

thuongle2210: Asynchronous problems are common in engineering, especially when handling events in systems like Kafka, Spark, and Flink. There are two main ways to address these issues:
First, combine all related events into sequential packages and use a single channel (similar to one partition in Kafka).
Second, add a retry mechanism with a timeout or retry limit to handle asynchronous events (such as watermarking).

dentiny: curious why don't we pick (1)?

thuongle2210: That's because I see MoonLink handles many background events, for example:
CdcEvent::PrimaryKeepAlive
CdcEvent::StreamCommit
CdcEvent::Commit
... etc.

Each Message Passing channel can be used in different Events or anywhere. To boost performance and make the code easier to scale, I see your team has broken it into many channels and run them asynchronously

dentiny: There're not too many channels: one for replication LSN, another for commit LSN?
Seems like related to
- https://github.com/Mooncake-Labs/moonlink/pull/2032
- https://github.com/Mooncake-Labs/moonlink/pull/2103

thuongle2210: Let me check. What about table_snapshot_watch_receiver? It is also an independent receiver and completely operates asynchronously.

thuongle2210: Honestly, retries are also a common mechanism to handle such issues. You can configure the retry count to be high or set an eventual timeout. This approach will help your code scale easily in the future.

